PR #8, opened 2021-11-26
enable HMR for development

ReeceHoffmann:
Changes

- HMR is now functional with webpack-dev-server.
  - HMR entry point is currently in index.js meaning that all subcomponents will be refreshed on every page change. Much faster than full refresh but it can be made faster for individual components by adding more granular updating can be added as needed by adding more `module.hot` handlers. (currently all changes to the react code bubble up to index.js)
- Dev and production servers now proxy WS and API calls to the backend rather than letting the client directly connect to the backend.
- Cleaned package.lock of now unnecessary libraries


igboyes (approved):
Funky, but it works.

PR #21, opened 2021-12-03
feat: use actionCreator for account actions

ReeceHoffmann:
Example implementation of action creation using createAction.

### DIFF
--- a/src/js/account/actions.js
+++ b/src/js/account/actions.js
@@ -3,7 +3,6 @@
  *
  * @module account/actions
  */
-import { simpleActionCreator } from "../utils/utils";
 import {
     GET_ACCOUNT,
     UPDATE_ACCOUNT,
@@ -18,6 +17,7 @@ import {
     CLEAR_API_KEY,
     RESET_PASSWORD
 } from "../app/actionTypes";
+import { createAction } from "@reduxjs/toolkit";
 
 /**
  * Returns action that can trigger an API call for getting the current account data.
@@ -25,7 +25,7 @@ import {
  * @func
  * @returns {object}
  */
-export const getAccount = simpleActionCreator(GET_ACCOUNT.REQUESTED);
+export const getAccount = createAction(GET_ACCOUNT.REQUESTED);
 
 /**
  * Returns an action that can trigger an API call for updating the current account.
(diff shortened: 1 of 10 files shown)


src/js/account/actions.js
```diff
@@ -46,9 +49,12 @@ export const updateAccount = update => ({
...
-export const updateAccountSettings = update => ({
-    type: UPDATE_ACCOUNT_SETTINGS.REQUESTED,
-    update
+export const updateAccountSettings = createAction(UPDATE_ACCOUNT_SETTINGS.REQUESTED, update => {
+    return {
+        payload: {
+            update
+        }
+    };
 });
```
igboyes: You can write this without a `return`.

src/js/account/actions.js
```diff
@@ -59,10 +65,13 @@ export const updateAccountSettings = update => ({
  * @param newPassword {string} the new account password
  * @returns {object}
  */
-export const changePassword = (oldPassword, newPassword) => ({
-    type: CHANGE_ACCOUNT_PASSWORD.REQUESTED,
-    oldPassword,
-    newPassword
+export const changePassword = createAction(CHANGE_ACCOUNT_PASSWORD.REQUESTED, (oldPassword, newPassword) => {
+    return {
```
igboyes: You can write this without a `return`.

src/js/account/actions.js
```diff
@@ -103,10 +110,8 @@ export const clearAPIKey = simpleActionCreator(CLEAR_API_KEY);
  * @param permissions {object} permissions configuration object for the new API key
  * @returns {object}
  */
-export const updateAPIKey = (keyId, permissions) => ({
-    type: UPDATE_API_KEY.REQUESTED,
-    keyId,
-    permissions
+export const updateAPIKey = createAction(UPDATE_API_KEY.REQUESTED, (keyId, permissions) => {
+    return { payload: { keyId, permissions } };
```
igboyes: You can write this without a `return`.

igboyes (changes requested):
This looks good to me and I think we should continue with it. I actually like the data being in a nested `payload` object.

Please deal with this warning from GitHub actions:
> Check warning on line 142 in src/js/errors/reducer.js

You can pare down a lot of anonymous arrow functions in this PR by eliminating `return` statements where possible. I only commented a few.

ReeceHoffmann:
> This looks good to me and I think we should continue with it. I actually like the data being in a nested `payload` object.
> 
> Please deal with this warning from GitHub actions:
> 
> > Check warning on line 142 in src/js/errors/reducer.js
> 
> You can pare down a lot of anonymous arrow functions in this PR by eliminating `return` statements where possible. I only commented a few.

Sounds good, I will get to work updating the rest of the action creators. I changed all the arrow functions to function as "one line" arrow functions (automatic return) and fixed the oversight that caused that warning.

igboyes (approved):

PR #23, opened 2021-12-07
Upgrade express server to use template engine

ReeceHoffmann:
Changes:
 - Upgrades the default path to serve the index page using EJS rather than reading the file fresh every time
 - Remove static path method for serving `.js .ico .map` extensions and replaced it with built in express.static function


igboyes (dismissed):
Awesome. Thanks!

ReeceHoffmann:
Sorry a push invalidated your review, changes are done now

igboyes (approved):

PR #24, opened 2021-12-07
chore: remove fonts.gstatic.com from CSP

ReeceHoffmann:
Changes:
 - removes fonts.gstatic.com from the CSP headers
 - fixes /api rewrite regex to allow the root of the `virtool` API to be called


igboyes (changes requested):
Please update commit body to mention change to API routing and why.

PR #26, opened 2021-12-13
Use createAction for all actions

ReeceHoffmann:
Fairly large refactor of the client state management that changes all actions over to use createAction from redux toolkit.

### DIFF
--- a/src/js/account/sagas.js
+++ b/src/js/account/sagas.js
@@ -1,4 +1,4 @@
-import { put, takeEvery, takeLatest } from "redux-saga/effects";
+import { takeEvery, takeLatest } from "redux-saga/effects";
 import {
     CHANGE_ACCOUNT_PASSWORD,
     CREATE_API_KEY,
@@ -49,7 +49,7 @@ export function* updateAccountSettings(action) {
 
 export function* changeAccountPassword(action) {
     yield apiCall(accountAPI.changePassword, action.payload, CHANGE_ACCOUNT_PASSWORD);
-    yield put({ type: GET_ACCOUNT.REQUESTED });
+    yield getAccount();
 }
 
 export function* getAPIKeys() {
@@ -63,12 +63,12 @@ export function* createAPIKey(action) {
 
 export function* updateAPIKey(action) {
     yield apiCall(accountAPI.updateAPIKey, action.payload, UPDATE_API_KEY);
-    yield put({ type: GET_API_KEYS.REQUESTED });
+    yield getAPIKeys();
 }
 
 export function* removeAPIKey(action) {
     yield apiCall(accountAPI.removeAPIKey, action.payload, REMOVE_API_KEY);
-    yield put({ type: GET_API_KEYS.REQUESTED });
+    yield getAPIKeys();
 }
 
 export function* login(action) {
--- a/src/js/administration/actions.js
+++ b/src/js/administration/actions.js
@@ -1,5 +1,5 @@
-import { simpleActionCreator } from "../utils/utils";
 import { GET_SETTINGS, UPDATE_SETTINGS } from "../app/actionTypes";
+import { createAction } from "@reduxjs/toolkit";
 
 /**
  * Returns action that can trigger an API call for retrieving settings.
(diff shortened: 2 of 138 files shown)


src/js/account/sagas.js
```diff
@@ -1,4 +1,4 @@
-import { put, takeEvery, takeLatest } from "redux-saga/effects";
```
igboyes, line 1: Why are you bypassing `put` here?

The danger of this is you are not dispatching `GET_ACCOUNT.REQUESTED`, which means you cannot handle that action more broadly than what is done in the saga.

src/js/administration/actions.js
```diff
@@ -30,7 +30,6 @@ export const updateSetting = (key, value) => {
  * @param update {object} update data of key-value pairs
  * @returns {object}
  */
-export const updateSettings = update => ({
-    type: UPDATE_SETTINGS.REQUESTED,
-    update
-});
+export const updateSettings = createAction(UPDATE_SETTINGS.REQUESTED, update => ({
+    payload: { update }
+}));
```
igboyes, line 35: In cases like these, I think the following is better:
```js
payload: update
```
The `update` object is not really necessary anymore.

I think cleaning this up should be a low priority issue for the future. Let me know what you think.

src/js/otus/sagas.js
```diff
@@ -19,59 +19,61 @@ import {
...
     } catch (err) {
         yield putGenericError(actionType, err);
         response = err.response;
     }
 
-    const getResponse = yield otusAPI.get(action);
-    yield put({ type: actionType.SUCCEEDED, data: getResponse.body });
+    const getResponse = yield otusAPI.get(action.payload);
+    const curAction = createAction(actionType.SUCCEEDED);
+    yield put(curAction(getResponse.body));
```
igboyes, line 38: Stick to complete words for variable names: `currentAction`.

Also, it seems a bit heavy to use `createAction` here as it is used only to create the action object in the scope of this generator.

src/js/otus/sagas.js
```diff
@@ -109,22 +111,18 @@ export function* removeSequence(action) {
...
         } else {
-            const otuResponse = yield otusAPI.get(action);
-            const historyResponse = yield otusAPI.getHistory(action);
-            yield put({
-                type: REVERT.SUCCEEDED,
-                data: otuResponse.body,
-                history: historyResponse.body
-            });
+            const otuResponse = yield otusAPI.get(action.payload);
+            const historyResponse = yield otusAPI.getHistory(action.payload);
```
igboyes: Could we use [`all`](https://redux-saga.js.org/docs/api#alleffects---parallel-effects) here?

igboyes (changes requested):

src/js/account/sagas.js
ReeceHoffmann: That's a really good point. Oddly enough `export function* updateAPIKey(action) ` already bypasses `put` in the same manner I did here. I am going to convert that over to using a standard redux action as well, unless there is some special reason that this function should bypass the normal redux pathway.

src/js/administration/actions.js
ReeceHoffmann: Agreed. There are a few occasions that the createAction is expecting to get a single object where we could avoid nested objects. Not sure for the specific case of update, but for some of the cases it can get complicated by the fact that the object can be passed by itself or with another variable depending on what the action type is. Not the end of the world but it could add some confusion in the reducer if update can be referred to by `action.payload` or `action.payload.update` depending on the action type. Definitely still worth looking into though.

src/js/otus/sagas.js
ReeceHoffmann: My bad on the name, noted for the future. I don't disagree on createAction being heavy, and I had the same thought while doing it, but I was going for complete conversion to createAction. For now I have reverted to the old way using custom actions. We could still use createAction by defining success action for each in `actions.js` and then passing it to `updateAndGetOTU` but it is a lot less elegant than the current implementation. Let me know which way you want to go, happy with either.

src/js/otus/sagas.js
ReeceHoffmann: Neat, looks like yes. Should be a decent time save

src/js/administration/actions.js
igboyes: I've created low priority issue for this.

src/js/otus/sagas.js
igboyes: Let's keep the reverted form. Your suggestion is probably the most desirable solution, but it is not a high priority at the moment.

igboyes (approved):

PR #31, opened 2021-12-15
Show attribution for subtractions

ReeceHoffmann:
Note: Created as draft PR since created_at object name is not yet guaranteed 

Changes:

- Created new SubtractionAttributions component for handling the logic needed to accommodate various levels of available user information. May be worth moving into attribution, but given that only subtraction components need this logic right now it has been kept separate to reduce load on long list renders.
- Added NameAttribution as a new attribution type for handling cases where we can't or don't want to present time information
- Updated Subtraction list to render different attribution styles depending on what information is present. Displays a message if both user and time are missing.
-  Updated detail to render different attribution styles depending on what information is present. Shows no attribution if both user and time are missing.

### DIFF
--- a/src/js/subtraction/components/Attribution.js
+++ b/src/js/subtraction/components/Attribution.js
@@ -1,0 +1,21 @@
+import { Attribution, NameAttribution } from "../../base";
+import React from "react";
+import { theme, getColor, getFontSize } from "../../app/theme";
+import styled from "styled-components";
+const StyledNoneFoundAttribution = styled.div`
+    color: ${getColor({ color: "grey", theme })};
+    font-size: ${getFontSize("sm")};
+    font-style: italic;
+    min-height: 20px;
+    top: 50%;
+`;
+
+export function SubtractionAttribution({ handle, time }) {
+    if (handle) {
+        if (time) {
+            return <Attribution user={handle} time={time} />;
+        }
+        return <NameAttribution user={handle} />;
+    }
+    return <StyledNoneFoundAttribution> Creator and timestamp unavailable</StyledNoneFoundAttribution>;
+}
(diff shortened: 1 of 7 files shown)


src/js/subtraction/components/Attribution.js
```diff
@@ -0,0 +1,11 @@
+import { Attribution, NameAttribution } from "../../base";
+import React from "react";
+
+export function SubtractionAttribution({ handle, time }) {
+    if (handle && time) {
+        return <Attribution user={handle} time={time} />;
+    } else if (handle) {
+        return <NameAttribution user={handle} time={time} />;
+    }
```
igboyes:
```suggestion
    if (handle) {
            if (time) {
                return <Attribution user={handle} time={time} />;
            }
            
            return <NameAttribution user={handle} time={time} />;
    }
```
Also, `NameAttribution` doesn't take a `time` prop.

igboyes (changes requested):
One minor change.

Also, are all subtraction items going to have the same height, regardless of having the attributions props?

src/js/subtraction/components/Attribution.js
ReeceHoffmann: Nice, small efficiency bump, nice catch. Also yes on the same size thing. I was worried the `InitialIcon` might change the size but the box model reports that either way the flexbox is 42.85 pixels tall. I could probably also hardcode it, but I can't say that I see a point.

Edit: Here is what the subtraction list view looks like with a mix of user information availability
image

ReeceHoffmann: Updated the look of the no data available below:
image

igboyes (approved):

PR #42, opened 2021-12-21
migrate quick analysis logic to formik

ReeceHoffmann:
Changes:
 - Migrates the logic from the bespoke form to formik.
   - Also fixes crashes that could occur if subtractions existed and when the form was submitted.
 - Mild style change that includes white space needed to display error message prior to the message popping, thereby preventing components of the page from noticeably shifting if an error is detected upon submission

### DIFF
--- a/src/js/analyses/components/Create/Quick.js
+++ b/src/js/analyses/components/Create/Quick.js
@@ -1,10 +1,21 @@
-import { forEach, get } from "lodash-es";
+import { forEach } from "lodash-es";
 import React, { useEffect } from "react";
 import { connect } from "react-redux";
 import styled from "styled-components";
 import { getAccountId } from "../../../account/selectors";
 import { pushState } from "../../../app/actions";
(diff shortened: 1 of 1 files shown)


src/js/analyses/components/Create/Quick.js
```diff
@@ -34,6 +46,22 @@ const QuickAnalyzeSelected = styled.span`
     margin: 0 15px 0 auto;
 `;
 
+const StyledError = styled.div`
+    margin -5px 0px 10px
+`;
```
igboyes: `QuickAnalyzeError`.

Only use the `Styled` prefix when you are using it as the top-level element in your component (eg. `StyledQuickAnalyze`).

The preferred solution here is actually to use:
```js
const QuickAnalyzeError = Styled(InputError)``;
```
And redefine its margin accordingly. You are adding an extra `div` here.

igboyes (changes requested):

src/js/analyses/components/Create/Quick.js
ReeceHoffmann: Ahh right, good point. Changed.

PR #50, opened 2022-01-06
Move multiselect sample options into sidebar

ReeceHoffmann:
Changes
- Removes large box when in sample multi select view. 
- Removes filter components from sidebar in multi select view.

Screenshot:
image

(Note: regular sample view remains unchanged)

### DIFF
--- a/src/js/base/SideBar.js
+++ b/src/js/base/SideBar.js
@@ -1,6 +1,6 @@
 import styled from "styled-components";
-import { Box } from "./Box";
 import { borderRadius, boxShadow, fontWeight, getFontSize } from "../app/theme";
+import { Box } from "./Box";
 
 export const SideBarSection = styled(Box)`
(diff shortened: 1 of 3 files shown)


igboyes (changes requested):
1. Commented out code cannot be included in PRs.
2. We discussed disabling filtering during multi-selection. This is still the way to go.
   * What happens when a user selects a sample, then it is hidden in a subsequent filtering operation? I don't see any code to handle this in the PR.
   * Locking the user into dealing only with the visible samples when selecting is easier for us and the user.
   * We can eliminate visual clutter by implementing this big context shift between filtering and selecting. Remind me to show you an example in GMail.
3. The filters are going to be used a lot so they should not be hidden behind collapse effects.
4. The selection clearing button is important and is in a frustrating position.

igboyes (approved):

PR #52, opened 2022-01-07
add file list to index detail view

ReeceHoffmann:
New index detail view shown below.

image

Updated image 10/01/22: Changed wording to requested

### DIFF
--- a/src/js/indexes/components/Files.js
+++ b/src/js/indexes/components/Files.js
@@ -1,0 +1,27 @@
+import { map } from "lodash-es";
+import React from "react";
+import { connect } from "react-redux";
+import { Badge, BoxGroup, BoxGroupHeader } from "../../base";
+import { File as IndexFile } from "../../subtraction/components/Detail/File";
+
+export const Files = ({ files }) => {
+    const fileComponents = map(files, file => <IndexFile file={file} key={file.id} />);
+    return (
+        <BoxGroup>
+            <BoxGroupHeader>
+                <h2>
+                    Files
+                    <Badge>{files.length}</Badge>
+                </h2>
+                Data files available to workflows using this index.
+            </BoxGroupHeader>
+            {fileComponents}
+        </BoxGroup>
+    );
+};
+
+const mapStateToProps = state => ({
+    files: state.indexes.detail.files
+});
+
+export default connect(mapStateToProps)(Files);
--- a/src/js/indexes/components/General.js
+++ b/src/js/indexes/components/General.js
@@ -1,12 +1,14 @@
 import React from "react";
 import Contributors from "./Contributors";
 import IndexOTUs from "./OTUs";
+import Files from "./Files";
 
 export const IndexGeneral = () => (
-    <div>
+    <>
         <Contributors />
+        <Files />
         <IndexOTUs />
-    </div>
+    </>
 );
 
 export default IndexGeneral;
--- a/src/js/indexes/components/__tests__/Files.test.js
+++ b/src/js/indexes/components/__tests__/Files.test.js
@@ -1,0 +1,41 @@
+import { Files } from "../Files";
+import { screen } from "@testing-library/react";
+
+describe("<IndexDetail />", () => {
+    let props;
+
+    beforeEach(() => {
+        props = {
+            files: [
+                { id: "1", name: "foo", download_url: "https://Virtool.ca/testUrl/foo", size: "1024" },
+
+                { id: "2", name: "bar", download_url: "https://Virtool.ca/testUrl/bar", size: "2048" }
+            ]
+        };
+    });
+
+    it("should render", () => {
+        const wrapper = shallow(<Files {...props} />);
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    it("should show component", () => {
+        renderWithProviders(<Files {...props} />);
+        expect(screen.getByText("Files")).toBeInTheDocument();
+    });
+
+    it("should render names", () => {
+        renderWithProviders(<Files {...props} />);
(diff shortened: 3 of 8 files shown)


igboyes (changes requested):
> Data files available to workflows using this index.

src/js/indexes/components/Files.js
```diff
@@ -0,0 +1,33 @@
+import { map } from "lodash-es";
+import React, { useMemo } from "react";
+import { connect } from "react-redux";
+import { Badge, BoxGroup, BoxGroupHeader } from "../../base";
+import { File } from "./File";
+
+export function Files({ files }) {
+    const fileComponents = useMemo(
+        () =>
+            map(files, file => (
```
igboyes: Prefer object spread here.
```js
map(files, ({ name, type }) => ())
```

src/js/indexes/components/General.js
```diff
@@ -1,10 +1,12 @@
 import React from "react";
 import Contributors from "./Contributors";
 import IndexOTUs from "./OTUs";
+import Files from "./Files";
 
 export const IndexGeneral = () => (
     <div>
```
igboyes: Can we use a fragment here instead of `div`?

src/js/indexes/components/File.js
```diff
@@ -0,0 +1,22 @@
...
+const StyledFile = styled(BoxGroupSection)`
+    display: flex;
+`;
+
+const StyledSize = styled.div`
+    margin-left: auto;
+    font-weight: ${getFontWeight("bold")};
+`;
+
+export function File({ fileName, download_url, id, size }) {
```
igboyes: Looks like there is no universal preference for declarations vs arrow functions. However, let's be consistent within Virtool. Almost everything is an arrow function right now, so let's stick with that.

src/js/indexes/components/Files.js
```diff
@@ -0,0 +1,33 @@
+import { map } from "lodash-es";
+import React, { useMemo } from "react";
+import { connect } from "react-redux";
+import { Badge, BoxGroup, BoxGroupHeader } from "../../base";
+import { File } from "./File";
+
+export function Files({ files }) {
```
igboyes: Needs a test.

src/js/indexes/components/Files.js
```diff
@@ -0,0 +1,33 @@
...
+    const fileComponents = useMemo(
```
igboyes: Use of `useMemo` is not advisable here.

1. Unnecessary re-render will already be prevented by `react-redux`. No render occurs unless the values returned from `mapStateToProps` change.
2. Conventional React wisdom: optimize when required, not by deefault. There will be little benefit to `useMemo` here even without `react-redux`. The added complexity and memory usage is not worth it.

I didn't know `useMemo` could return components. Good to know!

igboyes (changes requested):

src/js/indexes/components/Files.js
ReeceHoffmann: Good points, changed.

src/js/indexes/components/File.js
```diff
@@ -0,0 +1,22 @@
...
+    font-weight: ${getFontWeight("bold")};
+`;
+
+export const File = ({ fileName, download_url, size }) => {
+    return (
+        <StyledFile>
+            <a href={download_url}>{fileName}</a> <StyledSize>{byteSize(size)}</StyledSize>
+        </StyledFile>
+    );
+};
```
igboyes:
```suggestion
export const File = ({ fileName, download_url, size }) => (
    <StyledFile>
        <a href={download_url}>{fileName}</a> <StyledSize>{byteSize(size)}</StyledSize>
    </StyledFile>
);
```

src/js/indexes/components/__tests__/Files.test.js
```diff
@@ -0,0 +1,36 @@
...
+        renderWithProviders(<Files {...props} />);
+        expect(screen.getByText("Files")).toBeInTheDocument();
+    });
+
+    it("Render Names", () => {
+        renderWithProviders(<Files {...props} />);
+        expect(screen.getByText("foo")).toBeInTheDocument();
+    });
+
+    it("Render sizes", () => {
```
igboyes:
```suggestion
    it("should render Names", () => {
        renderWithProviders(<Files {...props} />);
        expect(screen.getByText("foo")).toBeInTheDocument();
    });

    it("should render sizes", () => {

```

src/js/indexes/components/__tests__/Files.test.js
```diff
@@ -0,0 +1,36 @@
+import { Files } from "../Files";
+import { screen } from "@testing-library/react";
+
+describe("<IndexDetail />", () => {
+    let props;
+
+    beforeEach(() => {
+        props = {
+            files: [
+                { id: "1", name: "foo", DownloadUrl: "https://Virtool.ca/testUrl/foo", size: "1024" },
```
igboyes: Is `DownloadUrl` correct?

src/js/indexes/components/File.js
```diff
@@ -0,0 +1,22 @@
...
+const StyledFile = styled(BoxGroupSection)`
+    display: flex;
+`;
+
+const StyledSize = styled.div`
+    margin-left: auto;
+    font-weight: ${getFontWeight("bold")};
+`;
+
+export const File = ({ fileName, download_url, size }) => {
```
igboyes: I think there is opportunity to share this with the file list in the subtraction detail view. Can you look at doing this?

igboyes (changes requested):

igboyes (approved):

PR #54, opened 2022-01-10
Prevent sample colours in label managment modal from turning white

ReeceHoffmann:
Appears to be a rendering bug in chromium. Adding a transform to the non-hover state prevents the loss of color. Working in chrome version 97.0.4692.71, and edge version 97.0.1072.55.


igboyes (commented):
Is `-webkit` prefix necessary here?

The vendor prefixes should be [applied automatically by `styled-components`](https://styled-components.com/docs/basics#motivation).

PR #56, opened 2022-01-13
Disable sample creation sidebar dropdowns when no options available

ReeceHoffmann:
Images below for design proofing.

### **Outdated: see comment below for updated images**

No subtractions or labels available: 
image

Subtractions and labels available:
image

### DIFF
--- a/src/js/samples/components/Create/__tests__/Create.test.js
+++ b/src/js/samples/components/Create/__tests__/Create.test.js
@@ -4,6 +4,16 @@ import React from "react";
 import { createStore } from "redux";
 import { LIST_LABELS } from "../../../../app/actionTypes";
 import { CreateSample, mapDispatchToProps, mapStateToProps } from "../Create";
+import { BrowserRouter } from "react-router-dom";
+
+const routerRenderWithProviders = (ui, store) => {
+    const routerUi = <BrowserRouter> {ui} </BrowserRouter>;
+    return renderWithProviders(routerUi, store);
+};
+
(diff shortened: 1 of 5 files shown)


igboyes (changes requested):
I like, but a few changes to improve UX:

1. Just hide the pencil button when there are no labels or subtractions.
    * No point of comparison for what an un-disabled button looks like if both are disabled.
    * I have found disabling elements is not as clear to users as just hiding them.
    * Easier to test. Presence or absence instead of HTML attribute.
2. I like how **Manage** looks when no labels or subtractions are available. A user is likely going to want a shortcut to changing that situation.

    However, when there are some present, **Manage** could become confusing. Does it refer to labels or sample-label assignments.

    I think it may be better to stick the manage button in the popover in this case. Thoughts?

ReeceHoffmann:
1. Sure, I think what your saying makes sense. I can see there being ambiguity if both are disabled at the same time.
2. I get what your saying, but I do worry about moving a single piece of ui to a different location depending on application state being confusing. That said, I do think having it in the drop down list makes a lot more sense, I will experiment and see what I can come up with.

igboyes:
For 2, you could use different link text.

It might be more obvious to have **Create one** when no labels exist.

ReeceHoffmann:
Updated look

No labels:
image

Samples exist:
image

Dropdown:
image


Dropdown close-up:
image

igboyes (approved):

PR #57, opened 2022-01-17
clarify file upload status in ui

ReeceHoffmann:
Changes:
 -  Removal of an item from the upload bar now occurs only when the file is ready.
 - When item is fully uploaded but not yet ready the upload bar displays a spinner instead of the the upload icon
 - File list is only updated with the new upload once the file is marked as ready

Screenshot of file manager when the file is done uploading but not ready:

image


igboyes (approved):
Looks great! Thanks.

PR #59, opened 2022-01-21
feat: add multiselect label managment to sample view

ReeceHoffmann:
**Changelog:**

Multiselect version of sample list view now permits label management for multiple samples at a time
 - Selecting a label will attempt to apply the label to all selected samples unless all select samples already have the label applied
 - selecting a label that is already applied to all selected labels will result in the label being cleared from all samples
 -  dropdown and label list indicate which labels are fully selected vs partly selected vs not selected. 
 - dropdown allows addition or removal of labels.

Images updated: 28/01/2022
Nothing selected:
image

**Multiselect no dropdown:**
image

**Multiselect dropdown:**
image

### DIFF
--- a/src/js/samples/components/Sidebar/List.js
+++ b/src/js/samples/components/Sidebar/List.js
@@ -1,6 +1,7 @@
 import React from "react";
 import styled from "styled-components";
-import { SampleLabel } from "../Label";
+import { SampleLabel, SampleMultiSelectLabel } from "../Label";
+import { map } from "lodash-es";
 
 const SampleSidebarListItem = styled(SampleLabel)`
     background-color: ${props => props.theme.color.white};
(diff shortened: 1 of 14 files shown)


src/js/samples/components/Sidebar/List.js
```diff
@@ -20,3 +21,17 @@ export const SampleSidebarList = ({ items }) => {
 
     return <StyledSampleSidebarList>{sampleItemComponents}</StyledSampleSidebarList>;
 };
+
+export const SampleSidebarMultiselectList = ({ items }) => {
+    const sampleItemComponents = map(items, item => (
```
igboyes: Unpack the object here:
```js
({ id, color, name, allLabeled }) => {}
```

src/js/samples/components/Sidebar/ManageLabels.js
```diff
@@ -0,0 +1,120 @@
...
+`;
+
+const StyledSideBarSection = styled(SideBarSection)`
+    grid-row: 2;
+    align-self: start;
+`;
+
+const getSelectedSamples = (selected, documents) => {
+    return filter(documents, document => selected.includes(document.id));
+};
```
igboyes: This is not efficient as you are calling `includes` for every `document`. This will bog down with larger sizes of `selected` and is a good example for thinking in terms of big O notation.

There is probably a function in Lodash that will do this for you much more efficiently.

igboyes (changes requested):
This is looking slick. I am noticing the following visual glitches:

* The shift in the heading of the labels sidebar component when shifting between the normal and selection context is a bit jarring. Can we eliminate that?

  It would be nice to do a universal fix that covers all sidebar components. I am fine considering a change in the pencil icon so that it takes up less vertical space.
* When changing from a check to nothing (all-selected to none-selected), the minus sign flashes for a split second.
* Hovering over the color dot on a label in the sidebar makes it turn black.

src/js/samples/components/Sidebar/ManageLabels.js
ReeceHoffmann: Turns out there was an existing `getSelectedSamples` in `samples/selectors.js` with identical implementation. I removed my own and replaced the code in the existing one with a version that should run faster.

src/js/samples/components/Sidebar/List.js
ReeceHoffmann: done

ReeceHoffmann:
Update 28/01/2022:

**Changes since last push:**
 - moved logic out of the `ManageLabels` component as much as possible
 - optimized the logic of `getSelectedSamples` and `getSelectedLabels`
 - fixed most known visual bugs 
 - added/updated tests as needed to accommodate the changes

igboyes (approved):
I am happy with this. Have a good weekend!

PR #61, opened 2022-01-21
Align SampleItem anaylsis view tags

ReeceHoffmann:
**Update: Fixed border**
New look with view on left:
image


igboyes (changes requested):
image

Thick borders are awkward.

1. No right border on rightmost tag probably would be best. See `:last-child` if you haven't already.
2. Border should just be normal thickness on _No Analyses_. Too thick here.

igboyes (dismissed):

igboyes (approved):

PR #62, opened 2022-01-22
prevent ScrollList page request duplication

ReeceHoffmann:
Prevents page request spam by having `ScrollList` remember the last page requested and prevent sending a repeat request if it matches the current page being requested.

Alternative solution would be to throttle via Sagas or Lodash, but both of those solutions would permit duplication in the case that the user had a very slow connection and/or potentially slow down legitimate scrolling.


igboyes (changes requested):
Focus on user impact of commit:
```
fix: prevent excess requests and lock-up during infinite scroll
```
User-facing release notes are generated from the `fix` and `feat` commit messages.

igboyes (approved):

PR #66, opened 2022-02-01
update pathoscope analysis export links

ReeceHoffmann:
**Changes:**
 - Updates export links for the pathoscope analysis view
 - Wraps export dropdown button children in span to prevent flexbox removing spaces between icons and text.

Wrapped with span:
image

Unwrapped:
image


igboyes (approved):
Thanks for attention to detail.

PR #71, opened 2022-02-07
feat: display and permit removal of failed uploads in upload overlay

ReeceHoffmann:
**Changes:**
 - Adds an error state to upload items and displays a new UploadItemError component for any upload that has failed
   - New component is visually distinct from UploadItem and add a button to remove the failed upload.
 - Modifies `base/ProgressBar` to allow colour changes in firefox as well as chrome
 - Adds test for new components using ReactTestingLibrary
 -  modifies `putGenericError` to prevent a client side crash in the event that the server error does not have a body or status 

Update: 16/02/2022

Updated overlay:
image

### DIFF
--- a/src/js/base/ProgressBar.js
+++ b/src/js/base/ProgressBar.js
@@ -13,6 +13,9 @@ const StyledProgress = styled.progress`
     ::-webkit-progress-value {
         background-color: ${getProgressColor};
     }
+    ::-moz-progress-bar {
+        background-color: ${getProgressColor};
+    }
 
     ::-webkit-progress-bar {
         background-color: ${props => props.theme.color.grey};
--- a/src/js/files/components/UploadItem.js
+++ b/src/js/files/components/UploadItem.js
@@ -1,8 +1,10 @@
 import React from "react";
 import styled from "styled-components";
-import { getBorder } from "../../app/theme";
+import { getBorder, getColor, getFontWeight } from "../../app/theme";
 import { AffixedProgressBar, Icon, Loader } from "../../base";
 import { byteSize } from "../../utils/utils";
+import { connect } from "react-redux";
+import { removeUpload } from "../actions";
 
 const StyledUploadItem = styled.div`
     padding: 0;
@@ -21,21 +23,52 @@ const UploadItemTitle = styled.div`
     i.fas {
         margin-right: 5px;
     }
-    div:first-child {
-        margin-right: 5px;
-    }
     span:last-child {
         margin-left: auto;
+        color: ${props => (props.failed ? getColor({ theme: props.theme, color: "red" }) : "inherit")};
+    }
+    i.fa-times {
+        font-size: 20px;
+    }
+    i.fa-trash {
+        margin-left: 5px;
+        font-size: 14px;
     }
 `;
 
-export const UploadItem = ({ name, progress, size }) => (
-    <StyledUploadItem>
-        <AffixedProgressBar now={progress} />
-        <UploadItemTitle>
-            {progress === 100 ? <Loader size="14px" /> : <Icon name="upload" />}
-            <span>{name}</span>
-            <span>{byteSize(size)}</span>
-        </UploadItemTitle>
-    </StyledUploadItem>
-);
+const StyledFilename = styled.span`
+    font-weight: ${getFontWeight("thick")};
+`;
+
+export const UploadItem = ({ name, progress, size, failed, localId, onRemove }) => {
+    let uploadIcon = progress === 100 ? <Loader size="14px" /> : <Icon name="upload" />;
+    let uploadBookend = byteSize(size);
+
+    if (failed) {
+        uploadIcon = <Icon name="times" color={"red"} hoverable={false} />;
+        uploadBookend = (
+            <>
+                Failed <Icon aria-label={`delete ${name}`} name="trash" color="red" onClick={() => onRemove(localId)} />
+            </>
+        );
+    }
+
+    return (
+        <StyledUploadItem>
+            <AffixedProgressBar now={failed ? 100 : progress} color={failed ? "red" : "blue"} />
+            <UploadItemTitle failed={failed}>
+                {uploadIcon}
+                <StyledFilename>{name}</StyledFilename>
+                <span>{uploadBookend}</span>
+            </UploadItemTitle>
+        </StyledUploadItem>
+    );
+};
+
+const mapDispatchToProps = dispatch => ({
+    onRemove: localId => {
+        dispatch(removeUpload(localId));
+    }
+});
+
+export default connect(null, mapDispatchToProps)(UploadItem);
--- a/src/js/files/reducer.js
+++ b/src/js/files/reducer.js
@@ -4,12 +4,14 @@
  * @module files/reducer
  */
 import { createReducer } from "@reduxjs/toolkit";
-import { map, reject } from "lodash-es";
+import { forEach, map, reject } from "lodash-es";
 import {
     FIND_FILES,
     UPLOAD,
     UPLOAD_PROGRESS,
+    UPLOAD_FAILED,
     UPLOAD_SAMPLE_FILE,
+    REMOVE_UPLOAD,
     WS_INSERT_FILE,
     WS_REMOVE_FILE,
     WS_UPDATE_FILE
@@ -37,7 +39,7 @@ export const appendUpload = (state, action) => {
 
     return {
         ...state,
-        uploads: state.uploads.concat([{ localId, progress: 0, name, size, type: fileType, context }])
+        uploads: state.uploads.concat([{ localId, progress: 0, name, size, type: fileType, context, failed: false }])
     };
 };
 
@@ -102,6 +104,16 @@ export const filesReducer = createReducer(initialState, builder => {
         })
         .addCase(UPLOAD_PROGRESS, (state, action) => {
             return updateProgress(state, action.payload);
+        })
+        .addCase(UPLOAD_FAILED, (state, action) => {
+            forEach(state.uploads, upload => {
+                if (upload.localId === action.payload.localId) {
+                    upload.failed = true;
+                }
+            });
+        })
+        .addCase(REMOVE_UPLOAD, (state, action) => {
+            state.uploads = reject(state.uploads, { localId: action.payload.localId });
         });
 });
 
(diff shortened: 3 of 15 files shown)


src/js/files/reducer.js
```diff
@@ -102,6 +104,16 @@ export const filesReducer = createReducer(initialState, builder => {
...
         .addCase(UPLOAD_PROGRESS, (state, action) => {
             return updateProgress(state, action.payload);
+        })
+        .addCase(UPLOAD_FAILED, (state, action) => {
+            forEach(state.uploads, upload => {
+                if (upload.localId === action.payload.localId) {
+                    upload.failed = true;
+                }
+            });
+        })
```
igboyes, line 114: Mutates state. Use object spread operator and `map`.

src/js/base/ProgressBar.js
```diff
@@ -13,6 +13,9 @@ const StyledProgress = styled.progress`
     ::-webkit-progress-value {
         background-color: ${getProgressColor};
     }
+    ::-moz-progress-bar {
+        background-color: ${getProgressColor};
+    }
 
     ::-webkit-progress-bar {
         background-color: ${props => props.theme.color.grey};
```
igboyes, line 21: Can you check if just `::progress-bar` works?

src/js/files/components/UploadItem.js
```diff
@@ -39,3 +41,42 @@ export const UploadItem = ({ name, progress, size }) => (
...
+    </StyledUploadItemError>
+);
+
+const mapDispatchToProps = dispatch => {
+    return {
+        onRemove: localId => {
+            dispatch(removeUpload(localId));
+        }
+    };
+};
```
igboyes:
```suggestion
const mapDispatchToProps = dispatch => ({
    onRemove: localId => {
        dispatch(removeUpload(localId));
    }
});
```

src/js/files/components/UploadItem.js
```diff
@@ -39,3 +41,42 @@ export const UploadItem = ({ name, progress, size }) => (
...
+    <StyledUploadItemError>
+        <AffixedProgressBar now={100} color="red" />
+        <UploadItemTitle>
+            {<Icon name="times" />}
+            {name}
+            <span>
+                Failed <Icon aria-label={`delete ${name}`} name="trash" color="red" onClick={() => onRemove(localId)} />
+            </span>
+        </UploadItemTitle>
+    </StyledUploadItemError>
```
igboyes: Combine this with `UploadItem`. What was the strategy around having separate components for this?

src/js/files/components/UploadItem.js
```diff
@@ -39,3 +41,42 @@ export const UploadItem = ({ name, progress, size }) => (
...
+`;
+
+export const UploadItemError = ({ name, onRemove, localId }) => (
+    <StyledUploadItemError>
+        <AffixedProgressBar now={100} color="red" />
+        <UploadItemTitle>
+            {<Icon name="times" />}
+            {name}
+            <span>
+                Failed <Icon aria-label={`delete ${name}`} name="trash" color="red" onClick={() => onRemove(localId)} />
```
igboyes, line 51: Prime use case for `useCallback`.

src/js/files/components/UploadItem.js
```diff
@@ -39,3 +41,42 @@ export const UploadItem = ({ name, progress, size }) => (
...
+
+const mapDispatchToProps = dispatch => {
+    return {
+        onRemove: localId => {
+            dispatch(removeUpload(localId));
+        }
+    };
+};
+
+export const ConnectedUploadItemError = connect(null, mapDispatchToProps)(UploadItemError);
```
igboyes: Combine upload item implementations into single connected component and:
```js
export default connect(mapDispatchToProps)(UploadItem);
```

igboyes (changes requested):

src/js/base/ProgressBar.js
ReeceHoffmann: Sadly it doesn't seem so, I tried it, and checked documentation, and it looks like customizing the `<progress>` element is currently left to vendor specific css pseudo classes.

src/js/files/components/UploadItem.js
ReeceHoffmann: fixed

src/js/files/components/UploadItem.js
ReeceHoffmann: When implementing I figured that two simple components would be easier to understand and possibly faster than one monolithic component. I have uploaded a new version with just one UploadItem, so you can make the call on which you think is better.

src/js/files/components/UploadItem.js
ReeceHoffmann: The two are now combined per earlier suggestion. Also per the suggest line of code, I assume the main point was the removal of the null positional argument from: 

```javascript 
export default connect(null, mapDispatchToProps)(UploadItem);
```
As annoying as it looks, the null is there intentionally so that the first positional argument slot is occupied and `mapDispatchToProps` gets handed dispatch as expected. Written the suggested way, `mapDispatchToProps` is handed state.

src/js/files/components/UploadItem.js
ReeceHoffmann: Hmm, honestly not convinced of the performance benefit here. My understanding of `useCallback` is that is best used in cases where referential equality matters. I could perhaps see it making a difference when react is diffing the dom tree, but most of the examples I have seen talking about efficient `useCallback` usage primarily suggested reserving it for cases where you needed the referential equality for proper functioning (ex lodash throttle) or to save massive downstream re-renders. (ex. when the function is a dependency for a large `useMemo`)

Icon technically does check referential equality when defining its own `useCallback` but looking closer I think this usage is a bit suspect to start:
```javascript
const handleClick = useCallback(props.onClick, [props.onClick]);
```
Since `useCallback` returns the original reference to the function there are two ways this can go. On a new render either the `props.onClick` has the exact same reference as before, and the usage of useCallback was redundant since the returned `handleClick` has the exact same reference as `props.onClick`, or `props.onClick` points to a new reference and useCallback now returns a reference identical to the new `props.onClick`.

Let me know what you think. Happy to change course if I have missed something.

src/js/files/components/UploadItem.js
igboyes: The suggestion is using a default export for the connected component. This is `react-redux` convention.

I forgot to put `null` in.

ReeceHoffmann: Ah makes sense, I had deliberately avoided it in the first push since it seemed strange to export the error variant of `UploadItem` as default. Once merged, I went back to exporting as default.

src/js/files/components/UploadItem.js
igboyes: You're right. I commented on the pattern without thinking to deeply about it.

src/js/files/components/UploadItem.js
```diff
@@ -26,16 +27,51 @@ const UploadItemTitle = styled.div`
...
-    </StyledUploadItem>
-);
+export const UploadItem = ({ name, progress, size, failed, localId, onRemove }) => {
+    let [now, progressColor] = [progress, "blue"];
+    let uploadIcon = progress === 100 ? <Loader size="14px" /> : <Icon name="upload" />;
+    let uploadBookend = byteSize(size);
+
+    if (failed) {
+        [now, progressColor] = [100, "red"];
+        uploadIcon = <Icon name="times" />;
```
igboyes: Use `color` prop of `Icon` instead of styling `i.fa-times`. Is that not working?

src/js/files/components/UploadItem.js
```diff
@@ -26,16 +27,51 @@ const UploadItemTitle = styled.div`
...
-        <AffixedProgressBar now={progress} />
-        <UploadItemTitle>
-            {progress === 100 ? <Loader size="14px" /> : <Icon name="upload" />}
-            <span>{name}</span>
-            <span>{byteSize(size)}</span>
-        </UploadItemTitle>
-    </StyledUploadItem>
-);
+export const UploadItem = ({ name, progress, size, failed, localId, onRemove }) => {
+    let [now, progressColor] = [progress, "blue"];
```
igboyes: Just define one variable per line. I am surprised ESLint is not complaining about this.

Actually, I prefer unary conditionals in the `return` statement for simple value switching like this.
```js
<AffixedProgressBar now={failed ? 100 : progress} color={failed ? "red" : "blue"} />
```

src/js/files/components/UploadItem.js
igboyes: I appreciate the abstraction offered by a single `UploadItem` component. You just pass it props from the upload object and it knows what to do.

src/js/files/components/UploadItem.js
```diff
@@ -26,16 +27,51 @@ const UploadItemTitle = styled.div`
     }
     span:last-child {
         margin-left: auto;
+        color: ${props => (props.failed ? getColor({ theme: props.theme, color: "red" }) : "inherit")};
+        font-size: ${props => (props.failed ? getFontSize("sm") : "inherit")};
+        font-style: ${props => (props.failed ? "italic" : "normal")};
```
igboyes: Italics are not used anywhere else in Virtool (AFAIK) and I don't consider it part of the design language.

For this, why don't we do the following to update the styling:

1. Use thick font-weight for the file name.
2. Use regular font-weight, style, and size for **Failed**.
3. Use `lg` font-size and thick weight for the upload overlay title (**Uploads (4)**).

igboyes (changes requested):

src/js/files/components/UploadItem.js
ReeceHoffmann: It works, but when colored by passing props to the `Icon` component the icon will change colors when hovered. Given that the the`fa-times` icon is not interactable I didn't think having a recolor on hover would make sense. I could instead modify the base `Icon` behavior to do something like only recolor on hover if there is a passed function, but that could easily have side effects elsewhere.

src/js/files/components/UploadItem.js
ReeceHoffmann: Changed.

src/js/files/components/UploadItem.js
igboyes: > I could instead modify the base `Icon` behavior to do something like only recolor on hover if there is a passed function

This is how it should work. I think that must have regressed at some point. Let's take the time to fix this problem now.

src/js/files/components/UploadItem.js
ReeceHoffmann: Changed to use your suggested styling.

Updated overlay: 
image

igboyes: Thanks for the screenshot. Looks great and integrated with the rest of the UI.

src/js/files/components/UploadItem.js
ReeceHoffmann: Alright fixed. I updated the icon color change to to recolor if either a) it is passed an onClick or b) hoverable is passed. On a quick inspection, I wasn't able to find any examples in the UI where this caused a button to not change color when it should have.

igboyes (approved):

PR #73, opened 2022-02-12
Revamp reference source type selection

ReeceHoffmann:
### **OUTDATED - see comment below for latest**

Rework of the sourceTypes component to behave more like a form.

No sourceType changes since last save:
image

With sourceType changes:
image


igboyes (changes requested):
1. Remove staging-saving behaviour. Include only the **Add** form.
2. Change **Add Sources** to **Add Source Type**. Remove colon. This is not a UI pattern in VT.
3. I think you already have some of the logic for undoing a previous source type deletion. Maybe we can allow reversion of the immediately previous deletion. I think this is a good compromise between complexity-features.

ReeceHoffmann:
**Outdated** ~~New design.~~ Undo icon allows the reversion of a single add or delete.

Disabled:
image


Enabled, no optional elements showing:
image



Enabled, all optional elements showing:
image

igboyes (changes requested):
1. **Mockup of what I was thinking**
image

  Currently, it is not obvious what is being undone.

2. Please make input and button take up full available width. You can limit the width of the input if looks bad.

ReeceHoffmann:
Latest Version:

Disabled:
image

Enabled, no error or undo:
image

Enabled, error + undo:
image

igboyes (approved):

PR #83, opened 2022-02-24
Improve look of infinite scrolling lists

ReeceHoffmann:
**Changes**

  - Adds padding to bottom of the ScrollList component to create space between bottom of page and last item in the ScrollList.
  -  Removed option to style via passed props to ScrollList. No instances of styling by this method were found in the code base. Support could be added for styling through styled components if this functionality becomes needed.

Jobs screenshot: 

image

### DIFF
--- a/src/js/base/ScrollList.js
+++ b/src/js/base/ScrollList.js
@@ -1,9 +1,14 @@
 import React from "react";
 import { map } from "lodash-es";
 import { LoadingPlaceholder } from "./index";
+import styled from "styled-components";
 
 export const getScrollRatio = () =>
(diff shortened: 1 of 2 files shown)


src/js/base/ScrollList.js
```diff
@@ -37,12 +42,11 @@ export class ScrollList extends React.Component {
         if (documents === null && page < pageCount) {
             loading = <LoadingPlaceholder margin="20px" />;
         }
-
         return (
-            <div style={this.props.style}>
+            <ScrollListContainer>
```
igboyes: Conventional naming for this is `StyledScrollList`.

igboyes (changes requested):

igboyes (approved):

PR #86, opened 2022-02-25
Ensure correct initial icon centering across browsers

ReeceHoffmann:
Changes the initial icon to using a browser rendered SVG rather than a standard div. Provides pixel perfect centering on Ubuntu for both Chrome and Firefox. 

Chrome:
Sample list view:
image
Reference settings:
image

Firefox:
Sample list view:
image
Reference settings:
image

Very close to perfect on windows. Can be off by a pixel if the font is not an even number of pixels tall.

Windows, Chrome: 

Sample list view: 
image

Reference Settings
image


igboyes (approved):
Awesome! Thanks.

PR #91, opened 2022-03-23
Ensure logical behaviour of reference clone view on refresh

ReeceHoffmann:
Updated 25/03/2022

Changes:
 - Moves cloning functionality into modal in the `/refs` view
 - Full conversion of class based logic to Hooks + formik
 - Full conversion of tests from enzyme to RTL
 - Removed ability to edit description when cloning, now defaults to: `Cloned from ${Original_Ref_name}`

Modal screenshot:

image


igboyes (commented):
Thoughts on the following?:
* Hookify components instead of using class-based.
* Get rid of description box. This can be modified after cloning.

ReeceHoffmann:
> Thoughts on the following?:
> 
>     * Hookify components instead of using class-based.
> 
>     * Get rid of description box. This can be modified after cloning.

Point 1: Sure. I didn't change it since it won't improve functionality, but I can convert it to hooks + formik to keep it more consistent with the way that forms are being done now.

Point 2: Yeah sure, I will just have it copy the description of the source. Honestly that kinda makes more sense than the current empty box. I suppose another option could be to keep the description box in clone view and prepopulate it with the source refs description.

igboyes:
1. Let's take the opportunity to update these things when we touch a module. Hooks will work well in some of these components because the class methods are there to do pretty basic stuff.
2. Keep it simple with first option. No user has ever actually entered a description AFAIK.

igboyes (approved):
Awesome!

PR #94, opened 2022-03-30
fix: prevent crash when interacting with reference import view

ReeceHoffmann:
Changes: 

- Converts reference import, empty, and edit view from custom class components to formik function components
- Full conversion of reference import and reference tests from enzyme to rtl
- Adds rtl tests for reference import view
- Fixes uploadOverlay filtering against reference type uploads
- Fixes reference edit view not allowing the organism to be changed
- Removes redundant Organism and dataType Fields from import and clone action/api calls
- Mild visual changes to remove extra gap below Name field in `referenceForm`

Screenshots:

Edit Reference View(with errors): 
image


igboyes (approved):
Great. Big improvements everywhere!

PR #95, opened 2022-04-01
feat: modernize login, password reset, and first user views

ReeceHoffmann:
Updated: 05/04/2022 

**Changes**
 - Convert remaining enzyme tests for login and first user to RTL
 - Add tests for password reset view and B2C logic
 - add Divider base component

**Screenshots**

New Login with b2c (FullScreen)
image

New Login without b2c (FullScreen)
image

New First User (FullScreen)
image

New Reset(FullScreen)
image


src/images/ms_login_logo.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="21" height="21" viewBox="0 0 21 21"><title>MS-SymbolLockup</title><rect x="1" y="1" width="9" height="9" fill="#f25022"/><rect x="1" y="11" width="9" height="9" fill="#00a4ef"/><rect x="11" y="1" width="9" height="9" fill="#7fba00"/><rect x="11" y="11" width="9" height="9" fill="#ffb900"/></svg>
```
igboyes: We are definitely allowed to use this legally?

I was thinking of the Font Awesome MS icon which should be fine. The colored one is nice so just double check we have the right to use the image without attribution etc.

igboyes (commented):
Looks great. One legal question.

src/js/wall/B2cLogin.js
```diff
@@ -0,0 +1,46 @@
+import React from "react";
+import { connect } from "react-redux";
+import styled from "styled-components";
+import { loginSucceeded } from "../account/actions";
+import { getLoginRequest } from "../app/authConfig";
+import { getColor } from "../app/theme";
+import { Button, Divider } from "../base";
+import { WallSubheader } from "./Container";
+
+const StyledB2cLogin = styled.div`
```
igboyes:
```suggestion
const StyledB2CLogin = styled.div`
```

src/js/wall/B2cLogin.js
```diff
@@ -0,0 +1,46 @@
...
+    color: ${props => getColor({ color: "greyDark", theme: props.theme })};
+    margin-left: auto;
+    transform-text: uppercase;
+`;
+
+const StyledDivider = styled(Divider)`
+    margin-bottom: 30px;
+`;
+
+export const B2cLogin = ({ setLogin }) => (
```
igboyes:
```suggestion
export const B2CLogin = ({ setLogin }) => (
```

src/js/wall/B2cLogin.js
```diff
@@ -0,0 +1,46 @@
...
+        </StyledB2cLogin>
+        <StyledDivider text="or" />
+    </>
+);
+
+export const mapDispatchToProps = dispatch => ({
+    setLogin: msal => {
+        window.msalInstance.setActiveAccount(msal.account);
+        dispatch(loginSucceeded());
+    }
```
igboyes: I don't think this is not the appropriate place for this.

Only the call to `dispatch` should be here. I would just define an `onClick` the deals with all the `msal` business and finishes up with a call to the `setLogin` prop.

src/js/wall/B2cLogin.js
igboyes: I may be fuzzy on this, but all Virtool callbacks in props are named like `onSetLogin`. This makes it easy to separate functions from other variables in component definitions.

Pretty sure it is React convention as well.

igboyes (changes requested):

igboyes (approved):

PR #96, opened 2022-04-12
Add support for saving form state when navigating

ReeceHoffmann:
**Update 11/05/2022:** Removes clear form button and adds an alert that renders when the form has been restored from previous data. See screenshots below. `undo` icon resets form to empty and `times` icon closes the alert.

**Update 25/04/2022:** Large rework in how data flows in and out of the component. Setting data is handled through redux-sagas while data is now fetched from redux rather than sessionStorage. (in case of page refresh or other redux state loss data is stored to sessionStorage and restored to redux state on reload)

Changes:

 - Add custom `<PersistForm>` component to base which when added to a form stores form state in local storage and restores it on first render
 - Added tests for new `<PersistForm>` base component
 - Added clear form button base component
- Add `<PersistForm>` component to `Create` view for `Samples` and `Subtractions`
- Added clear form button to Samples and Subtractions
- Added tests for Clear button and refresh persistence for Create Views of Sample and Subtractions

Screenshots:

Sample Create View:
image

Subtraction create view:
image


src/js/base/PersistForm.js
```diff
@@ -0,0 +1,34 @@
...
+
+            const prevFormData = getSessionStorage(formName);
+            if (prevFormData) {
+                const castFormData = castValues ? castValues(getSessionStorage(formName)) : getSessionStorage(formName);
+                formikContext.setValues(castFormData);
+            }
+        } else {
+            debouncedSetSessionStorage(formName, formikContext.values);
+        }
+    });
```
igboyes: You can possibly get rid of `useFirstRender` by splitting this into two hook calls.

Passing an empty array to `useEffect` as the second argument should cause the effect to be run only on first render.

https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

src/js/base/PersistForm.js
```diff
@@ -0,0 +1,34 @@
...
+
+    return null;
+};
+
+export const clearFormStorageItem = key => {
+    setSessionStorage("clearForm", key);
+};
```
igboyes: Should be with your other sessionStorage utilities.

igboyes (changes requested):
I will schedule a call to discuss this.

Since the saved form data can be thought of as application-level data, it could be managed by Redux. I think this would have a lot of benefits for better structuring how data flow works in this PR.

src/js/base/PersistForm.js
ReeceHoffmann: Function deleted with rework

src/js/forms/sagas.js
```diff
@@ -0,0 +1,22 @@
...
+    yield put(updateReduxFormState(formName, formValues));
+    const formState = yield select(getAllFormState);
+    setSessionStorage("formState", formState);
+}
+
+export function* deletePersistentFormState({ payload: { formName } }) {
+    yield put(updateReduxFormState(formName, null));
+    const formState = yield select(getAllFormState);
+    setSessionStorage("formState", formState);
+}
```
igboyes: How about:
1. Pull current state from Redux.
2. Merge update into current state in saga.
3. Update localStorage.
4. Update Redux.

igboyes (changes requested):

igboyes (changes requested):
Please remove the red **Clear Form** button.

Instead, when form state has been loaded from storage, display a grey alert at the top of the view describing the situation (_Resumed editing draft sample_) with a clear button.

This approach makes the situation clearer to the user.

igboyes (approved):

PR #100, opened 2022-05-05
Fix: Update account drop down menu and User not rendered correctly in OTU history issues

elanjm12:


igboyes:
Need to change the following:
* One PR per Linear issue. Way easier to review and deal with test failures.
* `change to repository. original fork incorrect` should not be a commit.
* Use Linear branch names.

The commit messages end up in release notes. They should describe what changes for the end user.

* _correct rendering in OTU history_ -> _render correct user handle in OTU history_
* _minor changes to NavBar_ -> _cleanup navigation bar and account dropdown_

I'm closing this. Please re-submit as two separate PRs with related Linear branch names.

Talk to @ReeceHoffmann about git work if needed.

PR #101, opened 2022-05-06
fix: user not rendered correctly in OTU history

elanjm12:


ReeceHoffmann (changes requested):
Looks pretty good to me but the commit message needs to be modified. The commit message should be written in the imperative and focus on the result than the problem.

 i.e.,  user not rendered correct in OTU history -> render user handle in OTU history. 

Also when making UI changes please include screenshots of the elements that changed.

elanjm12:
Screenshot from 2022-05-06 09-59-43

igboyes (approved):

ReeceHoffmann (approved):

PR #105, opened 2022-05-10
fix: change navbar dropdown icon to InitialIcon

elanjm12:
Change navbar dropdown icon to InitialIcon. Add text description of which user is logged in.

Screenshot from 2022-05-10 12-52-52

### DIFF
--- a/src/js/nav/components/NavBar.js
+++ b/src/js/nav/components/NavBar.js
@@ -3,8 +3,9 @@ import React from "react";
 import { connect } from "react-redux";
 import styled from "styled-components";
 import { logout } from "../../account/actions";
-import { Dropdown, DropdownMenuItem, DropdownMenuLink, DropdownMenuList, Icon, VTLogo } from "../../base";
+import { Dropdown, DropdownMenuItem, DropdownMenuLink, DropdownMenuList, Icon, InitialIcon, VTLogo } from "../../base";
 import { NavBarItem } from "./NavBarItem";
+import { getAccountHandle } from "../../account/selectors";
 
 const NavBarLeft = styled.div`
     display: flex;
@@ -53,7 +54,12 @@ const StyledNavBar = styled.div`
     z-index: 90;
 `;
 
-export const Bar = ({ administrator, dev, userId, onLogout }) => (
+const StyledDivider = styled.div`
+    color: ${props => props.theme.color.greyLight};
+    border-top: 2px solid;
+`;
+
+export const Bar = ({ administrator, dev, userId, onLogout, handle }) => (
     <StyledNavBar>
         <NavBarLeft>
             <NavBarLogo color="white" />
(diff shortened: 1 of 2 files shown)


src/js/nav/components/NavBar.js
```diff
@@ -1,9 +1,18 @@
 import { MenuButton } from "@reach/menu-button";
-import React from "react";
+import React, { useMemo } from "react";
```
ReeceHoffmann: Remove useMemo

src/js/nav/components/NavBar.js
```diff
@@ -74,11 +83,16 @@ export const Bar = ({ administrator, dev, userId, onLogout }) => (
 
             <Dropdown>
                 <NavDropdownButton>
-                    <Icon name="user" />
+                    <InitialIcon handle={handle} size="md"></InitialIcon>
```
ReeceHoffmann:
```suggestion
                    <InitialIcon handle={handle} size="md"/>
```

src/js/nav/components/NavBar.js
```diff
@@ -74,11 +83,16 @@ export const Bar = ({ administrator, dev, userId, onLogout }) => (
...
+
                     <span>{userId}</span>
                     <Icon name="caret-down" />
                 </NavDropdownButton>
                 <DropdownMenuList>
+                    <DropdownMenuLink to="/account">
+                        Signed in as <b>{handle}</b>
```
ReeceHoffmann:
```suggestion
                        Signed in as <strong>{handle}</strong>
```

Both should act identically in standard browsers, but can behave differently with assistive technology and non-standard browsers as `<b>` is strictly styling while `<strong>` indicates semantic importance

src/js/nav/components/NavBar.js
```diff
@@ -74,11 +83,16 @@ export const Bar = ({ administrator, dev, userId, onLogout }) => (
...
             <Dropdown>
                 <NavDropdownButton>
-                    <Icon name="user" />
+                    <InitialIcon handle={handle} size="md"></InitialIcon>
+
                     <span>{userId}</span>
                     <Icon name="caret-down" />
                 </NavDropdownButton>
                 <DropdownMenuList>
+                    <DropdownMenuLink to="/account">
```
ReeceHoffmann, line 89: I think having the username act like a link to account is a good idea, but I am personally not a huge fan of how `<DropdownMenuLink>` will lightly shade the background when hovered for this exact case. You might be able to override this behavior by restyling it via styled components, or implement the link another way. (Also possible Ian will disagree with me on the shading)

src/js/nav/components/NavBar.js
```diff
@@ -98,7 +112,8 @@ export const Bar = ({ administrator, dev, userId, onLogout }) => (
 export const mapStateToProps = state => ({
     ...state.account,
     dev: state.app.dev,
-    pending: state.app.pending
+    pending: state.app.pending,
+    handle: state.account.handle
```
ReeceHoffmann: When available, I try to use selectors to get data from redux state as they should be updated if the format of the redux state is ever changed. In this case there is the `getAccountHandle` selector from `account/selectors.js` I would use.

ReeceHoffmann (changes requested):
Overall I think the visual aspects of the PR look good. (though Ian's opinion is what really matters) As a general comment you should always take a look at what changes deep source is suggesting. For this PR deep source is highlighting a couple of issues that should be addressed. (redundant imports and should use self closing tag for `initialIcon`)

Anyway, I also left some specific feedback on parts of the code I would change. Ian might disagree on the visual style notes, so I wouldn't blame you for waiting to implement those. Otherwise you should be safe to implement everything else.

igboyes (changes requested):
Two things:

1. Extend thick grey bar to entire width of dropdown.
2. Commit message needs to change.
    This will appear in release notes and `InitialIcon` has no meaning to a user.
    How about: _fix: use user profile image for account drop down_

igboyes (approved):

PR #106, opened 2022-05-10
fix: skull icon display for terminated jobs

elanjm12:
Skull icon is displayed for terminated jobs. 

Screenshot from 2022-05-10 14-26-22

### DIFF
--- a/src/js/jobs/components/Item/Status.js
+++ b/src/js/jobs/components/Item/Status.js
@@ -27,23 +27,13 @@ export const JobStatus = ({ pad, state }) => {
         );
     }
 
-    if (state === "complete") {
-        return (
-            <StyledJobStatus pad={pad}>
-                <Icon name="check fa-fw" color="green" />
-                <span>{state}</span>
-            </StyledJobStatus>
-        );
-    }
-
-    if (state === "error" || state === "cancelled" || state === "terminated") {
-        return (
-            <StyledJobStatus pad={pad}>
-                <Icon name="times fa-fw" color="red" />
-                <span>{state}</span>
-            </StyledJobStatus>
-        );
-    }
-
-    return null;
+    return (
+        <StyledJobStatus pad={pad}>
+            <Icon
+                name={`${state === "complete" ? "check" : state === "terminated" ? "skull" : "times"} fa-fw`}
+                color={`${state === "complete" ? "green" : "red"}`}
+            />
+            <span>{state}</span>
+        </StyledJobStatus>
+    );
 };


src/js/jobs/components/Item/Status.js
```diff
@@ -26,20 +26,30 @@ export const JobStatus = ({ pad, state }) => {
...
             </StyledJobStatus>
         );
     }
 
-    if (state === "error" || state === "cancelled" || state === "terminated") {
+    if (state === "terminated") {
         return (
             <StyledJobStatus pad={pad}>
-                <Icon name="times fa-fw" color="red" />
+                <Icon name="fa-solid fa-skull" color="red" />
```
ReeceHoffmann:
```suggestion
                <Icon name="skull fa-fw" color="red" />
```

The `Icon` base component should only need `name="skull"` to get the correct icon. (`fa-fw` specifies that the icon needs to have a fixed width)

src/js/jobs/components/Item/Status.js
```diff
@@ -26,20 +26,30 @@ export const JobStatus = ({ pad, state }) => {
...
+    {
+        if (state === "complete") {
+            return (
+                <StyledJobStatus pad={pad}>
+                    <Icon name="check fa-fw" color="green" />
+                    <span>{state}</span>
+                </StyledJobStatus>
+            );
+        }
+    }
```
ReeceHoffmann:
```suggestion
        if (state === "complete") {
            return (
                <StyledJobStatus pad={pad}>
                    <Icon name="check fa-fw" color="green" />
                    <span>{state}</span>
                </StyledJobStatus>
            );
        }
```
Remove extra brackets

ReeceHoffmann (changes requested):
From the zoomed in view it looks good to me. Few small changes to the code below.

igboyes (changes requested):
The only difference between the two big `if` blocks is the `Icon.name`.

```js
<Icon name={`fa-fw ${state === "terminated" ? "skull" : "times"}`} color="red" />
```
* [Template Literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)
* [Ternary Operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)

ReeceHoffmann:
To add onto Ian's comment, I think using a ternary operator instead of a separate if statement here is a good idea. Though looking at how similar many of the states are in structure, (only changing the icon name and color) it might be a good idea to see if there is a way to simplify/optimize the code further by taking advantage of these similarities. 

Also forgot to mention this before, but the commit message should be changed here as well. Typical style is to write it in active imperative voice. 
ex, 
most basically: fix: skull icon display for terminated jobs -> fix: display skull icon for terminated jobs 
or to focus more on what makes the change beneficial:  fix:  update terminated job icon to visually differentiate from other failure states

igboyes (approved):

PR #107, opened 2022-05-11
fix: label description renders below label component

elanjm12:
Label descriptions render below the label component.

Screenshot from 2022-05-11 09-49-43

### DIFF
--- a/src/js/samples/components/Sidebar/Labels.js
+++ b/src/js/samples/components/Sidebar/Labels.js
@@ -10,10 +10,10 @@ import { getFontSize, fontWeight, getColor } from "../../../app/theme";
 import { xor } from "lodash-es";
 
 const SampleLabelInner = ({ name, color, description }) => (
-    <>
+    <div>
         <SmallSampleLabel color={color} name={name} />
-        <p>{description}</p>
-    </>
+        <StyledParagraph>{description}</StyledParagraph>
+    </div>
 );
 
 const SampleLabelsFooter = styled.div`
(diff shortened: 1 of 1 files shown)


src/js/samples/components/Sidebar/Labels.js
```diff
@@ -10,10 +10,10 @@ import { getFontSize, fontWeight, getColor } from "../../../app/theme";
 import { xor } from "lodash-es";
 
 const SampleLabelInner = ({ name, color, description }) => (
-    <>
+    <p>
```
igboyes: A `div` is more semantic here and should accomplish the same effect because it is also a block-level element.

Use `p` for blocks of text.

ReeceHoffmann (changes requested):
While this gets the job done I think a few style tweaks would be nice. 

For the description I would
1. Change the font size to `sm` 
2. Change the color to `greyDarkest`
3. Consider offsetting the description 5px to the right to have it align with the start of the label contents rather than the border of the label

Updated image:
image

For all the font sizes and colors you will want to get them from the theme component in `src/app/theme.js`. For styling use the library styled-components and take a look at other components for examples of how to leverage `theme` while styling other components. Feel free to experiment with other ideas as well, I am by no means attached to this specific style.

igboyes (changes requested):

igboyes:
> Consider offsetting the description 5px to the right to have it align with the start of the label contents rather than the border of the label.

Keep as is. Other points are fine.

elanjm12:
Updated screenshot of labels ui.

Screenshot from 2022-05-12 11-56-49

igboyes (approved):

PR #109, opened 2022-05-12
fix: label filtering working in sample detail view

elanjm12:
Label filtering works in sample detail view. The bug was caused by the structure of the results object being different when no filter is provided vs when a filter is typed.

Screenshot from 2022-05-12 10-53-09

### DIFF
--- a/src/js/samples/components/Sidebar/Selector.js
+++ b/src/js/samples/components/Sidebar/Selector.js
@@ -34,17 +34,20 @@ export const SampleSidebarSelector = ({
 }) => {
     const [results, term, setTerm] = useFuse(sampleItems, ["name"], [sampleId]);
     const [attributes, show, styles, setPopperElement, setReferenceElement, setShow] = usePopover();
-    const sampleItemComponents = results.map(item => (
-        <SampleSidebarSelectorItem
-            key={item.id}
-            selected={selectedItems.includes(item.id)}
-            partiallySelected={partiallySelectedItems.includes(item.id)}
-            {...item}
-            onClick={onUpdate}
-        >
-            {render(item)}
-        </SampleSidebarSelectorItem>
-    ));
+    const sampleItemComponents = results.map(item => {
+        const result = item.id ? item : item.item;
+        return (
+            <SampleSidebarSelectorItem
+                key={result.id}
+                selected={selectedItems.includes(result.id)}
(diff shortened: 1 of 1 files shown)


ReeceHoffmann (changes requested):
Looks like a bug caused by a relatively recent round of updates; very likely the object format was identical regardless of the search terms truthiness when first implemented. While the idea here of how to fix it is reasonable there are a few changes I would make.

1. I don't like that `useFuse` can return a result that is structured in different way depending on the truthiness of the search term term, I think a better solution here would be to change `useFuse` to ensure that the structure of the data being returned is independent of the term.

2. Instead of using a custom `for` loop I strongly recommend getting familiar with [lodash](https://lodash.com/docs/4.17.15) and using it whenever possible. Makes the code more readable and for more complex functions often faster than bespoke code. In this case you may find the `map` function useful

3. Likely not related to your code, but while testing I noticed that the list doesn't populate when the term consists of a single character. Given that fuse is a fuzzy search there might be a good reason for this, but I don't personally like that it shows a blank list when it seems obvious some results should show.

Also as deepSource is pointing out, when doing equality comparisons use `===` or `!==` not `==` or `!=`.  The triple equals variant is a strict comparison and way less likely to introduce insidious bugs. [MDN Equality_comparisons_and_sameness](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness)

igboyes:
> I don't like that `useFuse` can return a result that is structured in different way depending on the truthiness of the search term term, I think a better solution here would be to change `useFuse` to ensure that the structure of the data being returned is independent of the term.

Please just fix the issue first @elanjm12 so that it works for the user. @ReeceHoffmann, make a new issue to improve `useFuse` behaviour and assign to @elanjm12.

> I don't personally like that it shows a blank list when it seems obvious some results should show.

Yes, this is no good. Please include in secondary issue mentioned above @ReeceHoffmann.

> I strongly recommend getting familiar with [lodash](https://lodash.com/docs/4.17.15) and using it whenever possible.

I want to point out here that `.map()`, `.filter()`, and others are built-in methods on `Array`. This is the JS way with or without Lodash.

igboyes (changes requested):
Make changes using `map()` as we discussed. Just noting this here so well all know what is going on.

PR #110, opened 2022-05-13
fix: add overflow handling to label chooser in create sample view to prevent overflow

elanjm12:
Overflow handling added to label chooser in create sample view

Screenshot from 2022-05-13 10-30-34
.

### DIFF
--- a/src/js/samples/components/Sidebar/Selector.js
+++ b/src/js/samples/components/Sidebar/Selector.js
@@ -11,7 +11,7 @@ const SampleSidebarSelectorInputContainer = styled(BoxGroupSection)`
     padding: 10px;
 `;
 
-const StyledLinkButton = styled.div`
+const SampleSidebarSelectorButton = styled.div`
     display: flex;
 
     a {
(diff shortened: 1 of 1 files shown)


igboyes (changes requested):
Can you tweak this so the filter input component stays visible at the top and only the label list scrolls?

elanjm12:
filter input component remains visible while scrolling 

Screenshot from 2022-05-15 15-59-49

src/js/samples/components/Sidebar/Selector.js
```diff
@@ -69,10 +74,12 @@ export const SampleSidebarSelector = ({
                             autoFocus
                         />
                     </SampleSidebarSelectorInputContainer>
-                    {sampleItemComponents}
-                    <StyledLinkButton>
-                        <Link to={manageLink}> Manage</Link>
-                    </StyledLinkButton>
+                    <StyledDiv>
```
igboyes: `StyledLinkButton` is incorrectly named, so don't follow that as an example.

The Styled Components convention is to use the `Styled` prefix only in the following situation:
```js
const StyledBar = styled.div`...`;

const Bar = props => (
  <StyledBar>
    ...
  </StyledBar>
);
```

https://styled-components.com/docs/basics#how-do-styled-components-work-within-a-component

---

Can you rename `StyledLinkButton` to `SampleSidebarSelectorButton`?

igboyes (changes requested):
Please also change commit message to something a user will understand (eg. _prevent excessive label selector menu heights_).

igboyes (approved):

PR #111, opened 2022-05-13
fix: Update usefuse return matches on single character input and data…

elanjm12:
Usefuse returns matches if they exist on a single character input. Data is also returned in a consistent format regardless of what is entered in the search filter.

Screenshot from 2022-05-13 11-37-49


igboyes:
@elanjm12 and I discussed this one. He will:
* Make `useFuse` return the array (objects contain `score` and `refIndex`) outputted by Fuse instead of the array of our data objects.
* Test the hook.

PR #117, opened 2022-05-19
fix: make logout button visible from jobs page

elanjm12:
Logout button is visible when the jobs page is open and jobs list is being displayed.

Screenshot from 2022-05-19 11-43-20


igboyes (approved):
Looks great!

PR #118, opened 2022-05-20
fix: make manage button visible in label chooser overflow

elanjm12:
Manage button remains visible in create sample label chooser overflow.

Screenshot from 2022-05-20 09-22-42

### DIFF
--- a/src/js/samples/components/Sidebar/Selector.js
+++ b/src/js/samples/components/Sidebar/Selector.js
@@ -13,6 +13,11 @@ const SampleSidebarSelectorInputContainer = styled(BoxGroupSection)`
 
 const SampleSidebarSelectorButton = styled.div`
     display: flex;
+    border-top: 1px solid;
+    border-color: ${props => props.theme.color.greyLight};
+    width: 100%;
+    display: flex;
+    align-items: right;
 
     a {
         margin-left: auto;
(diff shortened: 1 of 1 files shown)


ReeceHoffmann (commented):
Code looks good, but personally I would add a border between the manage link and the scroll section for better visual separation. Something like this:


image

igboyes (changes requested):
Add border as suggested by @ReeceHoffmann.

elanjm12:
Border added

Screenshot from 2022-05-24 17-59-19

src/js/samples/components/Sidebar/Selector.js
```diff
@@ -77,12 +86,12 @@ export const SampleSidebarSelector = ({
                             autoFocus
                         />
                     </SampleSidebarSelectorInputContainer>
-                    <SampleItemComponentsContainer>
-                        {sampleItemComponents}
-                        <SampleSidebarSelectorButton>
+                    <SampleItemComponentsContainer>{sampleItemComponents}</SampleItemComponentsContainer>
+                    <SampleSidebarSelectorButton>
+                        <SampleLinkContainer>
```
ReeceHoffmann: You might be able to merge `<SampleSidebarSelectorButton>` and `<SampleLinkContainer>` into a single styled component. (I don't think `<SampleSidebarSelectorButton>` is used anywhere else in the component) Otherwise looks good to me!

igboyes (changes requested):
image

Fix white gap. Probably originates from margin on top of box containing **Manage** link.

igboyes (approved):

PR #120, opened 2022-05-24
fix: retrieve job step title and description from api

elanjm12:
Title and description for each jobs step accessed directly from the API. The hard coded step names and descriptions have been deleted.

Screenshot from 2022-05-24 16-56-50


src/js/jobs/components/Step.js
```diff
@@ -47,15 +47,14 @@ const StyledJobStepDescription = styled.div`
     }
 `;
 
-export const JobStepDescription = ({ stage, state, workflow, timestamp }) => {
-    const { description, title } = getStepDescription(stage, state, workflow);
-
+export const JobStepDescription = info => {
```
ReeceHoffmann: A component's `props` (here called `info`) should always be destructured. Doing so is conventional and more importantly greatly increases readability of the codebase by making the required props obvious at a glance. As an added bonus it makes referencing the contents of the object easier downstream. ie info.step -> step

src/js/jobs/components/Step.js
```diff
@@ -107,12 +106,12 @@ const StyledJobStep = styled(BoxGroupSection)`
...
 `;
 
-const JobStep = ({ complete, step, workflow }) => (
+const JobStep = ({ complete, step }) => (
     <StyledJobStep>
         <JobStepIconContainer>
             <JobStepIcon state={step.state} complete={complete} />
         </JobStepIconContainer>
-        <JobStepDescription {...step} workflow={workflow} />
+        <JobStepDescription {...step} step={step} />
```
ReeceHoffmann: As written the props being passed are duplicated. Your explicitly passing the reference to the step object with `step={step}` but all the contents of step are also being passed by the `{...step}` statement. In my opinion, destructing ln props is not great practice for readability reasons, so I would either destructure `step` and pass the contents explicitly or destructure step in the receiving component.

src/js/jobs/components/__tests__/Step.test.js
```diff
@@ -1,29 +1,36 @@
...
     let props;
-
-    getStepDescription.mockImplementation(() => ({
-        title: "Foo",
-        description: "Bar"
-    }));
-
+    const spyStepDescription = jest.spyOn(functions, "getStepDescription");
     beforeEach(() => {
+        spyStepDescription.mockClear().mockReturnValue({ description: "foo", title: "bar" });
```
ReeceHoffmann: While it was mocked in the earlier versions of the test, I think it would be better to allow the call to `getStepDescription`  to go through. I would first update your testing props to ensure that all expected data is being passed and correctly formatted.  Then you can make a real call to `getStepDescription` and check that the correct values are being returned.

src/js/jobs/components/__tests__/Step.test.js
```diff
@@ -1,29 +1,36 @@
...
-        const { stage, state, workflow } = props;
-        expect(getStepDescription).toHaveBeenCalledWith(stage, state, workflow);
-        expect(wrapper).toMatchSnapshot();
+    it("renders correctly", () => {
+        renderWithProviders(<JobStepDescription {...props} />);
+        expect(screen.getByText("foo")).toBeInTheDocument();
+        expect(screen.getByText("bar")).toBeInTheDocument();
+    });
+
+    it("renders correctly", () => {
```
ReeceHoffmann: Every test in a file should have a descriptive name unique to that file. Name should be changed to reflect what the specific thing the test is checking.

src/js/jobs/components/__tests__/Step.test.js
```diff
@@ -1,29 +1,36 @@
...
+        renderWithProviders(<JobStepDescription {...props} />);
+        expect(functions.getStepDescription).toHaveBeenCalledTimes(1);
+    });
+
+    it("renders correctly", () => {
+        let step = props;
+        let expectedVal = { description: "foo", title: "bar" };
+        expect(functions.getStepDescription(step)).toEqual(expectedVal);
```
ReeceHoffmann: Since the return value for `getStepDescription` is mocked this test isn't testing the actual code.

src/js/jobs/utils.js
```diff
@@ -1,233 +1,47 @@
...
-    },
-    pathoscope: {
-        title: "Pathoscope",
-        description: "Reassign ambiguous mappings using Pathoscope 2.0."
-    },
-    import_results: importResultsDescription,
-    cleanup_indexes: {
-        title: "Clean-up Indexes",
-        description: "Prune stale OTU indexes if they are no longer being used."
+export const getStepDescription = step => {
```
ReeceHoffmann: Not strictly required, but I would consider destructuring step here. Something like:
```suggestion
export const getStepDescription = step => {
  const {state, step_name, step_description} = step;
```

src/js/jobs/utils.js
```diff
@@ -1,233 +1,47 @@
...
             description: "There was a system malfunction"
         };
     }
 
-    return get(stepDescriptions, [workflow, stage], {
-        description: "",
-        title: startCase(stage)
-    });
+    return {
+        title: "",
```
ReeceHoffmann, line 44: I think it is still a good idea return the stage as the title as a fallback if none of the other conditions are met.
```suggestion
        title: startCase(stage),
```

src/js/jobs/components/__tests__/Step.test.js
```diff
@@ -1,29 +1,36 @@
 jest.mock("../../utils");
```
ReeceHoffmann: As a general note for the tests I would suggest changing your tests to check the most common situations that can occur. To do this I would suggest modifying the passed props to mimic what would be passed when the component is being rendered in the browser. 

Of the top of my head this would be three main configurations. 
 1. The `step_name` and `step_description` is sent from the server. 
 2. No `step_name` or `step_description` but the `state` is one of the special cases ie: waiting
 3. No `step_name` or `step_description` and `state` is not one of the special cases.

Create props that mimic each of these situations then write separate tests for each situation that checks for correct rendering using RTL.

ReeceHoffmann (changes requested):
Overall it works well but there are a some changes I have suggested below. Also a couple of the CI checks are failing. The tests seem to be failing due to an extra snapshot being included that should be removed. Also deep source is highlighting a couple of issues that should be addressed. Both should be fixed before merging.

src/js/jobs/utils.js
igboyes: The API can be relied on to provide a `step_name`. Keep the empty string so it is obvious if this breaks.

src/js/jobs/utils.js
igboyes: This is preferrable:
```js
export const getStepDescription = (state, name, description) => {};
```
Function signature is much better defined.

igboyes (changes requested):
See my comments on @ReeceHoffmann's review.

src/js/jobs/components/Step.js
```diff
@@ -31,7 +31,7 @@ const JobStepTimestamp = ({ timestamp }) => (
     </StyledJobStepTimestamp>
 );
 
-const StyledJobStepDescription = styled.div`
+const SampleJobStepDescription = styled.div`
```
igboyes: Wondering why this is prefixed with `Sample`?

igboyes (commented):
One question. Otherwise, looks ready to merge.

PR #122, opened 2022-05-26
fix: prevent display of previous reference data in reference views

ReeceHoffmann:
Small fix that prevents data from previous references being displayed when navigating to the `references/otus` and `references/indexes` views of a different reference.

### DIFF
--- a/src/js/indexes/__tests__/reducer.test.js
+++ b/src/js/indexes/__tests__/reducer.test.js
@@ -112,6 +112,21 @@ describe("Indexes Reducer", () => {
     });
 
     it("should handle FIND_INDEXES_REQUESTED", () => {
+        const state = { refId: "foo" };
+        const term = "bar";
+        const refId = "foo";
+        const action = {
+            type: FIND_INDEXES.REQUESTED,
+            payload: { refId, term, page: 4 }
+        };
+        const result = reducer(state, action);
+        expect(result).toEqual({
+            term,
+            refId
+        });
+    });
+
+    it("should handle FIND_INDEXES_REQUESTED when [payload.refId != state.refId]", () => {
         const state = {};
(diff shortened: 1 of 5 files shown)


igboyes (commented):
Can we not just show a loading icon in the index detail view when there is reference ID mismatch:
```js
if (index.reference.id !== match.params.refId) {...}
```

And repeat for each other view.

ReeceHoffmann:
> Can we not just show a loading icon in the index detail view when there is reference ID mismatch:
> 
> ```js
> if (index.reference.id !== match.params.refId) {...}
> ```
> 
> And repeat for each other view.

Yeah I think that can work.

Not sure if this is what your wanting to be changed, but for the index detailed views that is absolutely a viable option. The current implementation (not changed by this PR) just assumes the data in cache is wrong and deletes it when the request is sent. Checking this way would definitely help load times in the case the the user is navigating back to the same index detail repeatedly. Happy to update the PR to include this change 

That said the current changes in this PR are focused on the index and OTU list view. Looking at the data returned I think the same style of check is possible. Since the returned indexes/OTUs contain a reference id I can grab the id from one of the items returned and check if it matches the current URL. My assumption is that all items in the array should have the same refId so it should only be necessary to check the first item in the array for a match, but if I am mistaken please let me know.

igboyes:
That's all good with me!

I wasn't thinking about the lists, but getting the first item's reference ID will work.

ReeceHoffmann:
So turns out that using the reference ID stored in the first item of list view documents doesn't work as the list can be empty.  The index detail view has been updated the way you suggested and the index otu detail view was already implemented that way.

igboyes:
> documents doesn't work as the list can be empty.

Dang. Okay.

PR #123, opened 2022-05-27
Restore subtraction upload information icon

ReeceHoffmann:
Small tweak to restore the information icon that was intended to be included once file type restrictions were put in place for the subtractions view file manager.

Screenshot without tooltip:
image

Screenshot with tooltip:
image

### DIFF
--- a/src/js/base/UploadBar.js
+++ b/src/js/base/UploadBar.js
@@ -44,14 +44,19 @@ export const UploadBar = ({ message, onDrop, validator, tip }) => {
     const rootProps = getRootProps({
         onClick: e => e.stopPropagation()
     });
-
     return (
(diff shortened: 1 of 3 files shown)


igboyes (changes requested):
I don't think this looks good or provides good UX.

* Icon after period looks weird.
* The info icon is used elsewhere as a prefix before an informative piece of text. This deviates.
* The info icon is not hoverable anywhere else. How is the user to know there is more information on hover?

I think we should close this and fix up the `UploadBar` with the following:
* Allow custom 'Drag here' element. This will allow us to put a better formatted two line message for subtraction uploads.
* Change **Upload** button to **Browse Files**.
* Use vertical OR divider with drag message on left and upload button on right.

PR #124, opened 2022-05-27
fix: make filtering feature work on single character input and add tests

elanjm12:
Usefuse returns results on single character inputs. And tests added for useFuse hook.

Screenshot from 2022-05-27 09-05-43

### DIFF
--- a/src/js/base/__tests__/hooks.test.js
+++ b/src/js/base/__tests__/hooks.test.js
@@ -1,0 +1,74 @@
+import { useFuse } from "../hooks";
+import { act, renderHook } from "@testing-library/react-hooks";
+
+describe("useFuseHook", () => {
+    let collection;
+    let keys;
+    let deps;
+
+    beforeEach(() => {
+        collection = [{ name: "nan" }, { name: "test1" }, { name: "aaaaa" }, { name: "zzzzz" }];
+        keys = ["name"];
+        deps = ["zzz"];
+    });
+
+    it("should render hook correctly and return collection when term = empty string", () => {
+        const { result } = renderHook(() => useFuse(collection, keys, deps));
+        const length = result.current.length;
+        expect(length).toBe(3);
+        expect(result.current[0]).toBe(collection);
+        expect(result.current[1]).toBe("");
+    });
+
+    it("should change returnObj to match useFuse result when term is changed to non empty single character", () => {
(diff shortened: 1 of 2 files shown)


src/js/base/__tests__/UseFuse.test.js
```diff
@@ -0,0 +1,71 @@
...
+    it("when the term is changed from nonempty string to empty string collection is returned", ()=> {
+        const { result } = renderHook(() => useFuse(collection, keys, deps));
+        const [ returnObj, term, setTerm] = result.current;
+        
+        act(()=> {
+            setTerm("z");
+            setTerm("");
+        });
+        expect(result.current[1]).toBe("");
+        expect(result.current[0]).toBe(collection);
```
ReeceHoffmann: Since returning collection is currently the default state, I would personally separate the two calls to `setTerm` and confirm that collection is not being returned when a term is set. That way we can be fully confident that collection is being returned is really the result of reversion to returning collection rather than just always having returned collection.

src/js/base/__tests__/UseFuse.test.js
```diff
@@ -0,0 +1,71 @@
...
+        const [ returnObj, term, setTerm] = result.current;
+        const length = result.current.length;
+        expect(length).toBe(3);
+        expect(result.current[0]).toBe(collection);
+        expect(result.current[1]).toBe("");
+    })
+    
+    it("after term changed to non empty single character string returnObj should change", () => {
+        const { result } = renderHook(() => useFuse(collection, keys, deps));
+        const [ returnObj, term, setTerm] = result.current;
```
ReeceHoffmann: In cases like this where the restructured values for `returnObj` and `term` can't be used, I would just pull out the one term that is used.

ReeceHoffmann (changes requested):
Tests mostly look good couple of comments though:

1. Not directly test related, but my understanding was that we wanted to ensure the `useFuse` always returned an object of consistent structure. Not sure that goal changed in your talks with Ian, but right now the `collection` object and the object returned when there is a search term have a  different structure.

2. I mentioned this on slack, but for readability reasons I would suggest reformatting the test names to match the "should <action> when <action prompt>" format that is standard in the codebase

igboyes (changes requested):
Change `UseFuse.test.js` to `hooks.test.js` and I will merge.

igboyes (approved):

PR #125, opened 2022-06-03
Update look of UploadBar base component

ReeceHoffmann:
Updated with new images : 17/06/2022 

### Changes:
  - Changes look of base `uploadBar` component
  - Adds description of accepted file types for the subtraction file manager
  - Adds RTL tests for subtraction file manager
  - Enables styling of base `divider` by parent components by accepting class name prop
     - Slight tweak to B2C login view to prevent change in styling caused by this update 

###  Screenshots:

**Updated subtraction file manager:** (Custom message component)
image

**Updated samples subtraction manager:** (default message string)
image


igboyes (changes requested):
1. Bring text and button components closer to divider.
2. Maintain box padding instead of overriding for the divider.
3. Use uppercase for 'Or'.
4. Maybe introduce a `min-height` on the text element so that things don't get too compressed vertically.

image

src/js/base/__tests__/UploadBar.test.js
```diff
@@ -1,13 +1,44 @@
...
+    it("should display optional message", () => {
+        props.message = "test_message";
+        renderWithProviders(<UploadBar {...props} />);
+        expect(screen.getByText("test_message")).toBeInTheDocument();
+    });
+
+    it("should call onDrop with uploaded files", async () => {
+        renderWithProviders(<UploadBar {...props} />);
+        const file = new File(["test"], "test_file.fa.gz", { type: "application/gzip" });
+        await userEvent.upload(screen.getByTestId("upload-input"), file);
```
igboyes: Can you not use input role here instead of test-id?

ReeceHoffmann: Unfortunately no because `input` elements of `type=file` don't have a role. I originally used the test-id because this input element is invisible to the user, has no role, but must be targeted for the test. That said, I ended up changing it to use an aria-label as I could see the possibility that assistive technology may need to directly target the input element as well.

igboyes (approved):

PR #128, opened 2022-06-10
fix: render icons as buttons and fix tests

elanjm12:
Icons are rendered as buttons when onClick function is defined. Instances of data-testid work around have been removed.

### DIFF
--- a/src/js/base/Icon.js
+++ b/src/js/base/Icon.js
@@ -15,6 +15,8 @@ const fixedWidth = css`
 `;
 
 const StyledIcon = styled.i`
+    background: inherit;
+    border: none;
     color: ${getIconColor};
     ${props => (props.hoverable || props.onClick ? "cursor: pointer;" : "")};
     ${props => (props.fixedWidth ? fixedWidth : "")};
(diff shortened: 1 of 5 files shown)


src/js/base/Icon.js
```diff
@@ -31,17 +33,18 @@ export const Icon = ({ hoverable, style, ...props }) => {
     const handleClick = useCallback(props.onClick, [props.onClick]);
 
     const className = `${props.className ? props.className + " " : ""} ${props.faStyle} fa-${props.name}`;
-
+    const renderAs = !props.onClick ? "i" : "button";
```
igboyes: Don't think you need to declare this variable. Just provide prop directly in component call.

src/js/samples/components/Create/Create.js
```diff
@@ -204,7 +204,7 @@ export const CreateSample = ({
                                 />
                                 <InputIcon
                                     name="magic"
-                                    data-testid="Auto Fill"
+                                    aria-label="magic icon"
```
igboyes: Should be `aria-label="Auto fill"`.

igboyes (changes requested):

src/js/base/Icon.js
```diff
@@ -31,17 +33,17 @@ export const Icon = ({ hoverable, style, ...props }) => {
     const handleClick = useCallback(props.onClick, [props.onClick]);
 
     const className = `${props.className ? props.className + " " : ""} ${props.faStyle} fa-${props.name}`;
-
     const icon = (
         <StyledIcon
+            as={!props.onClick ? "i" : "button"}
```
igboyes:
```suggestion
            as={props.onClick ? "button" : "i"}
```

igboyes (changes requested):
I should have picked this up last time. Last one before merge.

igboyes (approved):

PR #129, opened 2022-06-13
fix: improve analysis view and add tests

elanjm12:
Make improvements to create analysis ui and add tests

Screenshot from 2022-06-13 12-44-41


elanjm12:
Converted to draft to address deepsource issues

src/js/analyses/components/Create/Create.js
```diff
@@ -18,13 +18,21 @@ import { ReferenceSelector } from "./ReferenceSelector";
 import { SubtractionSelector } from "./SubtractionSelector";
 import { CreateAnalysisSummary } from "./Summary";
 import { WorkflowSelector } from "./WorkflowSelector";
+import { SelectedWorkflows } from "./SelectedWorkflows";
+import { SelectedSubtractions } from "./SelectedSubtractions";
+import { SelectedReferences } from "./SelectedReferences";
+import Grid from "@material-ui/core/Grid";
```
ReeceHoffmann: Better to work with the CSS directly rather than using a predefined styling library. I would switch the to use css grid and remove `material-ui/core` from the package

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,94 @@
...
-            name={index.reference.name}
-            version={index.version}
-        />
-    ));
+    const [results, term, setTerm] = useFuse(indexes, ["reference.name"], [1]);
+
+    const references = results.filter(item => {
+        const obj = item.id ? item : item.item;
+        return !selected.includes(obj.reference.id);
+    });
```
ReeceHoffmann: A few comments:

1. For sake of ease of reading it is better to give variables descriptive names as much as possible. A common pattern when iterating through and array is to to the use the non-plural version of the name for the item. ie results(array) -> result (item)

2. I would look at using `difference` from lodash as that should accomplish the same goal while being easier to read. 

3. Since you don't use the extra information passed back by Fuse I would reformat the object so the final reference object is easier to work with in later functions

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,94 @@
...
+    const referenceComponents = references.map(item => {
+        const result = item.id ? item : item.item;
         return (
-            <>
-                <label htmlFor="references">References</label>
-                <NoneFoundBox noun="references" id="references">
-                    <Link to="/refs">Create one</Link>.
-                </NoneFoundBox>
-            </>
+            <Fragment key={item.id}>
```
ReeceHoffmann: Why is this wrapped in a `Fragment`? With only one child component the `Fragment` seems unnecessary.

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,94 @@
...
+                {!selected.includes(result.reference.id) && (
```
ReeceHoffmann: Shouldn't all the references that were select already have been removed when defining the `references` variable? If so this adds a lot of runtime to no effect.

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -63,3 +99,4 @@ ReferenceSelector.propTypes = {
     selected: PropTypes.arrayOf(PropTypes.string).isRequired,
     onChange: PropTypes.func.isRequired
 };
+<label htmlFor="references">References</label>;
```
ReeceHoffmann: Delete

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,94 @@
...
             <label>References</label>
-            <MultiSelector
-                error={hasError && "Reference(s) must be selected"}
-                noun="references"
-                selected={selected}
-                onChange={onChange}
-            >
-                {referenceComponents}
-            </MultiSelector>
+            {indexes.length - selected.length > 0 && (
```
ReeceHoffmann:
```suggestion
            {referenceComponents.length && (
```

src/js/analyses/components/Create/SelectedReferences.js
```diff
@@ -0,0 +1,38 @@
+import React from "react";
+import { MultiSelector } from "../../../base/MultiSelector";
+import { Icon } from "../../../base";
+import { NoneSelected, IconContainer, WorkflowItem, MaxWidthContainer } from "./SelectedWorkflows";
+
+const clearSelectedReference = (references, onRemove, reference) => {
+    onRemove(references.filter(item => item !== reference));
+};
+
+export const SelectedReferences = ({ referenceOptions, references, onRemove }) => {
```
ReeceHoffmann: Renaming `setReferences` to `onRemove` seems potentially confusing as `setReferences` is not limited in scope to removing elements from the arrary. I would leave it as `setReferences` and define your onRemove locally as the function actually passed to the Icon base component.

src/js/analyses/components/Create/SelectedReferences.js
```diff
@@ -0,0 +1,38 @@
...
+    const getReferenceName = reference => {
+        const item = referenceOptions.filter(obj => {
+            return obj.reference.id === reference;
+        });
+        return item[0].reference.name;
```
ReeceHoffmann: Two comments:

1. This is a confusing way of writing this. use something like `find` which has the dual benefits of being way more readable and stopping after it finds the one element it is looking for rather than filtering the whole list on every call.

2. In this case I would suggest avoiding this altogether by creating a filtered list of the actual reference objects at the start of the function. You may find `intesectionWith` or `intersectionBy` to be useful for this.

src/js/analyses/components/Create/SelectedSubtractions.js
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import { MultiSelector } from "../../../base/MultiSelector";
+import { Icon } from "../../../base";
+import { NoneSelected, IconContainer, WorkflowItem, MaxWidthContainer } from "./SelectedWorkflows";
+
+const clearSelectedSubtraction = (subtractions, onRemove, subtraction) => {
+    onRemove(subtractions.filter(item => item !== subtraction));
+};
+
+export const SelectedSubtractions = ({ subtractionOptions, subtractions, onRemove }) => {
```
ReeceHoffmann: Same comment as for references, not a big fan of the rename to `onRemove`

src/js/analyses/components/Create/SelectedSubtractions.js
```diff
@@ -0,0 +1,40 @@
...
+    const getSubtractionName = subtraction => {
+        const item = subtractionOptions.filter(obj => {
+            return obj.id === subtraction;
+        });
+        return item[0].name;
+    };
```
ReeceHoffmann: Same comment as `selectedReferences`.

src/js/analyses/components/Create/SelectedReferences.js
```diff
@@ -0,0 +1,38 @@
...
+    };
+    const items = references.map(reference => {
```
ReeceHoffmann: Please use descriptive variable names

src/js/analyses/components/Create/SubtractionSelector.js
```diff
@@ -1,20 +1,36 @@
...
+    const [results, term, setTerm] = useFuse(subtractions, ["name"], [1]);
+    const handleRemove = items => {
+        if (items.length) {
+            const obj = items[0].name
+                ? items.filter(item => !value.includes(item.id))
+                : items.filter(item => !value.includes(item.item.id));
+            return obj;
+        }
+        return items;
+    };
```
ReeceHoffmann: 1. Please use descriptive variable names for all variables 
2. The name of this function does not fit with convention.  The name `handle<x>` is typically used when responding to a user action. Since `handleRemove` does not respond to user input naming it this way is confusing.
3. Reconsider rewriting this to make it more readable, as the intention is not clear at a glance. Fixing the names will help a lot, but there is likely better ways to write the logic itself.

src/js/analyses/components/Create/SelectedWorkflows.js
```diff
@@ -0,0 +1,56 @@
...
+export const MaxWidthContainer = styled.div`
+    max-width: 60%;
+    overflow: hidden;
+`;
+
+const clearSelectedWorkflow = (workflows, onRemove, workflow) => {
+    onRemove(workflows.filter(item => item !== workflow));
+};
+
+export const SelectedWorkflows = ({ workflows, onRemove }) => {
```
ReeceHoffmann: Same comment about the naming of `onRemove`

src/js/analyses/components/Create/WorkflowSelector.js
```diff
@@ -18,9 +25,19 @@ export const getCompatibleWorkflows = (dataType, hasHmm) => {
 
 export const WorkflowSelector = ({ dataType, hasError, hasHmm, workflows, onSelect }) => {
     const workflowItems = map(getCompatibleWorkflows(dataType, hasHmm), workflow => (
-        <MultiSelectorItem key={workflow} name={workflow} value={workflow}>
-            {getWorkflowDisplayName(workflow)}
-        </MultiSelectorItem>
+        <Fragment key={workflow}>
```
ReeceHoffmann: unnecessary Fragment

src/js/analyses/components/Create/__tests__/SelectedReferences.test.js
```diff
@@ -0,0 +1,58 @@
...
+import { screen } from "@testing-library/react";
+import { SelectedReferences } from "../SelectedReferences";
+
+describe("<SelectedReferences />", () => {
+    let props;
+
+    beforeEach(() => {
+        props = {
+            referenceOptions: [
+                { id: "1", version: 0, reference: { id: "one", data_type: "genome", name: "zebra" } },
```
ReeceHoffmann: This isn't consistent across the codebase, but I have a strong preference for giving test variables values that are either accurate or easy to identify. ie I would replace `name: zebra` with something along the lines of `name: reference_1_name`. It can make tracking what went wrong when writing complex tests much easier.

src/js/base/MultiSelector.js
```diff
@@ -18,27 +18,43 @@ const StyledMultiSelectorItem = styled(SelectBoxGroupSection)`
...
+                    active={active}
+                    className={className}
+                    name={name}
+                    type="button"
+                    onClick={() => onSelect(value)}
+                >
+                    {children}
+                </StyledMultiSelectorItem>
+            )}
+        </>
```
ReeceHoffmann: Since these two are so similar I would toggle target only the checkbox:
```javascript 
               <StyledMultiSelectorItem
                    as="button"
                    active={active}
                    className={className}
                    name={name}
                    type="button"
                    onClick={() => onSelect(value)}
                >
                    {requireCheckbox && <Checkbox checked={active} />} {children}
                </StyledMultiSelectorItem>
```

src/js/base/MultiSelector.js
```diff
@@ -18,27 +18,43 @@ const StyledMultiSelectorItem = styled(SelectBoxGroupSection)`
     user-select: none;
 `;
 
-export const MultiSelectorItem = ({ children, className, name, value }) => {
+export const MultiSelectorItem = ({ children, className, name, value, requireCheckbox }) => {
```
ReeceHoffmann: Not a specific change, but when changing the definition and behavior of a base component I would be really careful to ensure it doesn't break the look or functionality of anything outside of what you were trying to style.

src/js/subtraction/components/Selector.js
```diff
@@ -1,27 +1,91 @@
...
+    subtractions,
+    onChange,
+    requireCheckbox,
+    requireInput,
+    term,
+    setTerm,
+    subtractionOptions
+}) => {
+    const subtractionComponents = subtractions.map(item => {
+        const result = item.id ? item : item.item;
```
ReeceHoffmann: Use descriptive variable names

src/js/subtraction/components/Selector.js
```diff
@@ -1,27 +1,91 @@
...
         );
-    }
+    });
+
     return (
-        <MultiSelector name={name} noun={noun} selected={selected} onChange={onChange}>
-            {subtractionComponents}
-        </MultiSelector>
+        <>
+            {requireInput === true && (
```
ReeceHoffmann: Comparison to true is probably redundant. Second statement will be included so long as `requireInput` is truthy.

src/js/subtraction/components/Selector.js
```diff
@@ -1,27 +1,91 @@
...
+                <>
+                    {subtractionOptions.length - selected.length > 0 && (
```
ReeceHoffmann: Why not just check if length of subtractions is truthy?

src/js/subtraction/components/Selector.js
```diff
@@ -1,27 +1,91 @@
...
+                <MultiSelector name={name} noun={noun} selected={selected} onChange={onChange}>
+                    {subtractionComponents}
+                </MultiSelector>
+            )}
+
+            {!subtractionOptions.length && (
+                <NoneFoundBox noun="subtractions">
+                    <Link to="/subtractions"> Create one</Link>.
+                </NoneFoundBox>
+            )}
```
ReeceHoffmann: Since this being true precludes all other states, I would just have this as a separate return statement the triggers if there are no subtraction options.

ReeceHoffmann (changes requested):
Alright so this round of changes is focused solely on the logic. A couple of notes up top:

1. I tried to pointed out problems as I noticed them, but given the size of the PR I would strongly suggest taking another good look through your code while trying to improve readability as a few sections take a bit to parse the intention.

2. I noticed a lot of duplication between the new `Selected{item}` components. It might be worth it to see if there is a clean way to reuse the core logic and styling.

3. In terms of general data flow, my instinct would be to reorganize it so that from the perspective of the form it only directly passes props to the `<{x}Selector>` component while having the `Selected{x}` component as a child of the `Selector` component. This is especially the case if you can generalize the `selected` component such that it can be mostly reused, then you can define the specific functions in the parent `Selector` component and  just pass the bare minimum to the `selected` components.

src/js/analyses/components/Create/ReferenceSelector.js
elanjm12: 1) addressed
2) still todo
3) version is used so result must remain in the current format. (line 49)

src/js/analyses/components/Create/ReferenceSelector.js
elanjm12: resolved

src/js/analyses/components/Create/ReferenceSelector.js
elanjm12: resolved. Fragment has been removed

src/js/analyses/components/Create/ReferenceSelector.js
elanjm12: resolved. Deleted

src/js/analyses/components/Create/ReferenceSelector.js
elanjm12: resolved. used referenceComponents.length < 0 && (

otherwise references0 renders when all references selected

src/js/analyses/components/Create/SelectedReferences.js
elanjm12: resolved

src/js/analyses/components/Create/ReferenceSelector.js
ReeceHoffmann: >     3. version is used so result must remain in the current format. (line 49)

Not sure what your getting at, `results.version` should be an intrinsic property of any index. I am suggesting that the first time you modify `useFuse`'s result you could prune the extra information added to the `results` object by `useFuse` .

src/js/analyses/components/Create/ReferenceSelector.js
ReeceHoffmann: Ah yeah rendering the zero is less than ideal. an alternative option is to directly cast `referenceComponents.length` directly to a boolean via `Boolean(referenceComponents.length)` rather than doing a comparison with zero.

elanjm12: This actually produces a different result. In this case indexes.length - selected.length is the remaining unselected references regardless of the search term. The referenceComponents length can be 0 if the search term yields no results. And in this case an empty search will result in the filter bar being taken away when it is still needed

src/js/analyses/components/Create/ReferenceSelector.js
elanjm12: I'll look into it again. thank you

src/js/analyses/components/Create/ReferenceSelector.js
ReeceHoffmann: A fair point, that is going to be a problem. Another solution would be to filter the the indexes against the selected indexes prior to passing them to `useFuse`, which would give a convenient way to check, but given the need for indexing when a Fuse is created I don't actually think it is a great idea.

Unfortunately I still think the logic about what items are rendered is non-obvious. One off the cuff idea that I think would work would be to:

  1. Check if indexes.length is falsy, if it is then return the NoneFound message and don't execute anything else.
  2. Check if all the references are selected. Since we know from (1) that indexes.length >== 1 we can skip that check and just test if whether selected.length and index.length are the same. If they are, then return the `All references selected` message.
  3. Finally since we know that indexes.length >== 1 and that selected.length !== index.length then we can assume that the searchBar + options should be rendered. (Since it should never be the case that selected.length > index.length)

I would imagine this to be best implemented using three different return statements, but you could play around with that. Assuming it works, the advantage I see here is making the conditions which trigger each option pretty obvious at a glance. 

I am definitely not stuck on the exact logic flow either, so feel free to mess around and try to find something better.

src/js/analyses/components/Create/SelectedReferences.js
elanjm12: Screenshot from 2022-06-15 10-36-48

Does this look clearer to you? It results in an object that contains the properties needed and no more ane eliminates the need for the getReferenceName function

src/js/analyses/components/Create/SelectedSubtractions.js
elanjm12: resolved in same manner as selectedReferences

src/js/analyses/components/Create/SelectedReferences.js
elanjm12: resolved

elanjm12: also resolved same issue in selectedSubtractions

src/js/analyses/components/Create/SelectedReferences.js
ReeceHoffmann: Yeah, I think this is a solid improvement. I like the fact that it dodges the need to repeatedly look through `referenceOptions` for the reference name when creating the `selectedReferenceComponents`, and once someone knows what `intersection` does it is pretty clear what is happening here. That said, I probably wouldn't worry about removing the extra data packed with each entry as is done with the map step. My guess is that the name would still be easy enough to grab while making the components even if there is some extra data.

src/js/analyses/components/Create/SubtractionSelector.js
elanjm12: Screenshot from 2022-06-15 12-34-27

Does this look clear enough? I've separated the logic for formatting the results and work directly with the formattedResults object. This also cleans up code in the selector.js file as there is no longer a need to check which type of data the results is each time.

src/js/base/MultiSelector.js
elanjm12: Ok i will watch for this, thank you

src/js/analyses/components/Create/SelectedReferences.js
elanjm12: Ok map section has been removed

src/js/analyses/components/Create/SubtractionSelector.js
ReeceHoffmann: On first glance I would say it is a big improvement. There are a couple things that stand out to me as a bit odd, but there may be a good reason for them.

 1. I don't see an obvious reason why we would pass the removeSelected function with a parameter rather than just running calculating a `filteredResults` variable and passing that
 
 3. Not sure if we need to check whether `formattedResults` has values prior to filtering it. I assume that the filtering function would just return an empty array in the case that `formattedResults` is empty.

Also this might be another good opportunity to exploit lodash functionality. Checkout differenceWith, should be useful.

elanjm12:
Ui is improved and addresses the styling changes requested by Ian, tests have been moved to the create.test file to prevent mocking being needed. And changes to logic as requested by Reece.

Screenshot from 2022-06-27 05-30-24

elanjm12:
Subtle box resizing when all selected fixed, CreateHeader, AllSelectedBox and CreateNoneFoundBox switched to be react components as requested. formatName function switched to FormattedLine to improve clarity. Reformatting of test file.

elanjm12:
Additionally, rounded borders are now consistent throughout.

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
+                                    <Icon name="times fa-fw" onClick={() => setTerm("")} aria-label="clear" />
+                                </IconContainer>
+                            </SampleInputContainer>
+                            <ItemComponentsContainer>{referenceComponents}</ItemComponentsContainer>
+                        </MultiSelector>
+                    )}
+                    {selectedReferenceItems.length === indexes.length && indexes.length > 0 && (
+                        <AllSelectedBox type="references" />
+                    )}
+                    {indexes.length === 0 && <CreateNoneFoundBox type="references" link="/refs" />}
```
ReeceHoffmann: So I have been thinking this over and I am still not a huge fan of how this logical flow works. I think there is still room for improvement in regards to making it more obvious why a given component is being rendered and that all states are mutually exclusive. Since the conditions are mutually exclusive I think this might make sense to factor out and parse the logic using and if/else statements. Doing it this way should reduce the number of checks we need to make. I would do something like this (pseudo code incoming):

```javascript
let referenceSelector;

if (!indexes.length){
  referenceSelector = <NoneFound />
}else if(selectedReferenceItems.length === indexes.length){
  referenceSelector = <AllSelected />
}else {
   referenceSelector =<normalSelectorComponent />
}

return (
<YourWrappers>
{selectorComponent}
<SelectedComponent>
</YourWrappers>
)

```

To me this logical flow is more clear as it is checking for the edge cases first and handling those before rendering the presumably default selector component. Assuming this works I would reuse the logic for your other selector components as well

src/js/analyses/components/Create/Selected.js
```diff
@@ -0,0 +1,53 @@
...
+const SelectedItem = styled(BoxGroupSection)`
+    display: flex;
+    flex-direction: row;
+`;
+
+const FormattedIcon = styled(Icon)`
+    margin-left: auto;
+`;
+
+export const SelectedAnalysesObject = ({ selected, type, clearSelected, formattedLine }) => {
```
ReeceHoffmann: I would personally avoid calling a parameter exactly `type` as that has a specific HTML meaning in the context of inputs. Probably not of the utmost importance here, but I would go with something like `resourceType`

src/js/analyses/components/Create/Selected.js
```diff
@@ -0,0 +1,53 @@
...
+                    StyledReferenceSelectorItem
+                    onClick={() => clearSelected(item)}
+                />
+            </SelectedItem>
+        );
+    });
+    return (
+        <SelectedContainer type={type}>
+            {!selected.length && <AllOrNoneSelectedBox type={type}>No {type} selected</AllOrNoneSelectedBox>}
+            <SelectedItemsContainer type={type}>{formattedSelected}</SelectedItemsContainer>
```
ReeceHoffmann:
```suggestion
            <SelectedItemsContainer>{formattedSelected}</SelectedItemsContainer>
```

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
+    background-color: white;
+    outline: 1px solid;
+    outline-color: ${props => props.theme.color.greyLight};
 `;
 
+const compare = (option1, option2) => {
+    return option1.reference.id === option2;
+};
+
+const formattedLine = item => {
```
ReeceHoffmann: 1. Should be a React component
2. No need to be so generic since you are defining a custom formatting function for every component. ie:
      -  item => index
      -  formattedLine => formattedReference

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
+    return (
+        <>
+            <NameOverflowItem type="references">{item.reference.name}</NameOverflowItem>
+            <span>
+                Index Version <Label>{item.version}</Label>
+            </span>
+        </>
+    );
+};
+
```
ReeceHoffmann:
```suggestion
const formattedLine = item =>  (
        <>
            <NameOverflowItem type="references">{item.reference.name}</NameOverflowItem>
            <span>
                Index Version <Label>{item.version}</Label>
            </span>
        </>
    );


```

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
 export const ReferenceSelector = ({ hasError, indexes, selected, onChange }) => {
-    const referenceComponents = map(indexes, index => (
-        <ReferenceSelectorItem
-            key={index.reference.id}
-            id={index.reference.id}
-            name={index.reference.name}
-            version={index.version}
-        />
-    ));
+    const clearSelected = object => {
```
ReeceHoffmann:
```suggestion
    const clearSelected = selectedReference => { //or selectedIndex, whichever is really getting passed here
```
Since you are defining a custom clearSelected for each resource type you can afford to be less generic in your naming

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
+        onChange(selected.filter(reference => reference !== object.reference.id));
+    };
+
+    const selectedReferenceItems = intersectionWith(indexes, selected, compare);
```
ReeceHoffmann:
```suggestion
    const selectedReferences = intersectionWith(indexes, selected, compare);
```

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
+    const [results, term, setTerm] = useFuse(indexes, ["reference.name"], [1]);
 
-    if (!referenceComponents.length) {
+    const formattedResults = results.map(result => (result.id ? result : result.item));
+    const references = difference(formattedResults, selectedReferenceItems);
```
ReeceHoffmann:
```suggestion
    const unselectedReferences = difference(formattedResults, selectedReferenceItems);
```

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
+                    {selectedReferenceItems.length < indexes.length && (
+                        <MultiSelector
+                            error={hasError && "Reference(s) must be selected"}
+                            noun="references"
+                            indexes={results}
+                            selected={selected}
+                            onChange={onChange}
+                            overflow={false}
+                        >
+                            <SampleInputContainer>
```
ReeceHoffmann:
```suggestion
                            <SampleSearchbarContainer>
```

src/js/subtraction/components/Selector.js
```diff
@@ -1,28 +1,121 @@
...
+import { AllSelectedBox, SelectorItem } from "../../analyses/components/Create/WorkflowSelector";
 
-export const SampleSubtractionSelector = ({ name, noun, selected, subtractions, onChange }) => {
-    const subtractionComponents = map(subtractions, ({ name, id }) => (
-        <MultiSelectorItem key={id} name={name} value={id} id={id}>
-            <span>{name}</span>
-        </MultiSelectorItem>
-    ));
-    if (!subtractions.length) {
+export const SampleInputContainer = styled.div`
```
ReeceHoffmann:
```suggestion
export const SearchbarContainer = styled.div`
```
(This change assumes `SampleInputContainer`  really does only get used as a SearchBar. Sample should probably be dropped from the name either way as I don't think there is anything sample specific about this component)

src/js/subtraction/components/Selector.js
```diff
@@ -1,28 +1,121 @@
...
+                                />
+                                <IconContainer>
+                                    <Icon name="times fa-fw" onClick={() => setTerm("")} aria-label="clear" />
+                                </IconContainer>
+                            </SampleInputContainer>
+                        )}
+                        <ItemComponentsContainer>{subtractionComponents}</ItemComponentsContainer>
+                    </MultiSelector>
+                )}
+            </>
```
ReeceHoffmann: Why has this logic been separated from `SubtractionSelector`? Seems to me the `Selector.js` could be merged into `SampleSelector.js`

src/js/base/MultiSelector.js
```diff
@@ -31,18 +31,22 @@ export const MultiSelectorItem = ({ children, className, name, value }) => {
...
+            {requireCheckbox && <Checkbox checked={active} />} {children}
         </StyledMultiSelectorItem>
     );
 };
 
-const MultiSelectorList = styled(BoxGroup)`
+export const MultiSelectorList = styled(BoxGroup)`
     ${props => (props.error ? `border-color: ${props.theme.color.red};` : "")};
-    max-height: 165px;
+    max-height: 160px;
```
ReeceHoffmann: Why make this change? Generally it is better to restyle where you specifically need it rather than changing the base component unless there is a reason you want it globally changed.

src/js/base/MultiSelector.js
```diff
@@ -31,18 +31,22 @@ export const MultiSelectorItem = ({ children, className, name, value }) => {
...
     overflow-y: auto;
 `;
 
-export const MultiSelector = ({ children, error, noun, selected, onChange }) => {
+const NoOverflowList = styled(BoxGroup)`
+    ${props => (props.error ? `border-color: ${props.theme.color.red};` : "")};
+`;
```
ReeceHoffmann: What problems was this box introduced to solve? I wonder if this might be more cleanly dealt with by allowing a new `maxItemHeight` prop to be passed to `Multiselector` and have that define the max height.

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
+
+    const selectedReferenceItems = intersectionWith(indexes, selected, compare);
+    const [results, term, setTerm] = useFuse(indexes, ["reference.name"], [1]);
 
-    if (!referenceComponents.length) {
+    const formattedResults = results.map(result => (result.id ? result : result.item));
+    const references = difference(formattedResults, selectedReferenceItems);
+
+    const referenceComponents = references.map(reference => {
         return (
```
ReeceHoffmann:
```suggestion
    const referenceComponents = references.map(reference => (
```

src/js/analyses/components/Create/ReferenceSelector.js
```diff
@@ -1,58 +1,119 @@
...
+                            indexes={results}
+                            selected={selected}
+                            onChange={onChange}
+                            overflow={false}
+                        >
+                            <SampleInputContainer>
+                                <Input
+                                    value={term}
+                                    placeholder="Filter References"
+                                    aria-label="Filter items"
```
ReeceHoffmann:
```suggestion
                                    aria-label="Filter References"
```

src/js/analyses/components/Create/Selected.js
```diff
@@ -0,0 +1,53 @@
...
+    flex-direction: row;
+`;
+
+const FormattedIcon = styled(Icon)`
+    margin-left: auto;
+`;
+
+export const SelectedAnalysesObject = ({ selected, type, clearSelected, formattedLine }) => {
+    const formattedSelected = selected.map(item => {
+        const key = item.id ? item.id : item;
```
ReeceHoffmann: Double check whether you need this check here. If all the parent components have already removed the fused formatting before passing selected you might not need this check

src/js/analyses/components/Create/Selected.js
```diff
@@ -0,0 +1,53 @@
...
+        return (
+            <SelectedItem key={key} name={key} value={key}>
+                {formattedLine(item)}
+
+                <FormattedIcon
+                    aria-label={`remove selected ${type}`}
+                    name="times fa-fw"
+                    color="black"
+                    StyledReferenceSelectorItem
```
ReeceHoffmann: Delete

src/js/analyses/components/Create/Selected.js
```diff
@@ -0,0 +1,53 @@
...
+
+const FormattedIcon = styled(Icon)`
+    margin-left: auto;
+`;
+
+export const SelectedAnalysesObject = ({ selected, type, clearSelected, formattedLine }) => {
+    const formattedSelected = selected.map(item => {
+        const key = item.id ? item.id : item;
+        return (
+            <SelectedItem key={key} name={key} value={key}>
```
ReeceHoffmann: What is name doing in this context?

src/js/analyses/components/Create/SubtractionSelector.js
```diff
@@ -1,20 +1,69 @@
...
+
+    const selectedSubtractionObjects = intersectionWith(subtractions, value, compare);
+
+    const [results, term, setTerm] = useFuse(subtractions, ["name"], [1]);
+    const formattedResults = results.map(result => (result.name ? result : result.item));
+
+    const removeSelected = () => {
+        return formattedResults.filter(result => !value.includes(result.id));
+    };
```
ReeceHoffmann: Why not just generate the list of results then pass that? Also just FYI, here the function is getting it reference to `formattedResults` from function scope, so passing it when you call the function is redundant in this case.

src/js/analyses/components/Create/WorkflowSelector.js
```diff
@@ -3,6 +3,81 @@ import PropTypes from "prop-types";
...
+
+export const AllSelectedBox = ({ type }) => {
+    return (
+        <AllSelectedContainer type={type}>
+            <AllOrNoneSelectedBox requireBorder={true} type={"workflows"}>
+                All {type} selected
+            </AllOrNoneSelectedBox>
+        </AllSelectedContainer>
+    );
+};
```
ReeceHoffmann:
```suggestion
export const AllSelectedBox = ({ type }) => (
        <AllSelectedContainer type={type}>
            <AllOrNoneSelectedBox requireBorder={true} type={"workflows"}>
                All {type} selected
            </AllOrNoneSelectedBox>
        </AllSelectedContainer>
    );
```

src/js/analyses/components/Create/WorkflowSelector.js
```diff
@@ -3,6 +3,81 @@ import PropTypes from "prop-types";
...
+            <GridTitle>
+                <LargeLabel>{analysesType}</LargeLabel>
+            </GridTitle>
+            <GridContainer>
+                <label>Available</label>
+                <label>Selected</label>
+            </GridContainer>
+        </>
+    );
+};
```
ReeceHoffmann:
```suggestion
export const CreateHeader = ({ analysesType }) => (
        <>
            <GridTitle>
                <LargeLabel>{analysesType}</LargeLabel>
            </GridTitle>
            <GridContainer>
                <label>Available</label>
                <label>Selected</label>
            </GridContainer>
        </>
    );
```

src/js/analyses/components/Create/WorkflowSelector.js
```diff
@@ -17,24 +92,45 @@ export const getCompatibleWorkflows = (dataType, hasHmm) => {
...
+        onSelect(workflows.filter(workflow => workflow !== object));
+    };
+
+    const workflowItems = map(getCompatibleWorkflows(dataType, hasHmm), workflow =>
+        !workflows.includes(workflow) ? (
+            <SelectorItem key={workflow} name={workflow} value={workflow} requireCheckbox={false} id={workflow}>
+                {getWorkflowDisplayName(workflow)}
+            </SelectorItem>
+        ) : null
+    );
```
ReeceHoffmann:
```suggestion
        workflows.includes(workflow)  && (
            <SelectorItem key={workflow} name={workflow} value={workflow} requireCheckbox={false} id={workflow}>
                {getWorkflowDisplayName(workflow)}
            </SelectorItem>
        ) 
    );
```

src/js/analyses/components/Create/WorkflowSelector.js
```diff
@@ -17,24 +92,45 @@ export const getCompatibleWorkflows = (dataType, hasHmm) => {
 };
 
 export const WorkflowSelector = ({ dataType, hasError, hasHmm, workflows, onSelect }) => {
```
ReeceHoffmann:
```suggestion
export const WorkflowSelector = ({ dataType, hasError, hasHmm, selectedWorkflows, onSelect }) => {
```
(or other name that makes it obvious that these are the selected workflows rather than just a generic list of workflows)

src/js/analyses/components/Create/WorkflowSelector.js
```diff
@@ -17,24 +92,45 @@ export const getCompatibleWorkflows = (dataType, hasHmm) => {
...
+                        selected={workflows}
+                        onChange={onSelect}
+                        overflow={false}
+                    >
+                        <WorkflowItemsContainer>{workflowItems}</WorkflowItemsContainer>
+                    </MultiSelector>
+                )}
+                {getCompatibleWorkflows(dataType, hasHmm).length === workflows.length && (
+                    <AllSelectedBox type="workflows" />
+                )}
```
ReeceHoffmann: Like `ReferenceSelector` I would suggest a similar reformatting of logic:

```javascript
const compatibleWorkflows = getCompatibleWorkflows(dataType, hasHmm);
let workflowSelector;
if(comaptibleWorkflows.length === workflows.length){
 workflowSelector = <AllSelected/>
}else{
 workflowSelector = ... // default selector logic here
}


```

src/js/base/MultiSelector.js
```diff
@@ -18,7 +18,7 @@ const StyledMultiSelectorItem = styled(SelectBoxGroupSection)`
     user-select: none;
 `;
 
-export const MultiSelectorItem = ({ children, className, name, value }) => {
+export const MultiSelectorItem = ({ children, className, name, value, requireCheckbox }) => {
```
ReeceHoffmann:
```suggestion
export const MultiSelectorItem = ({ children, className, name, value, displayCheckbox }) => {
```

src/js/base/MultiSelector.js
```diff
@@ -56,7 +60,13 @@ export const MultiSelector = ({ children, error, noun, selected, onChange }) =>
...
             }}
         >
-            {children.length ? <MultiSelectorList error={error}>{children}</MultiSelectorList> : content}
+            {children && overflow !== false ? (
+                <MultiSelectorList error={error}>{children}</MultiSelectorList>
+            ) : children && overflow === false ? (
+                <NoOverflowList error={error}>{children}</NoOverflowList>
+            ) : (
+                content
+            )}
```
ReeceHoffmann: Based on what I am seeing above I would try to get rid of the `NoOverflowList` component, but if not I would rework this to be a bit more readable. Nesting ternary operators can get confusing to read fast.

src/js/subtraction/components/Selector.js
```diff
@@ -1,28 +1,121 @@
...
         );
     }
-    return (
-        <MultiSelector name={name} noun={noun} selected={selected} onChange={onChange}>
-            {subtractionComponents}
-        </MultiSelector>
-    );
+    return <CreateNoneFoundBox type="subtractions" link="/subtractions" />;
```
ReeceHoffmann: I would rework the logic here in a way philosophically similar to the other selectors. ie:

```javascript

if (!subtractionOptions.length){
   return <NoneFound>
}
if(subtractionOptions.length === selected.length){
   return <AllFound>
}

return ... //Default there are items to select case

```

ReeceHoffmann (changes requested):
Overall I like a lot of how things are looking now, for the most part my suggestions are tweaking details or localized logic rather than big picture changes. Not included in the comments below, but you need to restore your package.lock now that material-ui core has been removed. (Either via git restore or rerun `npm i`).

src/js/subtraction/components/Selector.js
elanjm12: The logic was originally separated before I began working on the issue, so I left it separated how it was before. But other than that there's no reason. I agree it may make sense to merge them.

src/js/base/MultiSelector.js
elanjm12: I'm not sure why this change showed up. It is showing max-height: 160px in my current branch for the issue. Either way should be resolved now.

src/js/base/MultiSelector.js
elanjm12: No overflowList has been removed

src/js/analyses/components/Create/Selected.js
elanjm12: This check is necessary not for the fuse formatting as that is removed already. But the data is formatted differently in the case it is a workflow. There is no id property for workflows.

src/js/analyses/components/Create/Selected.js
elanjm12: aria-label and name needed. Color and styledReferenceSelectorItem have been removed.

src/js/analyses/components/Create/WorkflowSelector.js
elanjm12: workflows has been renamed to selectedWorkflows. Negation still must be present since the rendering is occuring if the workflow is not in fact selected. The other change has been implemented.

src/js/subtraction/components/Selector.js
elanjm12: Selector has been merged into SubtractionSelector

src/js/analyses/components/Create/ReferenceSelector.js
elanjm12: Logic for all selectors now follows this pattern. I will look at parsing it out now so that the logic can be reused.

src/js/analyses/components/Create/Selected.js
ReeceHoffmann: Ah my bad, I only meant to highlight `StyledReferenceSelectorItem` which looks to me like leftover code.

PR #132, opened 2022-06-20
Refresh job counts when updates received

ReeceHoffmann:
Updated: 22/06/2022

### Changes:
 - Makes API request to refresh jobs list whenever a jobs related websocket message is received (insert, update or remove) 
 - Removes websocket reducer logic in Jobs reducer
    - Functionality made redundant since whole list is refreshed when a websocket update is received 
 - Also removes gray background from job status bar when no jobs exist.

### DIFF
--- a/src/js/jobs/reducer.js
+++ b/src/js/jobs/reducer.js
@@ -1,20 +1,11 @@
 import { createReducer } from "@reduxjs/toolkit";
-import { assign } from "lodash-es";
-import {
-    ARCHIVE_JOB,
-    FIND_JOBS,
-    GET_JOB,
-    GET_LINKED_JOB,
-    WS_INSERT_JOB,
-    WS_REMOVE_JOB,
-    WS_UPDATE_JOB
-} from "../app/actionTypes";
-import { insert, remove, update, updateDocuments } from "../utils/reducers";
+import { assign, sortBy } from "lodash-es";
+import { ARCHIVE_JOB, FIND_JOBS, GET_JOB, GET_LINKED_JOB } from "../app/actionTypes";
+import { remove } from "../utils/reducers";
 
 export const initialState = {
     documents: null,
     term: "",
-    page: 0,
     total_count: 0,
     detail: null,
     filter: "",
@@ -25,15 +16,6 @@ export const initialState = {
 
 export const jobsReducer = createReducer(initialState, builder => {
     builder
-        .addCase(WS_INSERT_JOB, (state, action) => {
-            return insert(state, action.payload, "created_at");
-        })
-        .addCase(WS_UPDATE_JOB, (state, action) => {
-            return update(state, action.payload, "created_at");
-        })
-        .addCase(WS_REMOVE_JOB, (state, action) => {
-            return remove(state, action.payload);
-        })
         .addCase(GET_LINKED_JOB.SUCCEEDED, (state, action) => {
             assign(state.linkedJobs, { [action.payload.id]: action.payload });
         })
@@ -41,7 +23,7 @@ export const jobsReducer = createReducer(initialState, builder => {
             state.term = action.payload.term;
         })
         .addCase(FIND_JOBS.SUCCEEDED, (state, action) => {
-            return updateDocuments(state, action.payload, "created_at");
+            return { ...state, ...action.payload, documents: sortBy(action.payload.documents, "created_at") };
         })
         .addCase(GET_JOB.REQUESTED, state => {
             state.detail = null;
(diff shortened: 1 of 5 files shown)


src/js/jobs/sagas.js
```diff
@@ -47,5 +75,8 @@ export function* watchJobs() {
     yield takeEvery(CANCEL_JOB.REQUESTED, cancelJob);
     yield takeEvery(ARCHIVE_JOB.REQUESTED, archiveJob);
     yield takeLatest(WS_UPDATE_JOB, wsUpdateJob);
+    yield takeLatest(WS_INSERT_JOB, wsInsertJob);
+    yield takeLatest(WS_REMOVE_JOB, wsRemoveJob);
```
igboyes:
```suggestion
    yield takeLatest([WS_INSERT_JOB, WS_REMOVE_JOB], getJobCount);
```

src/js/jobs/sagas.js
```diff
@@ -18,6 +30,22 @@ export function* wsUpdateJob(action) {
...
+export function* wsRemoveJob() {
+    yield put(getJobCountAction());
+}
+
+export function* getJobCount() {
+    const location = yield select(getLocation);
+    if (location.pathname === "/jobs") {
+        yield apiCall(jobsAPI.getJobCount, {}, GET_JOB_COUNT);
+    }
 }
```
igboyes: Combine into single function

src/js/jobs/sagas.js
```diff
@@ -47,5 +75,8 @@ export function* watchJobs() {
     yield takeEvery(CANCEL_JOB.REQUESTED, cancelJob);
     yield takeEvery(ARCHIVE_JOB.REQUESTED, archiveJob);
     yield takeLatest(WS_UPDATE_JOB, wsUpdateJob);
```
igboyes, line 63: This needs to update counts as well. What if a job changes from waiting to preparing?

igboyes (changes requested):
Why not just dispatch `FIND_JOBS.REQUESTED` again instead of having separate code for getting the counts?

ReeceHoffmann:
> Why not just dispatch `FIND_JOBS.REQUESTED` again instead of having separate code for getting the counts?

That would work, but I ended up deciding against as I would have to request the server to filter by job state and term on every request when I really don't need to update the job list itself and only care about the counts. (`FIND_JOBS.REQUESTED` when returned will result in `FIND_JOBS.SUCCEEDED` which then ends up merging the current documents with the ones retrieved, so the filtering would be needed on every request) I can change it, but it seemed like extra unnecessary server load to me.

src/js/jobs/sagas.js
ReeceHoffmann: Useful bit of syntax, will update to this and remove the wsInsertJob and wsRemoveJob functions.

src/js/jobs/sagas.js
ReeceHoffmann: I assume your talking about `WS_UPDATE_JOB`? If so, it already does by pushing a get `getJobCount` action. Probably going to change how this works on next push though

Update: `WS_UPDATE_JOB`  does still update counts but it is no longer done as stated above.

src/js/jobs/sagas.js
ReeceHoffmann: Updated

src/js/jobs/sagas.js
ReeceHoffmann: Merged

src/js/jobs/reducer.js
```diff
@@ -25,23 +16,14 @@ export const initialState = {
...
-        })
         .addCase(GET_LINKED_JOB.SUCCEEDED, (state, action) => {
             assign(state.linkedJobs, { [action.payload.id]: action.payload });
         })
         .addCase(FIND_JOBS.REQUESTED, (state, action) => {
             state.term = action.payload.term;
         })
         .addCase(FIND_JOBS.SUCCEEDED, (state, action) => {
-            return updateDocuments(state, action.payload, "created_at");
+            return { ...state, ...action.payload, documents: sortBy(action.payload.documents, "create_at") };
```
igboyes:
```suggestion
            return { ...state, ...action.payload, documents: sortBy(action.payload.documents, "created_at") };
```

igboyes (changes requested):

src/js/jobs/reducer.js
ReeceHoffmann: Ah, that was not gonna work great, fixed

igboyes (approved):

PR #137, opened 2022-07-06
fix: remove unused modal code

elanjm12:
Remove ModalBodyOverlay from Modal.js file.


igboyes (changes requested):
Please change commit type to `chore`. This change doesn't impact the user and shouldn't trigger a release.

PR #138, opened 2022-07-08
Update user data modeling

ReeceHoffmann:
Changes:

  - Adds object modeling for `user` objects returned by GET requests to `/users`
     - Model enforces formatting of `user` object `groups` array items to object format. (`{id: String}`)
       - If passed groups array contains string they will be assumed to be group ids and reformatted to object format at model creation time
  - Changes tests as required to accommodate modeled user data

### DIFF
--- a/src/js/users/models.js
+++ b/src/js/users/models.js
@@ -1,0 +1,33 @@
+import { map } from "lodash-es";
+import { ArrayModel, Model, ObjectModel } from "objectmodel";
+
+export const Permissions = Model({
+    cancel_job: Boolean,
+    create_ref: Boolean,
+    create_sample: Boolean,
+    modify_hmm: Boolean,
+    modify_subtraction: Boolean,
+    remove_file: Boolean,
+    remove_job: Boolean,
+    upload_file: Boolean
+});
+
+/**
+ * @todo Remove group reformatting logic once API provides groups as objects be default
+ */
+export class Groups extends ArrayModel([String, { id: String }]) {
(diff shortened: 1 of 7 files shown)


src/js/users/models.js
```diff
@@ -0,0 +1,34 @@
+import { map } from "lodash-es";
+import { ArrayModel, Model, ObjectModel } from "objectmodel";
+
+export const permissionsModel = Model({
```
igboyes:
```suggestion
export const Permissions = Model({
```

src/js/users/models.js
```diff
@@ -0,0 +1,34 @@
...
+    upload_file: Boolean
+});
+
+/**
+ * @todo Remove group reformatting logic once API provides groups as objects be default
+ */
+export class groupsModel extends ArrayModel([String, { id: String }]) {
+    constructor(groups) {
+        const reformattedGroups = map(groups, item => (typeof item === "string" ? item : { id: item }));
+        super(reformattedGroups);
```
igboyes:
```suggestion
        super(map(groups, item => (typeof item === "string" ? item : { id: item })));
```

src/js/users/models.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
+
+export const userModel = ObjectModel({
```
igboyes:
```suggestion
export const User = ObjectModel({
```

igboyes (changes requested):

PR #139, opened 2022-07-08
Update primaryGroup component for new API

ReeceHoffmann:
Changes:
  - Updates `PrimaryGroup` component to call `onSetPrimaryGroup` with `primary_group=null` when None is selected
  - `PrimaryGroup` now defaults to displaying None as the selected option when the passed `PrimaryGroup` is not truthy (ex: null or undefined)
  - Set `primary_group` as optional in the `User` model
  - convert `PrimaryGroup` test to RTL


igboyes:
Should be a `fix`. This affects what API versions the UI is compatible with.

igboyes (changes requested):
> update primaryGroup component for new API

More user-focused commit message: `maintain compatibility with upcoming API version 8.0.0`.

PR #141, opened 2022-07-14
fix: add upload speed and estimate time remaining

elanjm12:
Overall upload speed estimate and time remaining added. Tests added and existing tests changed to use react testing library.

Screenshot from 2022-07-14 10-36-48


igboyes (changes requested):
* I think the total size (eg. 31.5 GB) should be removed. It would only be informative if it had 'left to upload' after it and decreased over time as data was uploaded.
* Please reduce size of line 2 header items by one. I think the size would be `sm` in the theme. 
* Please reverse order of speed and remaining time. I think remaining time is the main number the user will want to know.
* Use the whole word seconds in the time remaining text.
* Put a space between the speed and units.
* Use `MB/s` for units.

elanjm12:
Additional styling changes completed

Screenshot from 2022-07-14 11-52-54

src/js/files/components/UploadItem.js
```diff
@@ -38,6 +38,7 @@ const UploadItemTitle = styled.div`
 `;
 
 const StyledFilename = styled.span`
```
igboyes:
```suggestion
const UploadItemName = styled.span`
```

src/js/files/components/UploadItem.js
```diff
@@ -38,6 +38,7 @@ const UploadItemTitle = styled.div`
...
+    margin-right: 5px;
```
igboyes: What is this doing?

src/js/files/components/UploadOverlay.js
```diff
@@ -36,15 +38,54 @@ const UploadOverlayList = styled(BoxGroupSection)`
...
+    let formattedTimeRemaining = "";
+    if (secondsToMinutes(remainingTime) === 0) {
+        formattedTimeRemaining = `${remainingTime.toFixed(0)} seconds remaining`;
+    } else if (secondsToMinutes(remainingTime) > 0 && secondsToMinutes(remainingTime) < 60) {
+        formattedTimeRemaining = `${secondsToMinutes(remainingTime)} minute(s) remaining`;
+    } else if (secondsToHours(remainingTime) > 0 && secondsToHours(remainingTime) <= 12) {
+        formattedTimeRemaining = `${secondsToHours(remainingTime)} hour(s) remaining`;
+    } else {
+        formattedTimeRemaining = " > 12hr remaining";
+    }
```
igboyes: Can this be taken care of by [`formatDuration`](https://date-fns.org/v2.28.0/docs/formatDuration)?

src/js/files/components/UploadOverlay.js
```diff
@@ -36,15 +38,54 @@ const UploadOverlayList = styled(BoxGroupSection)`
     padding: 0;
 `;
 
+const UploadInfo = styled.div`
+    padding-bottom: 5px;
+`;
+
+const FormattedRemainingUploadTime = styled.div`
```
igboyes: Remember convention:
```suggestion
const StyledUploadInformation = styled.div`
```

src/js/files/components/UploadOverlay.js
```diff
@@ -36,15 +38,54 @@ const UploadOverlayList = styled(BoxGroupSection)`
     padding: 0;
 `;
 
+const UploadInfo = styled.div`
+    padding-bottom: 5px;
```
igboyes: I would put a `margin-top` on `UploadInformation` instead of defining a new component here.

src/js/files/components/UploadOverlay.js
```diff
@@ -36,15 +38,54 @@ const UploadOverlayList = styled(BoxGroupSection)`
...
+`;
+
+const FormattedRemainingUploadTime = styled.div`
+    display: flex;
+    justify-content: space-between;
+    font-size: ${getFontSize("md")};
+`;
+
+const UploadInformation = ({ remaining, uploadSpeed }) => {
```
igboyes: More accurate name would be `UploadTiming` or similar.

igboyes (changes requested):
Additionally, change `font-weight` for second line of header to normal.

Can you also put a space before the unit in the `UploadItem` @elanjm12? Thanks.

src/js/files/components/UploadItem.js
elanjm12: sorry, this is no longer necessary after other changes. I've removed it

src/js/files/components/UploadOverlay.js
elanjm12: yes

igboyes (approved):

PR #145, opened 2022-07-22
Improve groups view

elanjm12:
Draft pull request.

Still additional styling changes to be made and tests must be written.

On initial load of groups view, endless loading sign appears.

permissions render one step behind what they should.

Creating new group does not focus on newly created group.

And also commits must be combined into 1 before final push.


ReeceHoffmann:
Force pushed to enable merging via rebase and address `commitlint` problems. Additional commits added to improve/update logic for compatibility with the latest API. Current look:
image

igboyes (changes requested):
I have brought this PR up to date with `main`. The merge wasn't super clean, so please tidy things up.

The bones of these changes are good, but I find the UI is lacks hierarchy and is not as consistent with the app as it could be.

Take a look at this wireframe. Let me know if you'd like to discuss.

https://www.figma.com/file/BqiunpsPEw9O6YdF5LXuFa/Virtool?node-id=201%3A26

elanjm12:
Update styling to better match figma design

Screenshot from 2022-10-21 13-13-16

elanjm12:
Updated Groups View styling

Screenshot from 2022-10-27 13-01-40

ReeceHoffmann:
Overall it is starting to look really good. However there is a bug and a few visual tweaks I would change:

1. If there are no groups then it currently goes into a infinite loading state. I suspect this is because the value of `loading` is dependent on there being a selected group, and when there are no groups, there can be no selected group.  
2. The deletion banner as the header make the deletion function a bit too front and center and doesn't work great as a header I would at least:
   - Swap the order of the elements so that the delete banner is below the permissions and members sections.
   - Change the text of the delete banner to be more descriptive. ie: something like: "Permanently delete {group name}"
   - Display the name of the group as a dedicated prominent header for the section.

elanjm12:
Styling changes to address review from Reece. Additionally endless loading on no group bug fixed. activeGroup is set to null if the data is still loading. Otherwise activeGroup gets the value of an empty array if no groups exist or the proper activeGroup value if groups do exist.

elanjm12:
Ready for review. The bugs are fixed. There is still a minor detail that could be improved as a subsequent issue. When creating the first group a very brief loading symbol is shown before the group is displayed.

igboyes (approved):

PR #148, opened 2022-07-25
fix: display message when uploads finishing up loading

elanjm12:
Screenshot from 2022-07-25 15-15-56

### DIFF
--- a/src/js/files/components/UploadOverlay.js
+++ b/src/js/files/components/UploadOverlay.js
@@ -1,6 +1,6 @@
 import React from "react";
 import styled from "styled-components";
-import { map, reject, reverse, sum } from "lodash-es";
+import { map, reject, reverse, every, sumBy } from "lodash-es";
 import { connect } from "react-redux";
 import { Badge, BoxGroup, BoxGroupHeader, BoxGroupSection } from "../../base";
 import { UploadItem } from "./UploadItem";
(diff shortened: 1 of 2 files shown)


src/js/files/components/UploadOverlay.js
```diff
@@ -94,15 +94,24 @@ export const UploadOverlay = ({ uploads }) => {
     if (uploads.length) {
         const totalRemainingUploadTime = sum(uploads.map(upload => upload.remaining));
         const overallUploadSpeed = sum(uploads.map(upload => upload.uploadSpeed));
```
igboyes: Look at `sumBy` or `reduce` to avoid a second iteration. I missed this in the last review.

src/js/files/components/UploadOverlay.js
```diff
@@ -94,15 +94,24 @@ export const UploadOverlay = ({ uploads }) => {
...
-
         const uploadComponents = map(uploads, upload => <UploadItem key={upload.localId} {...upload} />);
 
+        const totalProgress = sum(uploads.map(upload => upload.progress)) / uploads.length;
```
igboyes: You don't actually need to do math here.

You are trying to find out if all of the uploads have `progress === 100`. This will not be `true` if even one upload has `progress !== 100`. That means there is a good chance we don't have to iterate through the entire array (twice) like you are doing above.

Lodash has a convenient function for this called [`every()`](https://lodash.com/docs/4.17.15#every).

src/js/files/components/UploadOverlay.js
```diff
@@ -94,15 +94,24 @@ export const UploadOverlay = ({ uploads }) => {
...
+
+        let uploadProgressBar;
+
+        if (totalProgress === 100) {
+            uploadProgressBar = <StyledUploadInformation>Finishing uploads</StyledUploadInformation>;
+        } else {
+            uploadProgressBar = <UploadTiming remaining={totalRemainingUploadTime} uploadSpeed={overallUploadSpeed} />;
+        }
```
igboyes: Use a ternary in the `return` statement.

igboyes (changes requested):

igboyes (approved):

PR #149, opened 2022-07-26
fix: remove sentry error reporting

elanjm12:


src/js/administration/components/Server.js
```diff
@@ -1,15 +1,9 @@
...
-    <>
-        <NarrowContainer>
-            <Sentry />
-        </NarrowContainer>
-        <NarrowContainer>
-            <API />
-        </NarrowContainer>
-    </>
+    <NarrowContainer>
+        <API />
```
igboyes: Deepsource is complaining about this not being Pascal case. I looked it up and the recommendation is to use `Api` instead.

Can you update this?

igboyes (changes requested):

igboyes (approved):

PR #151, opened 2022-07-29
fix: in progress typescript

elanjm12:
in progress. Bugs exist


package.json
```diff
@@ -58,6 +86,8 @@
     "semver": "^7.3.5",
     "styled-components": "^5.3.3",
     "superagent": "^7.1.1",
+    "ts-loader": "^9.3.1",
+    "typescript": "^4.7.4",
```
igboyes: Both of these should be in `devDependencies`. Uninstall and reinstall with `--save-dev`.

package.json
```diff
@@ -73,6 +103,9 @@
     "@testing-library/react": "^12.1.4",
     "@testing-library/react-hooks": "^7.0.2",
     "@testing-library/user-event": "^13.5.0",
+    "@types/babel__generator": "^7.6.4",
+    "@types/babel__template": "^7.4.1",
+    "@types/node": "^18.6.3",
```
igboyes: Uninstall. Not referenced in code.

tsconfig.json
```diff
@@ -0,0 +1,107 @@
+{
+  "compilerOptions": {
```
igboyes, line 2: Remove all commented out options.

src/TypescriptTest.js
```diff
@@ -0,0 +1,2 @@
+"use strict";
```
igboyes: Remove and use typing in existing module.

package.json
```diff
@@ -11,6 +11,34 @@
     "@reach/menu-button": "^0.16.2",
     "@reduxjs/toolkit": "^1.8.0",
     "@sentry/browser": "^6.18.2",
+    "@types/aria-query": "^5.0.0",
```
igboyes: 1. I don't think any of these `@types` packages need to be installed since we aren't referencing them in our code.
2. Most likely `@types` should be installed as dev dependencies as they aren't distributed in the bundle.

webpack.config.babel.js
```diff
@@ -9,9 +9,9 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /\.js$/,
+        test: /\.js|tsx?$/,
         exclude: /(node_modules)/,
```
igboyes: Instead of `exclude`, change this to `include` only `src/js`.

igboyes (changes requested):

elanjm12:
Typescript is working correctly with no errors. Two issues persist: 1) testing is not working after the removal of babel. 2) reloading does not work when typescript files are changed. eg: if a type error exists in a file this will be displayed as it should. If the error is corrected the server must be restarted for the changes in the typescript file to be noticed and for the error to go away.

package.json
```diff
@@ -37,6 +37,7 @@
     "marked": "^4.0.12",
     "minimist": "^1.2.6",
     "normalize.css": "^8.0.1",
+    "npm": "^8.19.2",
```
igboyes: I don't think should be in `package.json`.

package.json
```diff
@@ -26,6 +25,7 @@
     "d3-transition": "^3.0.1",
     "date-fns": "^2.28.0",
     "ejs": "^3.1.8",
+    "esbuild": "^0.15.8",
```
igboyes: Is this necessary? If so, it should be in `devDependencies`.

package.json
```diff
@@ -47,6 +48,7 @@
     "react-dom": "^17.0.2",
     "react-dropzone": "^12.0.4",
     "react-helmet": "^6.1.0",
+    "react-hot": "^0.0.1",
```
igboyes: Necessary?

package.json
```diff
@@ -95,19 +90,55 @@
     "eslint-plugin-vitest-globals": "^1.2.0",
     "eslint-webpack-plugin": "^3.1.1",
     "html-webpack-plugin": "^5.5.0",
+    "nodemon": "^2.0.19",
```
igboyes: Necessary?

package.json
```diff
@@ -95,19 +90,55 @@
...
     "terser-webpack-plugin": "^5.3.1",
+    "ts-loader": "^9.3.1",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.7.4",
+    "typescript-register": "^1.1.0",
     "webpack": "^5.70.0",
     "webpack-cli": "^4.9.2",
-    "webpack-dev-server": "^4.7.4"
+    "webpack-dev-server": "^4.7.4",
+    "yup": "^0.32.11"
```
igboyes: Should be in `dependencies`, not `devDependencies`.

src/js/base/CloseButton.tsx
```diff
@@ -7,3 +7,6 @@ const StyledButton = styled(Icon)`
 `;
 
 export const CloseButton = ({ onClick }) => <StyledButton name="times" onClick={onClick} />;
+
+
+
```
igboyes: Reduce to one line.

src/js/base/index.js
```diff
@@ -13,7 +13,7 @@ export * from "./ButtonGroup";
 export * from "./ButtonToolbar";
 export * from "./Checkbox";
 export * from "./Circle";
-export * from "./CloseButton";
+export * from "./CloseButton.tsx";
```
igboyes: Why is this necessary?

igboyes (changes requested):

src/js/base/CloseButton.tsx
elanjm12: fixed. Sorry I didn't realize that prettier was not configured for typescript files

igboyes (changes requested):
Please review `package.json`.

Only packages that are used in the browser or server image should be in `dependencies`. Tooling packages must be in `devDependencies`.

elanjm12:
Marked as draft to resolve package.json and also an error occuring in production

elanjm12:
Production is working

igboyes (approved):

PR #152, opened 2022-07-29
fix: jobs checkboxes change when clicked

elanjm12:


igboyes:
I changed commit message to imperative: `fix: repair broken job status filters`.

PR #156, opened 2022-08-12
fix: analyses view handle long names

elanjm12:
Long names do not cause issues in subtraction and index selectors anymore. Bottom double borders removed.

Screenshot from 2022-08-12 14-46-21


igboyes (commented):
Good catch.

I have pushed another commit to avoid using fixed pixel widths for text overflow. Using `text-overlow` in flex children is an annoying edge case. I mostly play around with the `flex` and `min-width` properties until I get what I want.

https://github.com/virtool/virtool-ui/pull/156/commits/2dda7befb16c656acb96e351ddb83ff49e43973f

igboyes (approved):
Let me know if this is good to merge.

elanjm12:
The changes look good to me

elanjm12 (commented):
Looks good to me

PR #157, opened 2022-08-15
fix: create analyses add default subtractions badge

elanjm12:
Screenshot from 2022-08-15 11-14-02

### DIFF
--- a/src/js/analyses/components/Create/SubtractionSelector.js
+++ b/src/js/analyses/components/Create/SubtractionSelector.js
@@ -1,4 +1,4 @@
-import { filter, intersectionWith, xor } from "lodash-es";
+import { filter, intersectionWith, xor, includes } from "lodash-es";
 import PropTypes from "prop-types";
 import React from "react";
 import { useFuse } from "../../../base/hooks";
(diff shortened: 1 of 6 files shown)


src/js/analyses/components/Create/SubtractionSelector.js
```diff
@@ -28,14 +28,26 @@ export const SubtractionSelector = ({ subtractions, selected, onChange }) => {
                 <CreateAnalysisSelectorList
                     items={unselectedSubtractions}
                     render={({ id, name }) => (
-                        <SubtractionSelectorItem key={id} id={id} name={name} onClick={handleClick} />
+                        <SubtractionSelectorItem
+                            key={id}
+                            id={id}
+                            name={name}
+                            onClick={handleClick}
+                            isDefault={includes(defaultSubtractions, id) ? true : false}
```
igboyes: I'd like to extract this to a new select in `src/js/analyses/selectors` called `getAnalysisSubtractions()`.

Use reselect to write a selector that returns the subtractions with an added `default` field on each object. This will avoid calling `includes` (expensive) for every subtraction on every render.

There a more efficient solution than `includes`, but putting this behind reselect will be good enough.

igboyes (changes requested):

igboyes (dismissed):

igboyes (approved):

PR #163, opened 2022-08-24
Prevent excessive websocket requests for unauthenticated users

ReeceHoffmann:
**(NB: Dependent on PR 2662 in virtool to properly function)**

Changes:

  -  Allows UI to interpret custom error code for API when authentication fails and does not attempt to reconnect
  - Adds support for reconnecting to web socket upon the user logging back in


igboyes (changes requested):
Defining the connection status values as constants will prevent potential typo bugs:
```js
const INITIALIZING = "initializing";
...
```

igboyes (approved):

PR #169, opened 2022-09-06
hide official reference banner on download start

elanjm12:


src/js/references/components/List.js
```diff
@@ -8,46 +8,47 @@ import ReferenceItem from "./Item/Item";
...
+            </NarrowContainer>
+            <Clone />
+        </>
+    );
+};
 
 const mapStateToProps = state => ({
-    ...state.references,
+    documents: state.references.documents,
+    total_count: state.references.total_count,
```
igboyes:
```suggestion
    totalCount: state.references.total_count,
```

src/js/references/components/List.js
```diff
@@ -8,46 +8,47 @@ import ReferenceItem from "./Item/Item";
...
+    page: state.references.page,
+    pageCount: state.references.pageCount,
```
igboyes: You sure this should be `state.references.page_count`?

igboyes (changes requested):

igboyes (approved):

PR #171, opened 2022-09-08
Elanjm2/vir 1548 use vitest in ci

elanjm12:


elanjm12:
Ready for review.

Test files switched to have .jsx ending.
Tests converted where needed to be compatible with vitest.
jest.fn() => vi.fn()
github actions updated

.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
```
igboyes: Remove your `.vscode` directory from the PR. Add the folder to `.gitignore`.

igboyes (changes requested):
What does this mean: `ShallowWrapper {}`? Are you sure our snapshot are still intact?

package.json
```diff
@@ -12,6 +12,7 @@
     "@reduxjs/toolkit": "^1.8.0",
     "@sentry/browser": "^6.18.2",
     "@vitejs/plugin-react": "^2.1.0",
+    "@vue/test-utils": "^2.0.2",
```
igboyes: Should be dev dependencies.

igboyes (changes requested):

PR #172, opened 2022-09-17
fix: get favicon once

elanjm12:
It looks like most browsers automatically request the favicon. This is true with google chrome and was causing the second fetch. On Firefox the favicon was only ever fetched once. 

The fix results in the favicon being fetched only once in both chrome and firefox.

Screenshot from 2022-09-17 11-03-55

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -27,7 +27,7 @@
         };
 
     </script>
-    <link rel="shortcut icon" href="<%= htmlWebpackPlugin.files.favicon %>?v=2" type="images/x-icon"/>
+    <link rel="shortcut icon" href="/images/favicon.ico" type="image/x-icon"/>
 </head>
 
 <body>


src/index.html
```diff
@@ -27,7 +27,7 @@
         };
 
     </script>
-    <link rel="shortcut icon" href="<%= htmlWebpackPlugin.files.favicon %>?v=2" type="images/x-icon"/>
+    <link rel="shortcut icon" href="./images/favicon.ico" type="image/x-icon"/>
```
igboyes:
```suggestion
    <link rel="shortcut icon" href="/images/favicon.ico" type="image/x-icon"/>
```

igboyes (changes requested):

igboyes (approved):

PR #178, opened 2022-10-07
chore: create legend manually

elanjm12:
d3-svg-legend is only used for creating the quality distribution and nucleotide legends. The package has been removed and replaced with a custom legend.

Screenshot from 2022-10-07 09-44-14


igboyes (approved):
Great. Looks good to me!

PR #184, opened 2022-10-24
chore: add storybook for some base components

elanjm12:
Add storybook for Alert, Attribution, Box and RelativeTime. Future storybook additions will not be attached to a linear issue and will be handled individually.

### DIFF
--- a/src/js/base/stories/Alert.stories.js
+++ b/src/js/base/stories/Alert.stories.js
@@ -4,6 +4,11 @@ import { Alert } from "../Alert";
 export default {
     title: "base/Alert",
     component: Alert,
+    parameters: {
+        controls: {
+            exclude: "level"
+        }
+    },
     argTypes: {
         color: {
             options: ["red", "orange", "purple", "blue", "black"],
--- a/src/js/base/stories/Alert.stories.js
+++ b/src/js/base/stories/Alert.stories.js
@@ -1,0 +1,17 @@
+import React from "react";
+import { Alert } from "../Alert";
+
+export default {
+    title: "base/Alert",
+    component: Alert
+};
+
+const Template = args => <Alert {...args} />;
+
+export const greenAlert = Template.bind({});
+
+greenAlert.args = {
+    children: "test alert",
+    color: "green",
+    level: true
+};
--- a/src/js/base/stories/Alert.stories.js
+++ b/src/js/base/stories/Alert.stories.js
@@ -3,7 +3,14 @@ import { Alert } from "../Alert";
 
 export default {
     title: "base/Alert",
-    component: Alert
+    component: Alert,
+    argTypes: {
+        color: {
+            options: ["red", "orange", "purple", "blue", "black"],
+            control: { type: "radio" },
+            defaultValue: "red"
+        }
+    }
 };
 
 const Template = args => <Alert {...args} />;
@@ -11,7 +18,6 @@ const Template = args => <Alert {...args} />;
 export const greenAlert = Template.bind({});
 
 greenAlert.args = {
-    children: "test alert",
-    color: "green",
+    children: "This is an example Alert!",
     level: true
 };
--- a/src/js/base/stories/AttributionStories/Attribution.stories.js
+++ b/src/js/base/stories/AttributionStories/Attribution.stories.js
@@ -1,0 +1,20 @@
+import React from "react";
+import { Attribution } from "../../Attribution";
+
+export default {
+    title: "base/Attribution/Attribution",
+    component: Attribution,
+    args: {
+        time: "2021-12-17T03:24:00"
+    }
+};
+
+const Template = args => <Attribution {...args} />;
+
+export const NoUserAttribution = Template.bind({});
+
+export const TestUserAttribution = Template.bind({});
+
+TestUserAttribution.args = {
+    user: "test user"
+};
--- a/src/js/base/stories/AttributionStories/Attribution.stories.js
+++ b/src/js/base/stories/AttributionStories/Attribution.stories.js
@@ -11,10 +11,10 @@ export default {
 
 const Template = args => <Attribution {...args} />;
 
-export const NoUserAttribution = Template.bind({});
+export const attributionWithUsername = Template.bind({});
 
-export const TestUserAttribution = Template.bind({});
-
-TestUserAttribution.args = {
-    user: "test user"
+attributionWithUsername.args = {
+    user: "James Smith"
 };
+
+export const attributionWithNoUsername = Template.bind({});
--- a/src/js/base/stories/AttributionStories/NameAttribution.stories.js
+++ b/src/js/base/stories/AttributionStories/NameAttribution.stories.js
@@ -8,8 +8,8 @@ export default {
 
 const Template = args => <NameAttribution {...args} />;
 
-export const TestUserAttribution = Template.bind({});
+export const nameAttribution = Template.bind({});
 
-TestUserAttribution.args = {
-    user: "test user"
+nameAttribution.args = {
+    user: "James Smith"
 };
--- a/src/js/base/stories/AttributionStories/NameAttribution.stories.js
+++ b/src/js/base/stories/AttributionStories/NameAttribution.stories.js
@@ -1,0 +1,15 @@
+import React from "react";
+import { NameAttribution } from "../../Attribution";
+
+export default {
+    title: "base/Attribution/NameAttribution",
+    component: NameAttribution
+};
+
+const Template = args => <NameAttribution {...args} />;
+
+export const TestUserAttribution = Template.bind({});
+
+TestUserAttribution.args = {
+    user: "test user"
+};
--- a/src/js/base/stories/BoxStories/Box.stories.js
+++ b/src/js/base/stories/BoxStories/Box.stories.js
@@ -1,0 +1,20 @@
+import React from "react";
+import { Box } from "../../Box";
+
+export default {
+    title: "base/Box/Box",
+    component: Box,
+    args: {
+        children: "test"
+    }
+};
+
+const Template = args => <Box {...args} />;
+
+export const sampleBox = Template.bind({});
+
+export const boxWithOnClick = Template.bind({});
+
+boxWithOnClick.args = {
+    onClick: () => console.log("clicked")
+};
--- a/src/js/base/stories/BoxStories/Box.stories.js
+++ b/src/js/base/stories/BoxStories/Box.stories.js
@@ -1,11 +1,15 @@
 import React from "react";
-import { Box } from "../../Box";
+import { Box, BoxGroupHeader, BoxGroupSection } from "../../Box";
 
 export default {
     title: "base/Box/Box",
     component: Box,
-    args: {
-        children: "test"
+    parameters: {
+        docs: {
+            description: {
+                component: "A simple box wrapper element"
+            }
+        }
     }
 };
 
@@ -13,8 +17,26 @@ const Template = args => <Box {...args} />;
 
 export const sampleBox = Template.bind({});
 
-export const boxWithOnClick = Template.bind({});
+sampleBox.args = {
+    children: "This is a Box!"
+};
 
-boxWithOnClick.args = {
+export const clickableBox = Template.bind({});
+
+clickableBox.args = {
+    children: "This is a clickable Box!",
     onClick: () => console.log("clicked")
 };
+
+const BoxWithElementsTemplate = args => (
+    <Box {...args}>
+        <BoxGroupHeader {...args}>
+            <h2>This is a Box with a header and 3 elements!</h2>
+        </BoxGroupHeader>
+        <BoxGroupSection>Element 1</BoxGroupSection>
+        <BoxGroupSection>Element 2</BoxGroupSection>
+        <BoxGroupSection>Element 3</BoxGroupSection>
+    </Box>
+);
+
+export const exampleBox = BoxWithElementsTemplate.bind({});
--- a/src/js/base/stories/BoxStories/Box.stories.js
+++ b/src/js/base/stories/BoxStories/Box.stories.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { Box, BoxGroupHeader, BoxGroupSection } from "../../Box";
+import { Box } from "../../Box";
 
 export default {
     title: "base/Box/Box",
@@ -7,7 +7,8 @@ export default {
     parameters: {
         docs: {
             description: {
-                component: "A simple box wrapper element"
+                component:
+                    "A box wrapper element that is useful for building custom wrapping elements or when a simple box is needed."
             }
         }
     }
@@ -27,16 +28,3 @@ clickableBox.args = {
     children: "This is a clickable Box!",
     onClick: () => console.log("clicked")
 };
-
-const BoxWithElementsTemplate = args => (
-    <Box {...args}>
-        <BoxGroupHeader {...args}>
-            <h2>This is a Box with a header and 3 elements!</h2>
-        </BoxGroupHeader>
-        <BoxGroupSection>Element 1</BoxGroupSection>
-        <BoxGroupSection>Element 2</BoxGroupSection>
-        <BoxGroupSection>Element 3</BoxGroupSection>
-    </Box>
-);
-
-export const exampleBox = BoxWithElementsTemplate.bind({});
--- a/src/js/base/stories/BoxStories/BoxGroup.stories.js
+++ b/src/js/base/stories/BoxStories/BoxGroup.stories.js
@@ -1,15 +1,51 @@
 import React from "react";
-import { BoxGroup } from "../../Box";
+import { BoxGroup, BoxGroupHeader } from "../../Box";
+import { Input } from "../../Input";
+import { Button } from "../../Button";
+import styled from "styled-components";
+
+const StyledDiv = styled.div`
+    button {
+        margin: 5px;
+    }
+    input {
+        margin: 5px 0px 0px 5px;
+        width: 50%;
+    }
+`;
 
 export default {
     title: "base/Box/BoxGroup",
-    component: BoxGroup
+    component: BoxGroup,
+    parameters: {
+        docs: {
+            description: {
+                component: "A wrapper element designed to contain children"
+            }
+        }
+    }
 };
 
-const Template = args => <BoxGroup {...args} />;
+const Template = args => (
+    <StyledDiv>
+        <BoxGroup {...args}>
+            <BoxGroupHeader>
+                <h2>This is a BoxGroup</h2>
+            </BoxGroupHeader>
+            <Input type="text" placeholder="Enter a valid email here!" />
+            <Button type="submit" color="blue">
+                Submit
+            </Button>
+        </BoxGroup>
+    </StyledDiv>
+);
 
-export const testBoxGroup = Template.bind({});
+export const exampleBoxGroup = Template.bind({});
 
-testBoxGroup.args = {
-    children: "test"
+const PlainTemplate = args => <BoxGroup {...args} />;
+
+export const emptyBoxGroup = PlainTemplate.bind({});
+
+emptyBoxGroup.args = {
+    children: "Empty BoxGroup"
 };
--- a/src/js/base/stories/BoxStories/BoxGroup.stories.js
+++ b/src/js/base/stories/BoxStories/BoxGroup.stories.js
@@ -1,0 +1,15 @@
+import React from "react";
+import { BoxGroup } from "../../Box";
+
+export default {
+    title: "base/Box/BoxGroup",
+    component: BoxGroup
+};
+
+const Template = args => <BoxGroup {...args} />;
+
+export const testBoxGroup = Template.bind({});
+
+testBoxGroup.args = {
+    children: "test"
+};
--- a/src/js/base/stories/BoxStories/BoxGroup.stories.js
+++ b/src/js/base/stories/BoxStories/BoxGroup.stories.js
@@ -1,17 +1,11 @@
 import React from "react";
-import { BoxGroup, BoxGroupHeader } from "../../Box";
+import { BoxGroup, BoxGroupHeader, BoxGroupSection } from "../../Box";
 import { Input } from "../../Input";
 import { Button } from "../../Button";
 import styled from "styled-components";
 
-const StyledDiv = styled.div`
-    button {
-        margin: 5px;
-    }
-    input {
-        margin: 5px 0px 0px 5px;
-        width: 50%;
-    }
+const StyledButton = styled(Button)`
+    margin-top: 10px;
 `;
 
 export default {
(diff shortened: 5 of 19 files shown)


src/js/base/stories/Alert.stories.js
```diff
@@ -0,0 +1,17 @@
...
+    title: "base/Alert",
+    component: Alert
+};
+
+const Template = args => <Alert {...args} />;
+
+export const greenAlert = Template.bind({});
+
+greenAlert.args = {
+    children: "test alert",
```
ReeceHoffmann: Given that this is meant to be a showcase of how each component can/should look and be used, l would use a properly punctuated example. "ex: This is an example Alert!"

Also, I think the alert window typically has an alert icon in it. Guessing this is lost here due to font-awesome shenanigans, but ideally it would render in the preview.

src/js/base/stories/Alert.stories.js
```diff
@@ -0,0 +1,17 @@
...
+    color: "green",
```
ReeceHoffmann: As it stands I don't love the fact that we can only see one color, and that the color used is not what I would think of as typical for an Alert. Here I would suggest leverageing the `select` functionallity of the "controls" part of storybook so that there is a selectable list of suggest colors that the user can view. (https://storybook.js.org/docs/react/essentials/controls) That way it is simple to flip through and see what each alert will look like with minimal effort.

src/js/base/stories/AttributionStories/Attribution.stories.js
```diff
@@ -0,0 +1,20 @@
...
+    }
+};
+
+const Template = args => <Attribution {...args} />;
+
+export const NoUserAttribution = Template.bind({});
+
+export const TestUserAttribution = Template.bind({});
+
+TestUserAttribution.args = {
```
ReeceHoffmann: Three things: 

1. The auto generated title of "Test User Attribution" imo fails to make it obvious what is really being changed. 
2. I would replace "test user" with a proper placeholder name like "James Smith" or other realistic name.

src/js/base/stories/AttributionStories/NameAttribution.stories.js
```diff
@@ -0,0 +1,15 @@
...
+    title: "base/Attribution/NameAttribution",
+    component: NameAttribution
+};
+
+const Template = args => <NameAttribution {...args} />;
+
+export const TestUserAttribution = Template.bind({});
+
+TestUserAttribution.args = {
+    user: "test user"
```
ReeceHoffmann: Same as above on the name, better to use something realistic

src/js/base/stories/BoxStories/Box.stories.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { Box } from "../../Box";
+
+export default {
+    title: "base/Box/Box",
+    component: Box,
+    args: {
+        children: "test"
```
ReeceHoffmann: This needs better children. Having an example with just text is fine, but I would at least have something along the lines of "This is an example Box!". Seeing as this is fundamentally a wrapping element I would also include a story that shows the box with more complex contents.

src/js/base/stories/BoxStories/Box.stories.js
```diff
@@ -0,0 +1,20 @@
...
+};
+
+const Template = args => <Box {...args} />;
+
+export const sampleBox = Template.bind({});
+
+export const boxWithOnClick = Template.bind({});
+
+boxWithOnClick.args = {
+    onClick: () => console.log("clicked")
```
ReeceHoffmann, line 29: Again I would change the internal text to be something descriptive. Something like " This is a clickable Box!"

src/js/base/stories/BoxStories/BoxGroup.stories.js
```diff
@@ -0,0 +1,15 @@
...
+    title: "base/Box/BoxGroup",
+    component: BoxGroup
+};
+
+const Template = args => <BoxGroup {...args} />;
+
+export const testBoxGroup = Template.bind({});
+
+testBoxGroup.args = {
+    children: "test"
```
ReeceHoffmann: I don't think showing this group in the absence of children is the ideal way to show the purpose, usage, or styling of a box group. I would strongly suggest having the main example be one that includes meaningful children, with the text only version being removed or relegated to a separate story. Something like the `Email` class is a nice small example of the kind of thing `BoxGroup` is used for. No need to use `Email` in particular but it is a good example.

As with the others, if you want to keep the text only version, the text should also be changed to something descriptive: "Empty BoxGroup."

src/js/base/stories/BoxStories/BoxGroupHeader.stories.js
```diff
@@ -0,0 +1,16 @@
...
+export default {
+    title: "base/Box/BoxGroupHeader",
+    component: BoxGroupHeader
+};
+
+const Template = args => (
+    <BoxGroupHeader {...args}>
+        <h2>test</h2>
+        <p>test paragraph</p>
+    </BoxGroupHeader>
```
ReeceHoffmann: Two things:

1. This component should always be used in the context of a `BoxGroup` component and looks strange without it
2. Again, the text should be more descriptive of what the user is seeing rather than just saying "test"

src/js/base/stories/BoxStories/BoxGroupSection.stories.js
```diff
@@ -0,0 +1,16 @@
...
+};
+
+const Template = args => <BoxGroupSection {...args} />;
+
+export const testBoxGroupSection = Template.bind({});
+
+testBoxGroupSection.args = {
+    children: "test",
+    onClick: () => console.log("clicked")
+};
```
ReeceHoffmann: Like BoxGroup, this component should really be shown in the context of how it would normally be used. (ei: used to wrap element(s) that are inside of a box group) As it is, there is no clear indication of what the element does or what it is for.

Also I would move the `onclick` version to a separate story, as that is not always how the section is use

src/js/base/stories/BoxStories/BoxTitle.stories.js
```diff
@@ -0,0 +1,15 @@
...
+    title: "base/Box/BoxTitle",
+    component: BoxTitle
+};
+
+const Template = args => <BoxTitle {...args} />;
+
+export const testBoxTitle = Template.bind({});
+
+testBoxTitle.args = {
+    children: "Test Title"
```
ReeceHoffmann: Again I think this would make more sense to be shown being used as a title for something rather than just having the words float in space for the main example. (Could be the title for something as basic as a Box with some Lorem ipsum text inside) As with others I would also change the title to be more descriptive. Also it seems in the preview version that the font weight change is not coming through: 

image

src/js/base/stories/BoxStories/LinkBox.stories.js
```diff
@@ -0,0 +1,16 @@
...
+    title: "base/Box/LinkBox",
+    component: LinkBox
+};
+
+const Template = args => <LinkBox {...args} />;
+
+export const testLinkBox = Template.bind({});
+
+testLinkBox.args = {
+    children: "test link",
```
ReeceHoffmann: Change the text, otherwise I think this stands well enough on its own.

src/js/base/stories/BoxStories/SelectBoxGroupSection.stories.js
```diff
@@ -0,0 +1,17 @@
...
+    title: "base/Box/SelectBoxGroupSection",
+    component: SelectBoxGroupSection
+};
+
+const Template = args => <SelectBoxGroupSection {...args} />;
+
+export const testSelectBoxGroupSection = Template.bind({});
+
+testSelectBoxGroupSection.args = {
+    active: true,
```
ReeceHoffmann: This component should never be used outside the context of a wrapping box and looks strange without it. Wrap this in a box and set the active to false rather than true by default.

I would also suggest including an example with a checkbox to show off the checkbox styling.

src/js/base/stories/ButtonStories/Button.stories.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { Button } from "../Button";
+import { Button } from "../../Button";
 
 export default {
```
ReeceHoffmann, line 4: Button would/should always have internal children. Remove the current default and make the version with text default. Like with `Alert` I would make the color in the storybook selectable from the virtool base colors. Finally, update the text to something the make sense for a button.

src/js/base/stories/ButtonStories/LinkButton.stories.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { LinkButton } from "../Button";
+import { LinkButton } from "../../Button";
 
 export default {
```
ReeceHoffmann, line 4: Not sure why the text here is underlined, it isn't in the actual app.

image

src/js/base/stories/NoneFoundStories/NoneFound.stories.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { NoneFound } from "../NoneFound";
+import { NoneFound } from "../../NoneFound";
```
ReeceHoffmann, line 3: I don't think the "None" story here makes sense, as not passing a noun is a mistake and not an intended use case. If the goal is to show that the noun is inserted into the sentence then you could replace the story with a version that shows the replaced location by passing a string literal like "{plural noun}" into the `noun` parameter. (Making the final string "No {plural nouns} found")

src/js/base/stories/NoneFoundStories/NoneFoundBox.stories.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { NoneFoundBox } from "../NoneFound";
+import { NoneFoundBox } from "../../NoneFound";
```
ReeceHoffmann, line 2: Looks good, though in the preview the icon is missing. Would be nice to have that fixed if possible.

src/js/base/stories/NoneFoundStories/NoneFoundSection.stories.js
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import { NoneFoundSection } from "../../NoneFound";
+
+export default {
+    title: "base/NoneFound/NoneFoundSection",
+    component: NoneFoundSection,
+    args: {
+        noun: "testNoun"
```
ReeceHoffmann:
```suggestion
        noun: "items"
```
If you want to add a more general case to show where things get substituted I would do that in a story

src/js/base/stories/NoneFoundStories/NoneFoundSection.stories.js
```diff
@@ -0,0 +1,20 @@
...
+
+const Template = args => <NoneFoundSection {...args} />;
+
+export const testNoneFoundSection = Template.bind({});
+
+export const NoneFoundSectionWithChild = Template.bind({});
+
+NoneFoundSectionWithChild.args = {
+    children: "test"
+};
```
ReeceHoffmann: Default text needs improvement. Should be a realistic seeming best practices example.

src/js/base/stories/RelativeTime.stories.js
```diff
@@ -0,0 +1,29 @@
...
+import { RelativeTime } from "../RelativeTime";
+
+const currentDate = new Date();
+
+export default {
+    title: "base/RelativeTime",
+    component: RelativeTime
+};
+
+const Template = args => <RelativeTime {...args} />;
```
ReeceHoffmann, line 11: You might need to mock the function needed to fetch the current time. As written now the "Future relative time" example is only in the future for another 6 days in the preview. https://deploy-preview-184--admirable-axolotl-a29e53.netlify.app/?path=/docs/base-relativetime--current-relative-time

ReeceHoffmann (changes requested):
A few general comments:

1. The placeholder text should be updated throughout to be more professional. As appropriate I would tend towards descriptive text or realistic replacements. In the case where there is components contained within that should have text, but are not the main focus, consider usage of a standard default text like lorem ipsum. (ex: you are writing a story for a title component but want to have a paragraph to label. In this case give the title a descriptive name and the paragraph lorem ipsum or similar)

2. Consider the context that a component would be used, and for the main example show it being used in that manner. in The primary example should give a good representation of why and where the component would be used. (Some component like alert are fine on their own, but others like `BoxGroup` make little sense to show solo) In my opinion it is okay to show more than just the one component if it make it's usage more clear and or looks better so long as it is still clear what component is being displayed

3. I think many of these stories would be improved by a short description of what the underlying component is for/does. Even a short description like the autogenerated one for `RelativeTime` could go a long way for components like`BoxGroup`. Storybook also has some examples with more detailed documentation. [example](https://brightlayer-ui-components.github.io/react/?path=/info/components-app-bar--get-read-me-story)

src/js/base/stories/ButtonStories/LinkButton.stories.js
elanjm12: I looked into this briefly and I'm not sure why it's being underlined. I will look into it more

elanjm12: resolved

src/js/base/stories/RelativeTime.stories.js
elanjm12: Anytime the preview is refreshed the times are reset and display correctly. I tested this by changing the future time to 5 minutes in the future and resetting it by refreshing the page.

elanjm12: I tested this by changing the future relative time to 5 minutes in the future. Refreshing the page resets the current date. And therefore resets the future and past dates. However the 5 minutes in the future is never displayed again. It always defaults to 4 minutes in the future after the first refresh. I am going to test the behaviour with days but it seems like refreshing the page resolves this issue.

src/js/base/stories/BoxStories/Box.stories.js
```diff
@@ -0,0 +1,42 @@
...
+const BoxWithElementsTemplate = args => (
+    <Box {...args}>
+        <BoxGroupHeader {...args}>
+            <h2>This is a Box with a header and 3 elements!</h2>
+        </BoxGroupHeader>
+        <BoxGroupSection>Element 1</BoxGroupSection>
+        <BoxGroupSection>Element 2</BoxGroupSection>
+        <BoxGroupSection>Element 3</BoxGroupSection>
+    </Box>
+);
```
ReeceHoffmann: I think this example would make more sense to be a child of a`BoxGroup` rather than `Box` as the `Box` component applies margins that make the example look strange.

src/js/base/stories/ButtonStories/Button.stories.js
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import { Button } from "../../Button";
+
+export default {
+    title: "base/Button/Button",
+    component: Button,
+    args: {
+        children: "Click Here"
+    },
+    argTypes: {
```
ReeceHoffmann, line 15: The interactable button in this component doesn't seem to react to the `tipPlacement` param in the control panel

src/js/base/stories/ButtonStories/Button.stories.js
```diff
@@ -0,0 +1,25 @@
...
+        color: {
+            options: ["red", "orange", "purple", "blue", "black"],
```
ReeceHoffmann: I would be tempted to reorder first couple options by likelihood of usage. ie default blue then red then ... Alternatively alphabetically would be a good choice as well

ReeceHoffmann (changes requested):
Overall this is looking a lot better! Far fewer comments this round, though there are some things I would still generally like to see changed:

1. I noticed while playing with the preview that some of the toggleable options that we have don't seem to have an obvious effect. (ex. the `level` parameter of the `alert` component) If possible I think it would be great to either make it more clear what those options are doing, or disable them and/or add a description of what they are intended for. If the functionality is important and visual but not easy to show in the `showcase` example then it might be a good to display as a story.
2. I noticed for the interactable version of the stories that many of the toggle-able elements did not react when clicked. (`checkbox` and `selectBoxGroupSection`) Not sure how easy this will be, but I think it would be great if the elements could still be interacted with. That said, for this to be a good experience any changes made by clicking the element would be to be reflected in the control panel. (ex. When clicking a checkbox the active parameter in the control panel should also toggle so that the the actual parameters and those appearing in the control panel stay in sync)
3. For the component descriptions I would focus on a bit more on why the component are useful. (ie: Why should a developer use a BoxGroupSection vs a normal `div`. Visual consistency and provided formatting + specific built in function will probably be the main reasons here)

src/js/base/stories/RelativeTime.stories.js
ReeceHoffmann: Not sure if it is still a problem, but before the preview version would have the unintended time persist through refresh. Best I can tell the underlying component is updating the current time correctly but the future time is static in the preview.

elanjm12:
For comment 2: The onClick functionality was already built in for the selectBoxGroupSection with checkbox but wasn't used for the plain selectBoxGroupSection or the checkbox by itself. This functionality has been added for the other stories.

elanjm12:
Changes addressed

src/js/base/stories/BoxStories/BoxGroupHeader.stories.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+import { BoxGroup, BoxGroupHeader } from "../../Box";
+
+export default {
+    title: "base/Box/BoxGroupHeader",
+    component: BoxGroupHeader,
+    subcomponents: BoxGroup
+};
+
+const Template = args => (
```
ReeceHoffmann, line 20: This component and a `SelectBoxGroupSection`  have an extra toolbar of options that seem like they could be coming from styled components and seem helpful. Would be better to suppress these

image

src/js/base/stories/BoxStories/BoxTitle.stories.js
```diff
@@ -0,0 +1,24 @@
...
+
+const Template = args => (
+    <BoxGroup>
+        <BoxTitle {...args} />
+        <p>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et
+            dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex
+            ea commodo consequat.
+        </p>
+    </BoxGroup>
```
ReeceHoffmann:
```suggestion
    <Box>
        <BoxTitle {...args} />
        <p>
            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et
            dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex
            ea commodo consequat.
        </p>
    </Box>
```

This component still looks a bit funny. Pretty sure that is because `BoxTitle` meant to be used as a child of a `Box` rather than `BoxGroup`.

src/js/base/stories/BoxStories/BoxGroup.stories.js
```diff
@@ -0,0 +1,65 @@
...
+        </BoxGroup>
+    </StyledDiv>
+);
+
+export const exampleBoxGroup = Template.bind({});
+
+const BoxWithElementsTemplate = args => (
+    <BoxGroup {...args}>
+        <BoxGroupHeader {...args}>
+            <h2>This is a Box with a header and 3 elements!</h2>
```
ReeceHoffmann:
```suggestion
            <h2>This is a BoxGroup with a header and 3 elements!</h2>
```

src/js/base/stories/BoxStories/BoxGroup.stories.js
```diff
@@ -0,0 +1,65 @@
...
+
+export const exampleBox = BoxWithElementsTemplate.bind({});
+
+const PlainTemplate = args => <BoxGroup {...args} />;
+
+export const emptyBoxGroup = PlainTemplate.bind({});
+
+emptyBoxGroup.args = {
+    children: "Empty BoxGroup"
+};
```
ReeceHoffmann: I would just remove this example. I don't think it effectively shows off the styling of the `BoxGroup`.

src/js/base/stories/BoxStories/BoxGroup.stories.js
```diff
@@ -0,0 +1,65 @@
...
+const Template = args => (
+    <StyledDiv>
+        <BoxGroup {...args}>
+            <BoxGroupHeader>
+                <h2>This is a BoxGroup</h2>
+            </BoxGroupHeader>
+            <Input type="text" placeholder="Enter a valid email here!" />
+            <Button type="submit" color="blue">
+                Submit
+            </Button>
```
ReeceHoffmann: I think you could avoid using `StyledDiv` and get a fairly similar effect styling wise by instead wrapping your `input` and `button` components in a single `BoxGroupSection`. Might still have to apply some styling to move the button, but it moves the example closer to the intended use case.

src/js/base/stories/RelativeTime.stories.js
```diff
@@ -0,0 +1,29 @@
...
+
+futureRelativeTime.args = {
+    time: new Date(currentDate.getFullYear(), currentDate.getMonth(), currentDate.getDate() + 7)
+};
+
+export const pastRelativeTime = Template.bind({});
+
+pastRelativeTime.args = {
+    time: new Date(currentDate.getFullYear(), currentDate.getMonth(), currentDate.getDate() - 7)
+};
```
ReeceHoffmann: The datetimes look like they are still shifting to me: 

image


I had assumed that this would be linked with when the the last preview was published, but then I probably would have expected it to have shifted by 5 days here and not one. Not fixed by refreshing.

src/js/base/stories/BoxStories/SelectBoxGroupSection.stories.js
```diff
@@ -0,0 +1,49 @@
...
+testSelectBoxGroupSection.args = {
+    children: "This is a SelectBoxGroupSection!"
+};
+
+const Template2 = args => {
+    const [{ active }, updateArgs] = useArgs(false);
+    return (
+        <BoxGroup>
+            <SelectBoxGroupSection {...args} onClick={() => updateArgs({ active: !active })}>
+                <Checkbox checked={active} label="this is a SelectBoxGroupSection with a Checkbox!" />
```
ReeceHoffmann:
```suggestion
                <Checkbox checked={active} label="This is a SelectBoxGroupSection with a Checkbox!" />
```

ReeceHoffmann (changes requested):
Overall I think this is looking good. Still a few tweaks left though. 

As a general thing I noticed a number of components which have no obvious input are rendering a message like:  
image

If it makes sense for a component to have no easily changeable parameters I would personally prefer to still suppress this message, as to me it implies that the story is incomplete.

src/js/base/stories/RelativeTime.stories.js
elanjm12: This behaves slightly unexpectedly but not incorrectly. The logic how it was with +7 days and -7 days display the in 6 days and 8 days ago. I updated the logic so that it always displays in 7 days and 7 days ago. But refreshing does reset it on my end both locally and in preview.

elanjm12:
For the first comment, there does not seem to be a way to get rid of the no inputs found for this component message. This is just the standard message when a story does not have controls. I resolved this by ensuring that all stories have at least 1 input.

src/js/base/stories/RelativeTime.stories.js
ReeceHoffmann: Ahhh okay, I assumed that the 6/8 were from the time shift and not a result of the core logic

ReeceHoffmann (approved):
Looks good!

PR #185, opened 2022-10-24
chore: badge add storybook

elanjm12:
Add storybook for badge component


igboyes (changes requested):
"test badge" is not an appropriate example value for public documentation of these components.

Please look at some examples of React component libraries to absorb what our stories should look like.

* https://chakra-ui.com/docs/components/badge/usage
* https://semantic-ui.com/elements/label.html
* https://mui.com/material-ui/react-chip/

igboyes (approved):

PR #187, opened 2022-11-08
feat: add pagination base component

elanjm12:
Screenshot from 2022-11-08 10-19-27

### DIFF
--- a/src/js/base/Pagination.jsx
+++ b/src/js/base/Pagination.jsx
@@ -1,15 +1,15 @@
-import React, { useState } from "react";
+import React, { useMemo } from "react";
 import { Button } from "../base";
 import styled from "styled-components";
-import { range } from "lodash-es";
+import { range, min, max, map } from "lodash-es";
 
 const StyledPaginationBox = styled.div`
-    justify-content: center;
     display: flex;
     width: 500px;
+    justify-content: center;
     margin-left: 25%;
-    margin-right: 25%;
-    button:first-of-type {
+
+    button:first-child {
         width: 100px;
     }
     button:last-child {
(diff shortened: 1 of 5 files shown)


src/js/base/Pagination.jsx
```diff
@@ -0,0 +1,77 @@
...
+        return pageNumbers.slice(0, 4);
+    } else if (currentPage === pageNumbers.length) {
+        return pageNumbers.slice(pageNumbers.length - 4, pageNumbers.length);
+    } else if (currentPage === pageNumbers.length - 1) {
+        return pageNumbers.slice(currentPage - 3, currentPage + 1);
+    }
+    return pageNumbers.slice(currentPage - 2, currentPage + 2);
+};
+
+export const Pagination = ({ itemsPerPage, totalItems, color }) => {
```
igboyes: Drop `color` prop. Pagination components are not going to vary in appearance over the application UI.

src/js/base/Pagination.jsx
```diff
@@ -0,0 +1,77 @@
...
+    if (pageNumbers.length <= 4) {
+        return pageNumbers;
+    } else if (currentPage === 1) {
+        return pageNumbers.slice(0, 4);
+    } else if (currentPage === pageNumbers.length) {
+        return pageNumbers.slice(pageNumbers.length - 4, pageNumbers.length);
+    } else if (currentPage === pageNumbers.length - 1) {
+        return pageNumbers.slice(currentPage - 3, currentPage + 1);
+    }
+    return pageNumbers.slice(currentPage - 2, currentPage + 2);
```
igboyes: Good use of `min()`, `max()`, and `range()` can greatly simplify this function.

If you are returning in `if` blocks, don't bother with `else if`. This is functionally equivalent:
```js
    if (pageNumbers.length <= 4) {
        return pageNumbers;
    } 

    if (currentPage === 1) {
        return pageNumbers.slice(0, 4);
    }

    if (currentPage === pageNumbers.length) {
        return pageNumbers.slice(pageNumbers.length - 4, pageNumbers.length);
    }

    ...        
```

src/js/base/Pagination.jsx
```diff
@@ -0,0 +1,77 @@
...
+        width: 100px;
+    }
+
+    button {
+        margin: 5px;
+        height: 25px;
+    }
+`;
+
+export const sliceData = (items, currentPage, itemsPerPage) => {
```
igboyes: Are you sure this is used somewhere?

src/js/base/Pagination.jsx
```diff
@@ -0,0 +1,77 @@
...
+        return pageNumbers.slice(currentPage - 3, currentPage + 1);
+    }
+    return pageNumbers.slice(currentPage - 2, currentPage + 2);
+};
+
+export const Pagination = ({ itemsPerPage, totalItems, color }) => {
+    const [currentPage, setCurrentPage] = useState(1);
+    const paginate = number => setCurrentPage(number);
+
+    const pageNumbers = range(1, Math.ceil(totalItems / itemsPerPage + 1));
```
igboyes: Are you sure defining this array is the most efficient and clear way to implement this?

src/js/base/Pagination.jsx
```diff
@@ -0,0 +1,77 @@
...
+            key={pageNumber}
+            active={currentPage === pageNumber ? true : false}
+            onClick={() => paginate(pageNumber)}
+        >
+            {pageNumber}
+        </Button>
+    ));
+
+    return (
+        <StyledPaginationBox size={pageNumbers.length}>
```
igboyes: What is `size` used for?

igboyes (changes requested):
For all paging endpoints we get the following:
```
{
  "found_count": 1022,
  "page": 1,
  "page_count": 41,
  "per_page": 25,
  "total_count": 1026
}
```
Keep this in mind when working on this component.

Think about where values used in your component come from. Do they belong in component state or is their authoritative source somewhere else?

src/js/base/Pagination.jsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+export const Pagination = ({ currentPage, pageCount, onPageChange }) => {
+    const paginationRange = usePagination(pageCount, currentPage);
+
+    const pageButtons = map(paginationRange, pageNumber => (
+        <Button
+            color={currentPage === pageNumber ? "blue" : ""}
+            key={pageNumber}
+            active={currentPage === pageNumber ? true : false}
```
igboyes:
```suggestion
            active={currentPage === pageNumber}
```

src/js/base/Pagination.jsx
```diff
@@ -0,0 +1,72 @@
...
+            {pageNumber}
+        </Button>
+    ));
+
+    return (
+        <StyledPaginationBox>
+            <Button
+                onClick={() => onPageChange(currentPage - 1)}
+                color="blue"
+                disabled={currentPage === 1 ? true : false}
```
igboyes:
```suggestion
                disabled={currentPage === 1}
```

src/js/base/Pagination.jsx
```diff
@@ -0,0 +1,72 @@
...
+            >
+                {"< Previous"}
+            </Button>
+            {pageButtons}
+            <Button
+                onClick={() => onPageChange(currentPage + 1)}
+                color="blue"
+                disabled={currentPage === pageCount ? true : false}
+            >
+                {"Next >"}
```
igboyes:
```suggestion
                Next
```

igboyes (changes requested):
Fix the bool issues and I will merge.

When we use this component, I think we are going to be forced to change it a bit. The pagination should work via URLs (eg. `?page=2`), no these page numbers and buttons will have to be links instead of having `onClick` handlers.

At that point, I'd like all of the clickable elements, except for **Next** to be styled as basic links. Next will be the most commonly clicked and should have the most visual weight.

elanjm12:
bool issues addressed. I will address the first comment when I use this component in issue 1623.

igboyes (approved):

PR #188, opened 2022-11-11
ensure old download urls are not

igboyes:


ReeceHoffmann (approved):
Front-end stuff looks good to me. Though from a quick dig around I am not sure that `has_json` will ever be populated for the index. The Virtool core models for the index define a `has_files` variable, but not a `has_json` variable. So far as I can tell the presence of a json file is not indicated elsewhere in the response, so this seems to be a backend problem anyway

PR #190, opened 2022-11-28
feat: use pagination for uploaded files

elanjm12:
Currently there is no way to tell when removing a file that it is the correct type so it is necessary to assume that it is. Additionally, the documents come back in a weird order. Adding a new document sometimes adds it to a page other than the first. And documents are not always returned in order. The order does not change on refresh. There does not appear to be a reason that the documents are loaded in the order they are. These issues may be related to the backend.

Additionally, storybook does not appear to have a good way of handling the links present in this component. The story functions correctly and the current page can be changed by using the storybook controls. But clicking the buttons does not change the page. I think this story could be optimised in a future pr.

Screenshot from 2022-11-28 14-15-13

### DIFF
--- a/src/js/base/Pagination.jsx
+++ b/src/js/base/Pagination.jsx
@@ -1,8 +1,11 @@
-import React, { useMemo, useEffect } from "react";
+import React, { useEffect, useMemo } from "react";
 import styled from "styled-components";
-import { range, min, max, map } from "lodash-es";
+import { map, max, min, range } from "lodash-es";
 import { Link, useHistory, useLocation } from "react-router-dom";
 import { LinkButton } from "./Button";
+import { connect } from "react-redux";
+
+import { getColor, getFontSize } from "../app/theme";
 
 const StyledPaginationBox = styled.div`
     display: flex;
@@ -14,15 +17,15 @@ const StyledPaginationBox = styled.div`
 
 const StyledLink = styled(Link)`
     margin: 5px;
-    font-size: ${props => props.theme.fontSize.lg};
+    font-size: ${getFontSize("lg")};
     pointer-events: ${props => (props.disabled ? "none" : "")};
-    color: ${props => (props.active === "true" ? props.theme.color.blue : props.theme.color.blueDarkest)};
+    color: ${props => getColor({ color: props.$active ? "blue" : "blueDarkest", theme: props.theme })};
 `;
 
 const StyledNextLink = styled(LinkButton)`
     display: flex;
     margin: 5px;
-    width: 60px;params
+    width: 60px;
     justify-content: center;
     pointer-events: ${props => (props.disabled ? "none" : "")};
 `;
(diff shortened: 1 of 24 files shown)


elanjm12:
Ready for review

src/js/base/Pagination.jsx
```diff
@@ -36,33 +45,62 @@ const usePagination = (pageCount, currentPage) => {
...
+    const page = new URLSearchParams(location.search).get("page");
+
+    const history = useHistory();
+    useEffect(() => {
+        if (page > 1 && page > pageCount) {
+            history.push({ search: `?page=${page - 1}` });
+        } else {
+            history.push({ search: `?page=${page}` });
+        }
+    }, [pageCount]);
```
ReeceHoffmann: I think you could simplify this by using the min function. Something like:
```suggestion
    useEffect(() => {
        if (page > pageCount) {
            history.push({ search: `?page=${min(page, pageCount}` });
        } 
    }, [pageCount]);
```

src/js/base/Pagination.jsx
```diff
@@ -36,33 +45,62 @@ const usePagination = (pageCount, currentPage) => {
     return paginationRange;
 };
 
-export const Pagination = ({ currentPage, pageCount, onPageChange }) => {
+export const Pagination = ({ documents, renderRow, currentPage, pageCount, onLoadNextPage }) => {
     const paginationRange = usePagination(pageCount, currentPage);
+    const location = useLocation();
+    const page = new URLSearchParams(location.search).get("page");
```
ReeceHoffmann: If possible I would suggest making the pagination component agnostic of the source of method of setting the page numbers. On first blush I would suspect that the passed page number, onLoadNextPage, and page count should be able to handle most navigation

src/js/base/Pagination.jsx
```diff
@@ -36,33 +45,62 @@ const usePagination = (pageCount, currentPage) => {
...
+        }
+    }, [pageCount]);
+
+    const entries = map(documents, (_, index) => renderRow(index));
+
+    useEffect(() => {
+        if (page !== currentPage) {
+            onLoadNextPage(page);
+        }
+    }, [page]);
```
ReeceHoffmann: I feel like this should be impossible to be true... when/how does this come up?

src/js/files/api.js
```diff
@@ -19,10 +19,12 @@ export const find = ({ fileType }) =>
  * @param page {number} the page of results to get
  * @returns {promise}
  */
-export const list = ({ fileType }) =>
+export const list = ({ fileType, page }) =>
     Request.get("/api/uploads").query({
         type: fileType,
```
ReeceHoffmann:
```suggestion
        upload_type: fileType,
```
Not your change, but api expects `upload_type` rather than `type`. Filtering for the correct type should work again with this change

src/js/files/selectors.js
```diff
@@ -1,19 +1,13 @@
-import { keyBy, map, reject, sortBy } from "lodash-es";
+import { keyBy, map } from "lodash-es";
 import { createSelector } from "reselect";
 
-const getFiles = state => state.files.documents;
+const getFiles = state => state.files.items;
 
 export const getFilesById = createSelector(getFiles, files => keyBy(files, "id"));
 
 export const getFilteredFileIds = createSelector(getFiles, documents => {
```
ReeceHoffmann:
```suggestion
export const getFileIds = createSelector(getFiles, documents => {
```

src/js/files/components/Manager.jsx
```diff
@@ -1,115 +1,132 @@
...
-        this.props.onLoadNextPage(this.props.fileType, this.props.term, 1);
+export const FileManager = ({
+    onLoadNextPage,
+    canUpload,
+    onDrop,
+    documents,
+    validationRegex,
+    message,
+    total_count,
+    page,
```
ReeceHoffmann, line 18: Not sure it makes sense to have two sources to define the current page. I would have the url query by the single source of truth on the current page and remove page from redux.

ReeceHoffmann (dismissed):
Overall looking good, there are some logic tweaks I would suggest making if you have time. I can always take it across the finish line if you can't finish today.

igboyes (approved):

PR #195, opened 2022-12-13
chore: convert checkboxes use radix

elanjm12:
Convert checkbox to Radix. Maintain the same styling as before. Checkbox now accessed in tests through 
getByRole("checkbox", {name: "labelName"}).  toBeChecked property now works for testing if a checkbox is checked or not.

Screenshot from 2022-12-13 10-33-28


igboyes (changes requested):
Please add a utility `attachResizeObserver` instead of repeating the class definition and assignment to `window.ResizeObserver`.

igboyes (approved):

PR #196, opened 2022-12-20
chore: remove enzyme update hmm tests

elanjm12:


src/js/hmm/Classes.ts
```diff
@@ -0,0 +1,133 @@
...
+    accession: string;
+    gi: string;
+    name: string;
+    organism: string;
+};
+
+type families = {
+    family1: number;
+    family2: number;
+};
```
igboyes:
```suggestion
type Families = {
    family1: number;
    family2: number;
};
```

src/tests/setupTests.jsx
```diff
@@ -64,11 +55,21 @@ const renderWithRouter = (ui, state, history, createReducer) => {
...
 
+class ResizeObserver {
+    observe() {}
+    unobserve() {}
+}
+
+export const attachResizeObserver = () => {
+    window.ResizeObserver = ResizeObserver;
+};
+
```
igboyes, line 66: Didn't you already do this in the checkbox PR?

igboyes (changes requested):

src/tests/setupTests.jsx
elanjm12: This is the same code. I believe that all I did is move it below renderWithRouter.

igboyes: Sorry. That was a stupid question.

igboyes (approved):

PR #230, opened 2023-06-19
Use react-query for file listing

ReeceHoffmann:
Changes: 
 - Switches file manager views to use react-query
 - Switches Sample and subtraction creation view file listing to react-query
 - Adds websocket functionality for resources fetched via react-query
 - Add standardised factory functions for generating query keys
 - Fixed automatic test data generation to permit falsy custom values
 - Remove redux code relevant to listing files


src/js/app/websocket.ts
```diff
@@ -13,6 +12,24 @@ import { wsInsertTask, wsUpdateTask } from "../tasks/actions";
...
+    account: accountKeys,
+    roles: roleKeys,
+    users: userKeys,
+    uploads: fileKeys,
+};
+
+const reactQueryHandler = (queryClient: QueryClient) => (iface, operation) => {
+    const keyFactory = keyFactories[iface];
+    if (keyFactory === undefined) return;
+    queryClient.invalidateQueries(keyFactory.all);
```
igboyes:
```suggestion
    const keyFactory = keyFactories[iface];
    if (keyFactory) {
        return queryClient.invalidateQueries(keyFactory.all);
    }

```

src/js/administration/querys.ts
```diff
@@ -2,12 +2,24 @@ import { useMutation, useQuery, useQueryClient } from "react-query";
...
 export const useGetAdministratorRoles = () => {
     return useQuery("roles", getAdministratorRoles);
 };
 
+export const userKeys = {
+    all: ["users"] as const,
+    lists: () => [...userKeys.all, "list"] as const,
+    list: (filters: Array<string | number | boolean>) => [...userKeys.lists(), ...filters] as const,
+    details: () => [...userKeys.all, "details"] as const,
+    detail: (user_id: string) => [...userKeys.details(), user_id] as const,
```
igboyes: It is difficult to understand what is going on here at a glance.

I'm not sure of the benefit of these `somethingKeys` objects. It seems that each property is only read in a single hook (eg. `useGetUsers`) and that the hook is the best place to define the keys once.

Thoughts?

igboyes (approved):

PR #235, opened 2023-07-17
Display HMMs when installation is complete

ReeceHoffmann:
Fixes two bugs which prevented HMM install progress from being updated. 
 1. On install starts the UI to fetches HMM status, allowing the UI to associate the HMM install with the correct task and render the progress bar
 2. When the associated task is marked as complete the client makes a request for the first page of HMMs which in turn updates the status by default


src/js/hmm/api.js
```diff
@@ -1,9 +1,46 @@
...
 
-export default {
-    find: ({ term, page }) => Request.get("/hmms").query({ find: term, page }),
-    list: ({ page }) => Request.get("/hmms").query({ page }),
-    install: () => Request.post("/hmms/status/updates"),
-    get: ({ hmmId }) => Request.get(`/hmms/${hmmId}`),
-    purge: () => Request.delete("/hmms"),
-};
+/**
+ * Sagas compatible function for fetching a list of HMMs from the API.
```
igboyes: I think 'sagas compatible' is misleading. These functions return promises. That's it.

Putting this phrase at the beginning of the string also makes the developer scan to the end to see what the function actually does.

```suggestion
 * Fetch a page of HMM search results from the API.
```

src/js/hmm/api.js
```diff
@@ -1,9 +1,46 @@
...
+
+/**
+ * Sagas compatible function for fetching a deleting all HMMs.
+ * Note: the backend currently does not support purging hmms
+ *
+ * @returns {promise} resolves to an object containing the response.
+ */
+export function purge() {
+    return Request.delete("/hmms");
+}
```
igboyes: We should fix this in this PR.

igboyes (changes requested):

igboyes (changes requested):
* Remove unnecessary imports in saga module.
* Adopt TSDoc spec.
* Fix component naming.

igboyes (dismissed):

igboyes (approved):

PR #239, opened 2023-07-26
ci: run prettier and list non-compliances

igboyes:
* During CI, run a check to make sure all modules are compliant with Prettier.
* Fix all files that are currently non-compliant.


ReeceHoffmann:
@igboyes This is back up to date and ready to merge. DS is (rightfully in some cases) complaining about a large amount of code, but fixing it seems out of scope for this PR.

ReeceHoffmann (approved):

PR #241, opened 2023-08-16
Indicate job progress in subtractions list view

ReeceHoffmann:
Changes:
  - Adds progress circle base component for indicating job progress
  - Adds real time updating indication of job progress to subtractions list view

~~In view preview: (See storybook for interactive version of the progress circle)~~ Out of date see comment below for updated screenshot
image


igboyes (commented):
I haven't reviewed the code. Please change or comment on the following:

* **Timed out** shouldn't have a dot in the middle. The dot was meant to indicate that the job was active. I think the design is great otherwise.
* Remove the **Ready** word. I think we already discussed.
* For **Waiting**, I think a complete grey ring with a pulsing dot would by better. This shows that the job is active but there is no progress.
* For **Running**, why is there a halo around the progress circle.

ReeceHoffmann:
Updated styling:
Screenshot from 2023-08-23 15-57-18

Summary of changes:

- The track of the circle now updates to match the colour of the stage (grey for waiting, blue for running, red for failed, green for complete)
- The outer ring is removed
- The inner circle only renders when the job state is either `running` or `preparing`
- Ready is never rendered in subtractions list

src/js/base/ProgressCircle.tsx
```diff
@@ -0,0 +1,223 @@
...
+import { getColor, sizes } from "../app/theme";
+import { JobState } from "../jobs/types";
+
+/**
+ * Calculate the stroke width based on the total size of the progress circle
+ *
+ * @param size - The total size of the progress circle in pixels
+ * @returns  The stroke width in pixels
+ */
+
```
igboyes, line 13: Should be `calculateStrokeWidth` since `get` is pretty much reserved from getting values out of an object.

If you are going to pass a props object right to the function, the `get` naming is appropriate. I think I'd make all these functions take `props`. It will clean up your SC styling as well.

```suggestion
 */
```

src/js/base/ProgressCircle.tsx
```diff
@@ -0,0 +1,223 @@
...
+/**
+ * The indeterminate loader and track for the determinate progress circle.
+ *
+ * @param size - The total size of the progress circle in pixels
+ * @param state - The current state of the job or task
+ * @returns The progress circle track
+ */
+const ProgressCircleTrack = styled(ProgressCircleBase)<ProgressCircleTrackProps>`
+    stroke-dashoffset: ${props =>
+        props.state === "waiting" ? `${getStrokeDashoffset({ progress: 0.75, size: props.size })}px` : "0px"};
```
igboyes: Could this be pushed into `getStrokeDashOffset`?

src/js/base/ProgressCircle.tsx
```diff
@@ -0,0 +1,223 @@
...
+}
+
+/**
+ * Calculate stroke offset needed to display the correct amount of progress
+ *
+ * @param size - The total size of the progress circle in pixels
+ * @param progress - the amount of progress from 0 to 1
+ * @returns the stroke offset in pixels
+ */
+function getStrokeDashoffset({ progress, size }: { progress: number; size: number }): number {
```
igboyes:
```suggestion
function getStrokeDashOffset({ progress, size }: { progress: number; size: number }): number {
```

src/js/base/ProgressCircle.tsx
```diff
@@ -0,0 +1,223 @@
...
+    r: ${props => getRadius(props.size) / 3}px;
+    fill: ${props => getColor({ theme: props.theme, color: "blue" })};
+    opacity: ${props => (props.state == "waiting" ? 0 : 1)};
+    animation: ${fade} 1.25s ease-in-out infinite alternate;
+    vector-effect: non-scaling-stroke;
+    transform-box: fill-box;
+    transform-origin: center;
+`;
+
+const ProgressCircleSizes = {
```
igboyes:
```suggestion
const progressCircleSizes = {
```

src/js/base/ProgressCircle.tsx
```diff
@@ -0,0 +1,223 @@
...
+    xxl: 60,
+};
+
+/**
+ * Determine the appropriate color for the ProgressCircle based on the state of the job or task.
+ *
+ * @param state - The current state of the job or task
+ * @returns The color of the progress circle
+ */
+function getProgressColour(state: JobState): string {
```
igboyes: Don't use `get`.
```suggestion
function determineProgressColour(state: JobState): string {
```

src/js/utils/reducers.js
```diff
@@ -68,3 +68,25 @@ export const remove = (state, payload) => {
...
+
+    forEach(state.documents, document => {
+        if (document.job.id === payload.id) {
+            document.job = {
+                ...document.job,
+                ...payload,
+            };
+        }
+    });
+}
```
igboyes: I am guessing redux-toolkit is making this non-mutating. In that case do you even need to return `state`.

I think maybe this function already existed and you didn't write it...

src/js/base/ProgressCircle.tsx
```diff
@@ -0,0 +1,223 @@
...
+    return size / 5;
+}
+
+/**
+ * Calculate the radius of the circle based on the total size and stroke width
+ *
+ * @param size - The total size of the progress circle in pixels
+ * @returns The stroke width in pixels
+ */
+function getRadius(size: number): number {
```
igboyes: This might be a good idea as it could get confused with theme helpers for getting the border radius.
```suggestion
function getProgressRadius(size: number): number {
```

igboyes (changes requested):
This was a pleasure to read through.

src/js/base/ProgressCircle.tsx
ReeceHoffmann: The state check? I would prefer not to since the progress indicator uses the same function to calculate `stroke-dashoffset` so it can display the jobs progress. I could do it, but would have to account for it in the `ProgressIndicator`.  I suspect it would overall make the component logic less clear.

src/js/utils/reducers.js
ReeceHoffmann: Uhhh yeah it does, but my code is different from what I intended in another way. I wanted the API to be consistent with how other document updaters worked. Redux-toolkit hid my mistake, but I am planning to rework to always return state with next push

src/js/base/ProgressCircle.tsx
igboyes: I agree.

PR #247, opened 2023-09-08
fix: thicker bottom border in labels list view

raastinn:


src/js/labels/components/Item.tsx
```diff
@@ -8,6 +8,7 @@ import { RemoveLabel } from "./Remove";
 const LabelItemBox = styled(BoxGroupSection)`
     align-items: center;
     display: flex;
+    bottom: -1px;
```
ReeceHoffmann: Bottom border visually looks good, but the top now has an extra px of space. See if you can find the root cause of the thicker border. (double border, thicker border css, ect) Fixing that will give you a more durable change in the case that other styling changes

raastinn: Yes, it is a double border. I'm trying to figure out what is causing it and how to get rid of it

igboyes: I'd suggest looking at other places the base components are used and see how this glitch is avoided.

If I am understand the situation properly, one possible solution is using `:last` to set the bottom border to `none` on the last child.

igboyes (approved):

PR #249, opened 2023-09-08
fix: ensure terminated jobs have the correct badge colour

raastinn:


ReeceHoffmann:
Change looks good to me. Run prettier to reformat the file on the changed file this should be good to merge!

(ps: you can setup pycharm to run prettier whenever you save, super handy and saves time too)

ReeceHoffmann (approved):

PR #250, opened 2023-09-11
chore: add label models with docstrings

raastinn:


src/js/labels/type.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @prop description - the description of the label
+ * @prop id - the unique identifier of the label
+ * @prop name - the name of the label
+ */
+export type LabelNested = {
+    color: string;
+    description: string;
+    id: number;
+    name: string;
+};
```
ReeceHoffmann:
```suggestion

/*Smaller label object for representing labels in a list */
export type LabelNested = {
    /* A hex encoded color*/
    color: string;
    /*A detailed description*/
    description: string;
    <ect>
    id: number;
    <ect>
    name: string;
};
```
Few general tweaks:
  1. Descriptions of each field in a type should be right above the field
  2. The overall description should just be a one liner describing the overall object
  3. Update the Type descriptions to include more helpful info. Many of the current docstrings don't add much information that couldn't be grabbed at a glance.

Please rework the descriptions and fill out remaining fields not changed here

src/js/labels/type.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Label
+ * @extends LabelNested
+ * @prop count - the count associated with the label
+ */
+export type Label = LabelNested & {
+    count: number;
+};
```
ReeceHoffmann: Same comments as above

ReeceHoffmann (changes requested):
Good call adding docstrings to the types. We should be adding them as we create types. I did some research and have added comments suggesting the best practice I could find for documenting types/interfaces

src/js/labels/type.ts
```diff
@@ -0,0 +1,17 @@
...
+    description: string;
+    /* A unique identifier */
+    id: number;
+    /* A name for the label */
+    name: string;
+};
+
+/* A label object that extends LabelNested */
+export type Label = LabelNested & {
+    /* The count associated with the label */
```
ReeceHoffmann: What is being counted?

src/js/labels/type.ts
```diff
@@ -0,0 +1,17 @@
+/* Smaller label object for representing labels in a list */
```
ReeceHoffmann:
```suggestion
/*A partial label including essentials for rendering a list of labels*/
```
My bad. In general it is redundant to list types in the description

src/js/labels/type.ts
```diff
@@ -0,0 +1,17 @@
...
+export type LabelNested = {
+    /* A hex encoded color */
```
ReeceHoffmann:
```suggestion
    /* The hex encoded color */
```
On re-read I don't like my usage of "A" here.  Should be "The" as "A" leaves more ambiguity suggesting there could be more than one property doing the same thing.

Applies for all fields below as well.

src/js/labels/type.ts
```diff
@@ -0,0 +1,17 @@
...
+    color: string;
+    /* A detailed description */
+    description: string;
+    /* A unique identifier */
+    id: number;
+    /* A name for the label */
```
ReeceHoffmann: No need to specify relation to the label in the docstring. Doing so is redundant since the relation is inferred due to the `name` being a field of the model

src/js/labels/type.ts
```diff
@@ -0,0 +1,17 @@
...
+    name: string;
+};
+
+/* A label object that extends LabelNested */
```
ReeceHoffmann:
```suggestion
/* A sample label */
```
`Label` is the default case so there isn't much to say here.

ReeceHoffmann (changes requested):
A few tweaks to the docstrings still. Couple of mistakes in my examples and some tweaks still for the other ones. Overall this is looking really close to what we want.

src/js/labels/type.ts
```diff
@@ -0,0 +1,16 @@
+/* A partial label including essentials for rendering a list of labels */
+export type LabelNested = {
+    /* The hex encoded color */
+    color: string;
+    /* The detailed description */
+    description: string;
+    /* The unique identifier */
+    id: number;
+    name: string;
```
ReeceHoffmann, line 10: All fields should have a docstring. What about something like `/** The display name */`?

src/js/hmm/components/InstallOption.jsx
```diff
@@ -35,6 +35,7 @@ export const mapStateToProps = state => ({
 
 export const mapDispatchToProps = dispatch => ({
     onInstall: releaseId => {
+        // alert(installHMMs(releaseId));
```
ReeceHoffmann: ?

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #251, opened 2023-09-11
chore: refactor user type as intersection of usernested

raastinn:


src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
 import { AdministratorRoles } from "../administration/types";
 import { GroupMinimal, Permissions } from "../groups/types";
 
+/* Business to consumer specific details */
 type UserB2C = {
+    /* The display name (optional) */
     display_name?: string;
+    /* The family name (optional) */
     family_name?: string;
+    /* The given name (optional) */
```
ReeceHoffmann:
```suggestion
    /* The given name  */
```
No need to add optional, should be obvious at a glance from the type itself. Same is true for other instances as well

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
 
-export type User = {
-    id: string;
+/* A user with the essential information */
+export type UserNested = {
+    /* Indicates if the user is an administrator */
     administrator: boolean;
+    /* The unique identifier */
+    id: string;
+    /* The user's handle or username */
```
ReeceHoffmann, line 22:
```suggestion
    /* The display name */
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     handle: string;
+};
+
+/* A sample user */
```
ReeceHoffmann:
```suggestion
/* A virtool user */
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
+export type User = UserNested & {
+    /* Indicates if user is active */
```
ReeceHoffmann, line 28:
```suggestion
    /* Whether the account is enabled */
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     b2c?: UserB2C;
+    /* The user's display name (optional) */
     b2c_display_name?: string;
+    /* The user's family name (optional) */
     b2c_family_name?: string;
+    /* The user's given name (optional) */
     b2c_given_name?: string;
+    /* The user's object ID (optional) */
     b2c_oid?: string;
+    /* Indicates if the user is required to reset password */
```
ReeceHoffmann:
```suggestion
    /* Whether the user will be forced to reset their password on next login*/
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     force_reset: boolean;
+    /*  */
```
ReeceHoffmann: ?

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     groups: Array<GroupMinimal>;
+    /* The date of the user's last password change */
     last_password_change: Date;
+    /* The user's permissions */
     permissions: Permissions;
+    /* The user's primary group */
     primary_group: GroupMinimal;
+    /* THe role of the administrator user */
```
ReeceHoffmann:
```suggestion
    /* Their administrator role defining what resources they can modify*/
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     administrator_role: AdministratorRoles;
 };
 
+/* A response containing a list of users */
 export type UserResponse = {
+    /* The array of user objects */
```
ReeceHoffmann:
```suggestion
    /* The page of users*/
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     last_password_change: Date;
+    /* The user's permissions */
     permissions: Permissions;
+    /* The user's primary group */
     primary_group: GroupMinimal;
+    /* THe role of the administrator user */
     administrator_role: AdministratorRoles;
 };
 
+/* A response containing a list of users */
```
ReeceHoffmann:
```suggestion
/* User search results from the API */
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     b2c_family_name?: string;
+    /* The user's given name (optional) */
     b2c_given_name?: string;
+    /* The user's object ID (optional) */
     b2c_oid?: string;
+    /* Indicates if the user is required to reset password */
     force_reset: boolean;
+    /*  */
     groups: Array<GroupMinimal>;
+    /* The date of the user's last password change */
```
ReeceHoffmann:
```suggestion
    /* The date of their last password change */
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     active: boolean;
+    /* The B2C user information (optional) */
     b2c?: UserB2C;
+    /* The user's display name (optional) */
     b2c_display_name?: string;
+    /* The user's family name (optional) */
     b2c_family_name?: string;
+    /* The user's given name (optional) */
     b2c_given_name?: string;
+    /* The user's object ID (optional) */
```
ReeceHoffmann:
```suggestion
    /* Their B2C object ID*/
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
+    id: string;
+    /* The user's handle or username */
     handle: string;
+};
+
+/* A sample user */
+export type User = UserNested & {
+    /* Indicates if user is active */
     active: boolean;
+    /* The B2C user information (optional) */
```
ReeceHoffmann:
```suggestion
    /* Their B2C specific information*/
```

src/js/users/types.ts
```diff
@@ -1,42 +1,68 @@
...
     b2c?: UserB2C;
+    /* The user's display name (optional) */
```
ReeceHoffmann:
```suggestion
    /* Their display name (optional) */
```

... ect for other fields

ReeceHoffmann (changes requested):

src/js/users/types.ts
igboyes: > 

It should be left as is. 'Display name' is not used anywhere else as a variable name or terminology.

src/js/users/types.ts
igboyes: Terminology is right here. Users are activated or inactived. Not enabled or disabled.

ReeceHoffmann: `whether the user account can be logged into`? `active` alone could be interpreted as `currently online`

ReeceHoffmann (approved):

PR #256, opened 2023-09-15
feature: allow all ambiguous nucleotides

raastinn:


ReeceHoffmann (changes requested):
Change looks good, tests need updating to:

1. ensure the new chars continue to be accepted
2. Expect the chars to exist

Right now testings is failing b/c it only expects the 4 basic nucleotides and unknown

ReeceHoffmann (changes requested):
Super close just a couple tweaks to the testing:

1. make sure all the test names list the correct nucleotide. Some are only mention ATCGN when they should include the extended name
2. Add the new characters the successful submission test, as that is now part of the reqs that those chars are allowed

PR #257, opened 2023-09-18
chore: add job models

raastinn:


ReeceHoffmann (changes requested):
Typing looks good, the docstrings could use another quick run through. Things to look for:

1. Docstrings that don't add new information, or improve clarity, should be removed or modified to include more information. 
2. Docstrings with awkward phrasing or missing contents

ReeceHoffmann (approved):

PR #259, opened 2023-09-19
chore: modernise the subtractions list view

raastinn:


src/js/subtraction/components/List.jsx
```diff
@@ -1,58 +1,62 @@
...
 import { Badge, LegacyScrollList, LoadingPlaceholder, NoneFoundBox, ViewHeader, ViewHeaderTitle } from "../../base";
 import { findSubtractions } from "../actions";
 import SubtractionItem from "./Item";
 import SubtractionToolbar from "./Toolbar";
 
-export class SubtractionList extends React.Component {
-    componentDidMount() {
-        if (!this.props.fetched) {
-            this.props.onLoadNextPage(this.props.term, 1);
+export const SubtractionList = ({ onLoadNextPage, documents, page, page_count, total_count, fetched, term }) => {
```
ReeceHoffmann: Couple changes:

1. Needs typing
~~2. Should use ` function` notation. Arrow function notation is common throughout the codebase, but we are moving away from it except for anon functions~~
On second thought typing additions can come with the react-query conversion

src/js/subtraction/components/List.jsx
```diff
@@ -1,58 +1,62 @@
...
+    useEffect(() => {
+        if (!fetched) {
+            onLoadNextPage(term, 1);
         }
-    }
+    }, []);
 
-    renderRow = index => <SubtractionItem key={index} index={index} />;
+    const renderRow = index => <SubtractionItem key={index} index={index} />;
```
ReeceHoffmann: This is best defined outside the component to prevent redefinition on every render

src/js/subtraction/components/List.jsx
```diff
@@ -1,58 +1,62 @@
...
+                documents={documents}
+                onLoadNextPage={page => onLoadNextPage(term, page)}
+                page={page}
+                pageCount={page_count}
+                renderRow={renderRow}
+            />
+        );
+    } else {
+        subtractionComponents = <NoneFoundBox noun="subtractions" />;
+    }
```
ReeceHoffmann: Nitpick: use a ternary operator for situations like this. Makes it harder to have undefined vars by accident

src/js/subtraction/components/List.jsx
```diff
@@ -1,58 +1,62 @@
...
+};
 
 const mapStateToProps = state => ({
     ...state.subtraction,
+    fetched: state.subtraction.fetched,
+    documents: state.subtraction.documents,
+    page: state.subtraction.page,
+    page_count: state.subtraction.page_count,
+    total_count: state.subtraction.total_count,
+    term: state.subtraction.term,
```
ReeceHoffmann:
```suggestion
    fetched: state.subtraction.fetched,
    documents: state.subtraction.documents,
    page: state.subtraction.page,
    page_count: state.subtraction.page_count,
    total_count: state.subtraction.total_count,
    term: state.subtraction.term,
```
Generally a fan of reducing the usage of spread operators. That said you might be able to reduce the verbosity of this by taking advantage of object destructuring.

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #260, opened 2023-09-19
chore: add subtraction models

raastinn:


src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
+import { JobMinimal } from "../jobs/types";
+import { SampleNested } from "../samples/types";
+import { UserNested } from "../users/types";
+import { SearchResult } from "../utils/types";
+
+/* The composition of a Nucleotide */
```
ReeceHoffmann: Biology thing. The sample reads are made up a sequence of nucleotides (ATCG). This field is measuring, in percentage, the relative amount of each nucleotide. (And N for unknown)

src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
...
+    n: number;
+};
+
+/* A Subtraction file */
+export type SubtractionFile = {
+    download_url: string;
+    id: number;
+    name: string;
+    size: number;
+    subtraction: string;
```
ReeceHoffmann, line 24:
```suggestion
    subtraction: string;
```
I think this is probably a subtraction id but probably worth double checking and adding a docstring to the effect

src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
...
+    type: string;
+};
+
+/* A Subtraction with essential information  */
+export type SubtractionNested = {
+    /* The unique identifier */
+    id: string;
+    /* The subtraction's name */
```
ReeceHoffmann: Nit: I think this docstring can be dropped

src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
...
+export type SubtractionUpload = {
+    id: number | string;
+    name: string;
+};
+
+/* Minimal Subtraction model used for websocket messages and resource listings */
+export type SubtractionMinimal = SubtractionNested & {
+    count?: number;
+    created_at: Date;
+    /* Name of associated subtraction file */
```
ReeceHoffmann: Misleading docstring imo as it is an object providing both the human readable name and the unique identifier

src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
...
+    file: SubtractionUpload;
+    /* Information related to job of subtraction upload */
+    job?: JobMinimal;
+    nickname: string;
+    ready: boolean;
+    /* The associated user */
```
ReeceHoffmann: Remove or flesh out. In what way is the user associated? (ie: what is this users relationship with the subtraction)

src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
...
+    name: string;
+};
+
+/* Minimal Subtraction model used for websocket messages and resource listings */
+export type SubtractionMinimal = SubtractionNested & {
+    count?: number;
+    created_at: Date;
+    /* Name of associated subtraction file */
+    file: SubtractionUpload;
+    /* Information related to job of subtraction upload */
```
ReeceHoffmann: Awkward phrasing

src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
...
+    job?: JobMinimal;
+    nickname: string;
+    ready: boolean;
+    /* The associated user */
+    user?: UserNested;
+};
+
+/* Complete Subtraction model */
```
ReeceHoffmann: Nit: drop model. Should be obvious from the code and works a little better semantically imo since when you type something as a subtraction your saying: "this is a subtraction" vs "this is a subtraction model"

```suggestion
/* A complete subtraction */
```

src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
...
+    name: string;
+};
+
+/* Subtraction file being uploaded */
+export type SubtractionUpload = {
+    id: number | string;
+    name: string;
+};
+
+/* Minimal Subtraction model used for websocket messages and resource listings */
```
ReeceHoffmann: Nit: drop `model`
```suggestion
/* Minimal Subtraction used for websocket messages and resource listings */
```

src/js/subtraction/types.ts
```diff
@@ -0,0 +1,66 @@
...
+    t: number;
+    n: number;
+};
+
+/* A Subtraction file */
+export type SubtractionFile = {
+    download_url: string;
+    id: number;
+    name: string;
+    size: number;
```
ReeceHoffmann, line 22: Worth adding a doctstring with units, probably bytes

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #261, opened 2023-09-19
fix: rename label type.ts to types.ts

raastinn:


ReeceHoffmann:
Git is showing this as just adding a new file. Might just be a visual bug but I would probably just do this on a clean branch or include it with another bigger change

PR #262, opened 2023-09-22
chore: createStore function deprecated

raastinn:


src/js/references/components/Detail/Targets/__tests__/Targets.test.jsx
```diff
@@ -1,18 +1,22 @@
...
-import { createStore } from "redux";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { renderWithProviders } from "../../../../../../tests/setupTests";
 import { checkReferenceRight } from "../../../../selectors";
 import { mapDispatchToProps, mapStateToProps, Targets } from "../Targets";
 
 vi.mock("../../../../selectors.js");
 
-const createAppStore = state => () => createStore(state => state, state);
+const createAppStore = state => () =>
```
igboyes: One thing we're trying to do in the frontend codebase is change top level functions from arrows to `function` declarations.

In this, it means `createAppStore` would use `function` and the function is returning would be an arrow function.

```js
function createAppStore(state) {
  return () => createStore(state => state, state)
}
```

Please do this everywhere in this PR where it makes sense.

igboyes (changes requested):
One minor nitpick. Looks good otherwise.

ReeceHoffmann (approved):

PR #263, opened 2023-09-25
Add progress circle to index list view

ReeceHoffmann:
Changes:

 - Adds progress indicators to Index list view
 - Updates list view to use react-query


src/js/app/websocket.ts
```diff
@@ -17,21 +17,60 @@ import { accountKeys } from "../account/querys";
...
+}
+
+/**  Get resource keys from interface name */
 const keyFactories = {
     account: accountKeys,
     groups: groupQueryKeys,
     roles: roleQueryKeys,
     uploads: fileQueryKeys,
     users: userQueryKeys,
+    indexes: indexQueryKeys,
```
igboyes: Maintain alpha sort.

src/tests/fake/indexes.ts
```diff
@@ -0,0 +1,78 @@
...
+    version?: number;
+};
+
+export function createFakeIndexNested(props?: CreateFakeIndexNestedProps): IndexNested {
+    const defaultIndexNested = {
+        id: faker.random.alphaNumeric(8),
+        version: faker.datatype.number(5),
+    };
+
+    return merge(defaultIndexNested, props);
```
igboyes, line 24: Same thing?
```suggestion
    return {
        id: faker.random.alphaNumeric(8),
        version: faker.datatype.number(5),
        ...props
    };
```

src/tests/fake/indexes.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type createFakeIndexMinimalProps = CreateFakeIndexNestedProps & {
```
igboyes: I believe this is TS convention. This inconsistency is introduced in a few other spots as well.
```suggestion
type CreateFakeIndexMinimalProps = CreateFakeIndexNestedProps & {
```

src/js/indexes/components/RebuildAlert.tsx
```diff
@@ -3,9 +3,17 @@ import { connect } from "react-redux";
 import { Link } from "react-router-dom";
 import { Alert, Icon } from "../../base";
 import { checkReferenceRight } from "../../references/selectors";
+import { useInfiniteFindIndexes } from "../querys";
 
-export const RebuildAlert = ({ refId, showCountAlert, showIndexAlert }) => {
-    if (showCountAlert === 0) {
+export function RebuildAlert({ refId, hasRights }) {
```
igboyes: Shouldn't typing go here?

src/js/app/websocket.ts
```diff
@@ -17,21 +17,60 @@ import { accountKeys } from "../account/querys";
 import { roleQueryKeys, userQueryKeys } from "../administration/querys";
 import { fileQueryKeys } from "../files/querys";
 import { groupQueryKeys } from "../groups/querys";
+import { indexQueryKeys } from "../indexes/querys";
 
+/** Get effected resource query keys from job name  */
```
igboyes:
```suggestion
/** Get affected resource query keys by workflow name.  */
```

src/js/app/websocket.ts
```diff
@@ -17,21 +17,60 @@ import { accountKeys } from "../account/querys";
...
+const workflowQueries = {
+    build_index: [indexQueryKeys.lists()],
+};
+
+/**
+ * Invalidate queries that are effected by a job update
```
igboyes:
```suggestion
 * Invalidate queries that are affected by a job update
```

src/tests/fake/references.ts
```diff
@@ -0,0 +1,19 @@
+import { faker } from "@faker-js/faker";
+import { merge } from "lodash";
+import { ReferenceDataType } from "../../js/references/types";
+
+type createFakeReferenceNestedProps = {
```
igboyes:
```suggestion
type CreateFakeReferenceNestedProps = {
```

src/tests/fake/utils.ts
```diff
@@ -0,0 +1,5 @@
+export type FakeSearchResults = {
+    page_count?: number;
+    per_page?: number;
+    total_count?: number;
```
igboyes, line 4: Shouldn't these properties always be defined?

igboyes (changes requested):
* Use `function` in touched modules.
* Follow one component export per module. This optimizes the Vite build and HMR.

src/tests/fake/indexes.ts
ReeceHoffmann: Yeah it basically does, but if props is undefined it throws an error. This could be easily fixed by assigning an empty object to props if nothing is passed. But this soln reads more clearly to me

src/tests/fake/utils.ts
ReeceHoffmann: Ah no, which points to a naming mistake. The idea here was that it would work as a base model for allowing the fake functions to take optional manual overrides for the search result data returned by nock. `fakeSearchResultOptions`?

PR #266, opened 2023-09-26
Rastinrashidi/UI 54 use react query for creating the first user

raastinn:


src/js/wall/FirstUser.jsx
```diff
@@ -12,9 +11,23 @@ const initialValues = {
     password: "",
 };
 
-export const FirstUser = ({ onSubmit, errors }) => {
+export default function FirstUser() {
+    const [error, setError] = useState("");
```
ReeceHoffmann: Why not use the error handling baked into useMutation? https://tanstack.com/query/v4/docs/react/guides/mutations

ReeceHoffmann (commented):
See if you can use the built in error handling from `useMutation`, otherwise this looks good to me!

src/js/wall/FirstUser.jsx
```diff
@@ -12,9 +11,23 @@ const initialValues = {
     password: "",
 };
 
-export const FirstUser = ({ onSubmit, errors }) => {
+export default function FirstUser() {
```
ReeceHoffmann, line 19: Please add a docstring for `FirstUser`

src/js/wall/FirstUser.jsx
```diff
@@ -12,11 +11,23 @@ const initialValues = {
     password: "",
 };
 
-export const FirstUser = ({ onSubmit, errors }) => {
+/**
+ * Responsible for creating the first Virtool user
```
igboyes: Think of component docstrings like you are telling another developer what they will get from it.
```suggestion
 * A form for creating the first instance user.
```
Take a look at these docs for examples like:
* A vertically stacked set of interactive headings that each reveal a section of content.
* A date field component that allows users to enter and edit date.
* A control that allows the user to toggle between checked and not checked.

igboyes (changes requested):

PR #268, opened 2023-09-27
Add ml model list view

ReeceHoffmann:
Adds a barebones ML model list view

**Changes:**
  - adds the ml model list view + tests
  - Add ML model typing
  - Add ML model data faking

**Screenshots:**

Complete model:
image

`current_release = null`:
image

No Models:
image


src/js/app/websocket.ts
```diff
@@ -17,13 +17,15 @@ import { accountKeys } from "../account/querys";
...
 import { groupQueryKeys } from "../groups/querys";
+import { modelQueryKeys } from "../ml/queries";
 
 const keyFactories = {
     account: accountKeys,
     groups: groupQueryKeys,
     roles: roleQueryKeys,
     uploads: fileQueryKeys,
     users: userQueryKeys,
+    models: modelQueryKeys,
```
igboyes: Maintain order.

src/js/ml/components/MLModel.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Version tag for a MLModel
+ *
+ * @param name - The name of the version
+ * @param github_url - URL to the model on github
+ * @returns A version tag
+ */
+
+function ModelVersion({ name, github_url }: { name: string; github_url: string }) {
```
igboyes: Why separate component for this? I prefer no 'v' in front of the version.

src/tests/fake/ml.ts
```diff
@@ -0,0 +1,93 @@
...
+    ready?: boolean;
+    size?: number;
+};
+
+/**
+ * Create a fake ML model release object
+ *
+ * @param props - optional properties for creating a fake ML models with specific values
+ */
+export function createFakeMLModelRelease(props?: CreateFakeMLModelReleaseProps): MLModelRelease {
```
igboyes: I find the 'props' naming confusing since this is not a component. The naming of the type could be especially confusing when using autocomplete or code search. Maybe `overrides`? I'll leave it to you.

Do we really need to allow any of the model fields to be overridden? I think we should have no overriding possible unless use cases arise.

igboyes (changes requested):
This looks great.

src/js/ml/components/MLModel.tsx
ReeceHoffmann: Good catch, probably a more complicated earlier version. Will be single component with no "V" in next push

igboyes (approved):

PR #269, opened 2023-09-28
fix: create new useUrlSearchParams custom hook

raastinn:


src/js/subtraction/components/Toolbar.jsx
```diff
@@ -2,31 +2,38 @@ import React from "react";
...
 export const mapStateToProps = state => ({
-    term: state.subtraction.term || "",
     canModify: checkAdminRoleOrPermission(state, "modify_subtraction"),
 });
 
 export const mapDispatchToProps = dispatch => ({
-    onFind: e => {
-        dispatch(findSubtractions(e.target.value || null, 1));
+    onFind: term => {
+        dispatch(findSubtractions(term));
```
ReeceHoffmann: Does this always work correctly without specifying the page?

src/js/utils/hooks.js
```diff
@@ -42,3 +42,38 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
     }, deps);
 };
+
+export function getUrlSearchParams(key) {
+    const params = new URLSearchParams(window.location.search);
+    return params.get(key) || "";
+}
+
+export function updateUrlSearchParams(newTerm, key) {
+    if (window.location.search) {
```
ReeceHoffmann: Wouldn't you still want to set params even if `window.location.search` is empty? Also is this check required? URLSearchParams can probably handle either situation

src/js/subtraction/components/Toolbar.jsx
```diff
@@ -2,31 +2,38 @@ import React from "react";
 import { connect } from "react-redux";
 import { checkAdminRoleOrPermission } from "../../administration/utils";
 import { InputSearch, LinkButton, Toolbar } from "../../base";
+import { useUrlSearchParams } from "../../utils/hooks";
 import { findSubtractions } from "../actions";
 
-export const SubtractionToolbar = ({ term, onFind, canModify }) => {
+export const SubtractionToolbar = ({ onFind, canModify }) => {
```
ReeceHoffmann, line 15: docstring

src/js/subtraction/components/List.jsx
```diff
@@ -2,19 +2,22 @@ import React, { useEffect } from "react";
...
 import { Badge, LegacyScrollList, LoadingPlaceholder, NoneFoundBox, ViewHeader, ViewHeaderTitle } from "../../base";
+import { useUrlSearchParams } from "../../utils/hooks";
 import { findSubtractions } from "../actions";
 import SubtractionItem from "./Item";
 import SubtractionToolbar from "./Toolbar";
 
 const renderRow = index => <SubtractionItem key={index} index={index} />;
 
-export function SubtractionList({ onLoadNextPage, documents, page, page_count, total_count, fetched, term }) {
+export function SubtractionList({ onLoadNextPage, documents, page, page_count, total_count, fetched }) {
```
ReeceHoffmann, line 23: docstring + typing for function signature.

src/js/subtraction/components/__tests__/Toolbar.test.jsx
```diff
@@ -32,59 +31,39 @@ describe("<SubtractionToolbar />", () => {
...
+        const term = {
             target: {
                 value: "Foo",
             },
         };
-        wrapper.find(InputSearch).simulate("change", e);
-        expect(props.onFind).toHaveBeenCalledWith(e);
+        wrapper.find(InputSearch).simulate("change", term);
+        expect(props.onFind).toHaveBeenCalledWith("Foo");
     });
```
ReeceHoffmann: All these component tests need to be updated to use RTL

src/js/utils/hooks.js
```diff
@@ -42,3 +42,38 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+        const search = new URLSearchParams(window.location.search);
+        search.set(key, newTerm);
+        const newUrl = `${window.location.pathname}?${search.toString()}`;
+        window.history.replaceState({}, "", newUrl);
```
ReeceHoffmann: Is this the default history object? We use a streamlining library that acts as a shim over top of the built in history API to make it easier to work with for react-router. Would make for a cleaner call here as well https://github.com/remix-run/history/tree/v4/docs.

src/js/utils/hooks.js
```diff
@@ -42,3 +42,38 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+    }
+}
+
+export function useUrlSearchParams({ key, defaultValue = "" }) {
+    const [value, setValue] = useState(defaultValue);
```
ReeceHoffmann: I think you can remove this usage of state. Get the value from the URL which removes the possibility of desync. should also simplify the rest of the hook

src/js/utils/hooks.js
```diff
@@ -42,3 +42,38 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+    useEffect(() => {
+        const searchTerm = getUrlSearchParams(key);
+        if (searchTerm) {
+            setValue(searchTerm);
+        } else {
+            updateUrlSearchParams(defaultValue, key);
+        }
+    }, [key, defaultValue]);
+    const setUrlValue = newValues => {
```
ReeceHoffmann: nitpick:
```suggestion
    }, [key, defaultValue]);
    
    const setUrlValue = newValues => {
```

src/js/utils/hooks.js
```diff
@@ -42,3 +42,38 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+export function updateUrlSearchParams(newTerm, key) {
+    if (window.location.search) {
+        const search = new URLSearchParams(window.location.search);
+        search.set(key, newTerm);
+        const newUrl = `${window.location.pathname}?${search.toString()}`;
+        window.history.replaceState({}, "", newUrl);
+    }
+}
+
+export function useUrlSearchParams({ key, defaultValue = "" }) {
```
ReeceHoffmann: docstring

ReeceHoffmann (changes requested):

igboyes (approved):

src/js/utils/hooks.js
```diff
@@ -42,3 +42,41 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
         firstRef.current = true;
     }, deps);
 };
+
+export function getUrlSearchParams(key) {
```
igboyes: * Is this used anywhere else than line 66 of this module? If not, no need to export. Further, I don't think a separate function is much benefit.
* The naming is not good. It doesn't get params, it gets a search term.

src/js/utils/hooks.js
```diff
@@ -42,3 +42,41 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+    const params = new URLSearchParams(window.location.search);
+    return params.get(key) || "";
+}
+
+export function updateUrlSearchParams(newTerm, key) {
+    const search = new URLSearchParams(window.location.search);
+    search.set(key, newTerm);
+    const newUrl = `${window.location.pathname}?${search.toString()}`;
+    window.history.replaceState({}, "", newUrl);
+}
```
igboyes: The `newUrl` variable is not actually a URL. It is a URL path. You can avoid the naming issue by just doing.
```suggestion
export function updateUrlSearchParams(newTerm, key) {
    const search = new URLSearchParams(window.location.search);
    search.set(key, newTerm);
    window.history.replaceState({}, "", `${window.location.pathname}?${search.toString()}`);
}
```
---
* Variable naming is confusing. See other comments
* Don't export if not used in another module (I did not check thoroughly if this is the case).

src/js/utils/hooks.js
```diff
@@ -42,3 +42,41 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
     }, deps);
 };
+
+export function getUrlSearchParams(key) {
+    const params = new URLSearchParams(window.location.search);
+    return params.get(key) || "";
+}
+
+export function updateUrlSearchParams(newTerm, key) {
+    const search = new URLSearchParams(window.location.search);
```
igboyes: You use `params` as the name for an object of the same type in the above function. Be consistent. I'd go with `params` as `search` implies action.

src/js/utils/hooks.js
```diff
@@ -42,3 +42,41 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
         firstRef.current = true;
     }, deps);
 };
+
+export function getUrlSearchParams(key) {
+    const params = new URLSearchParams(window.location.search);
+    return params.get(key) || "";
+}
+
+export function updateUrlSearchParams(newTerm, key) {
```
igboyes: You write this generically for setting any search parameter, but it uses variable naming specific to the `term` use case.

```suggestion
export function updateUrlSearchParams(value, key) {
```

src/js/utils/hooks.js
```diff
@@ -42,3 +42,41 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+export function useUrlSearchParams({ key, defaultValue }) {
+    const searchTerm = getUrlSearchParams(key);
+
+    useEffect(() => {
+        if (!searchTerm && defaultValue) {
+            updateUrlSearchParams(defaultValue, key);
+        }
+    }, [key, defaultValue, searchTerm]);
+
+    const setUrlValue = newValue => {
```
igboyes: * Use `function`.
* Why call this `setUrlValue`, then return it as `setValue`?

src/js/utils/hooks.js
```diff
@@ -42,3 +42,41 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+
+/**
+ * Hook for managing and synchronizing URL search parameters with a component's state
+ *
+ * @param key - The search parameter key to be managed
+ * @param defaultValue - The default value to use when the search parameter is not present in the URL
+ * @returns Object - An object containing the current value and a function to set the URL search parameter
+ */
+export function useUrlSearchParams({ key, defaultValue }) {
+    const searchTerm = getUrlSearchParams(key);
```
igboyes: This is a generic hook for managing any search parameter. Why are you using the word 'term'?

src/js/utils/hooks.js
```diff
@@ -42,3 +42,41 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+    search.set(key, newTerm);
+    const newUrl = `${window.location.pathname}?${search.toString()}`;
+    window.history.replaceState({}, "", newUrl);
+}
+
+/**
+ * Hook for managing and synchronizing URL search parameters with a component's state
+ *
+ * @param key - The search parameter key to be managed
+ * @param defaultValue - The default value to use when the search parameter is not present in the URL
```
igboyes:
```suggestion
 * @param defaultValue - The default value to use when the search parameter key is not present in the URL
```

igboyes (changes requested):
* Please update the files you touched to use TS and use `function` declarations.
* The naming in hooks is confused and inconsistent. Pay close attention to consistency and meaning in your names.

raastinn:
- After testing this again, I realised that the `setValue` served no purpose. With my previous code, if you refreshed on a filtered page, it didn't allow you to edit the text in the toolbar. This version fixes that and works as intended, using `useState` in Toolbar.jsx. 
- I only converted the `hooks.js` file to a TS file, as another issue already requires me converting the touched Subtraction files to TS.

src/js/utils/hooks.ts
```diff
@@ -43,40 +43,30 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
  * @returns Object - An object containing the current value and a function to set the URL search parameter
  */
 export function useUrlSearchParams({ key, defaultValue }) {
-    const searchTerm = getUrlSearchParams(key);
+    const params = new URLSearchParams(window.location.search);
+    const searchValue = params.get(key);
 
     useEffect(() => {
-        if (!searchTerm && defaultValue) {
+        if (!searchValue && defaultValue) {
```
igboyes: Why is there a naming difference between `searchValue` in this hook and `value` in `updateUrlSearchParams()`?

src/js/utils/hooks.ts
```diff
@@ -43,40 +43,30 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
-
-    const setUrlValue = newValue => {
-        updateUrlSearchParams(newValue, key);
-    };
+    }, [key, defaultValue]);
 
     return {
-        value: searchTerm || defaultValue,
-        setValue: setUrlValue,
+        value: searchValue || defaultValue,
```
igboyes: There is a potential bug here. Think about what values `searchValue` could have.

igboyes (changes requested):

ReeceHoffmann:
> * After testing this again, I realised that the `setValue` served no purpose.

There are two concerns I have with this change:

1. It removes the ability for the hook to be able to write the URLSearchParams, reducing its usefulness. (Think about cases where your using react-query instead of a prebuilt saga and need to update the URL)

~~2. When passing defaultValues they no longer will get written to the URL. As a result the default value being returned does not reflect the actual URL search params.~~ Nevermind, On reread see that this case is still covered

Happy to talk these over if you want.

src/js/utils/hooks.ts
```diff
@@ -42,3 +42,31 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+}
+
+/**
+ * Hook for managing and synchronizing URL search parameters with a component's state
+ *
+ * @param key - The search parameter key to be managed
+ * @param defaultValue - The default value to use when the search parameter key is not present in the URL
+ * @returns Object - An object containing the current value and a function to set the URL search parameter
+ */
+export function useUrlSearchParams({ key, defaultValue }) {
```
ReeceHoffmann: Needs typing

src/js/utils/hooks.ts
```diff
@@ -42,3 +42,31 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
...
+    const value = params.get(key) || "";
+
+    useEffect(() => {
+        if (!value && defaultValue) {
+            updateUrlSearchParams(defaultValue, key);
+        }
+    }, [key, defaultValue]);
+
+    return {
+        value: value || defaultValue,
```
ReeceHoffmann: To extend Ian's comment, what happens if the `value` is `falsy` (ex: `""`) but correct.

ReeceHoffmann: Nitpick: I would use an array instead of an object. Since the calling function should always want values and will often want `setValue` using an array allows for nicer renaming syntax on destructure with little drawback much like how `react` implements `useState`

src/js/utils/hooks.ts
ReeceHoffmann: Nitpick: I would use positional arguments here. Using an object is more cumbersome and doesn't obviously provide a benefit here

src/js/utils/hooks.ts
```diff
@@ -42,3 +42,31 @@ export const useDidUpdateEffect = (onUpdate, deps) => {
         firstRef.current = true;
     }, deps);
 };
+
+function updateUrlSearchParams(value, key) {
```
ReeceHoffmann: types + docstring

raastinn:
That makes more sense. I did some testing to see how it'll be once Redux is removed and you were right. Added it back

ReeceHoffmann (changes requested):
Looking pretty good, mostly styling nitpicks

src/js/utils/hooks.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Hook for managing and synchronizing URL search parameters with a component's state
+ *
+ * @param key - The search parameter key to be managed
+ * @param defaultValue - The default value to use when the search parameter key is not present in the URL
+ * @returns Object - An object containing the current value and a function to set the URL search parameter
+ */
+export function useUrlSearchParams(key: string, defaultValue?: string): [string, (newValue: string) => void] {
+    const params = new URLSearchParams(window.location.search);
+    const refValue = useRef(params.get(key) || defaultValue);
```
ReeceHoffmann: This still leaves open the possibility of desync between the URL and the state reported back. Imagine the case that the hook is called in two places, one to write (ex. toolbar) and one to read (ex. list view).

ReeceHoffmann (changes requested):

src/js/subtraction/components/List.jsx
```diff
@@ -2,27 +2,39 @@ import React, { useEffect } from "react";
...
+ * @param onLoadNextPage - Function to load the next page of subtraction operations
+ * @param documents - List of subtraction documents to display
+ * @param page - The current page of list
+ * @param page_count - The total number of pages
+ * @param total_count - The total number of subtractions
+ * @param fetched - Indicates whether subtraction data has been fetched
+ * @returns SubtractionList - A list of subtractions
+ */
+export function SubtractionList({ onLoadNextPage, documents, page, page_count, total_count, fetched }) {
+    const [value] = useUrlSearchParams("find");
```
ReeceHoffmann:
```suggestion
    const [term] = useUrlSearchParams("find");
```
Nitpick: More descriptive

src/js/subtraction/components/Toolbar.jsx
```diff
@@ -1,32 +1,50 @@
...
+/**
+ * Managing the toolbar search filtering
+ *
+ * @param onFind - The dispatch action to update the search filtering results
+ * @param canModify - The user's permissions to modify
+ * @returns Toolbar - A search filtering toolbar
+ */
+export const SubtractionToolbar = ({ onFind, canModify }) => {
+    const [value, setValue] = useUrlSearchParams("find");
+    const [term, setTerm] = useState(value);
```
ReeceHoffmann: I don't think you need this since the term is stored in the URL. Removing should reduce opportunity for desync between the URL and searchbar

src/js/subtraction/components/Toolbar.jsx
```diff
@@ -1,32 +1,50 @@
...
-    let createButton;
+/**
+ * Managing the toolbar search filtering
+ *
+ * @param onFind - The dispatch action to update the search filtering results
+ * @param canModify - The user's permissions to modify
+ * @returns Toolbar - A search filtering toolbar
+ */
+export const SubtractionToolbar = ({ onFind, canModify }) => {
+    const [value, setValue] = useUrlSearchParams("find");
```
ReeceHoffmann:
```suggestion
    const [term, setTerm] = useUrlSearchParams("find");
```
Nitpick: more descriptive imo

ReeceHoffmann (changes requested):
Few small tweaks and I think this is ready to merge

ReeceHoffmann (approved):

PR #276, opened 2023-10-18
Prevent crash from unknown workflows in detailed jobs view

ReeceHoffmann:
Notes: 

  - List view should already cope with unknown workflows w/o causing a crash
  - Detailed view shouldn't crash, but would render the head of the job arguments table without any arguments
    - added generic handler that renders all arguments with a type of `string` or `number`
  - Steps should render as expected since most step name and descriptions are now taken directly from the `Status` object itself

Example generic arguments: (index build forced to use the generic rows)
image


ReeceHoffmann:
Updated args list:
image

src/js/jobs/components/JobArgs.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+        case "update_sample":
+            return <UpdateSampleRows {...args} />;
+
+        default:
+            return <GenericJobArgsRows args={args} />;
+    }
+}
+
+type JobArgsProps = {
```
igboyes: Why is there a separate `type` for this component, but not the others.

src/js/jobs/components/JobArgs.tsx
```diff
@@ -0,0 +1,191 @@
...
+};
+
+/**
+ * A table of arguments used to run a job.
+ *
+ * @param workflow - The workflow name
+ * @param args - The complete list of arguments used to run the job
+ * @returns An table of arguments
+ */
+export function JobArgs({ workflow, args }: JobArgsProps) {
```
igboyes: This should be the only exported component from this module. ViteJS HMR and bundling works better following this convention.

src/js/jobs/components/JobArgs.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <JobArgsRow title="Subtraction">
+            <Link to={`/subtractions/${subtraction_id}`}>{subtraction_id}</Link>
+        </JobArgsRow>
+    );
+}
+
+/**
+ * Rows showing important arguments when running an "update_sample" workflow.
+ *
+ * @param args - The complete list of arguments used to run the job
```
igboyes: This may have happened while you were away...

I found a lot of opinions online saying that documenting the props `type` for the component in TS is better than using `@param`.

I have to agree. This one is a good example where `@param` line is a bit distant from the function signature and is less descriptive than a type would be.

src/js/jobs/components/JobArgs.tsx
```diff
@@ -0,0 +1,191 @@
...
+ * @returns Table rows containing the arguments of the job
```
igboyes: I think this is redunant when it comes to functional components. If you switch to documenting the type, problematic decision goes away.

src/js/jobs/components/JobArgs.tsx
```diff
@@ -0,0 +1,191 @@
...
+ */
+export function UpdateSampleRows({ sample_id }: { sample_id: string }) {
```
igboyes: I think this may be a relic of a job that no longer exists...

src/js/jobs/components/JobArgs.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+const UnsupportedJobArgsRow = styled(JobArgsRow)`
+    th {
+        font-weight: normal;
+        font-family: monospace;
+    }
+`;
+
+/**
+ * Generic rows displaying the arguments passed to the job when the workflow type is not know.
```
igboyes:
```suggestion
 * Generic rows displaying the arguments passed to the job when the workflow type is not known.
```

src/js/jobs/components/JobArgs.tsx
```diff
@@ -0,0 +1,191 @@
...
+                    );
+                }
+                return null;
+            })}
+        </>
+    );
+}
+
+/**
+ * The tabel rows containing arguments used to run a job.
```
igboyes:
```suggestion
 * The table rows containing arguments used to run a job.
```

src/js/jobs/components/JobArgs.tsx
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * @param workflow - The workflow name
+ * @param args - The complete list of arguments used to run the job
+ * @returns Table rows containing arguments used to run a job
+ */
+export function JobArgsRows({ workflow, args }) {
```
igboyes: I am not sure why some of these components are exported. Testing?

igboyes (changes requested):

src/js/jobs/components/JobArgs.tsx
ReeceHoffmann: ~~Ah leftover cruft. The original defn for `args` was more complex making it a bit verbose for inline imo. New version has it inline~~ This is wrong now, all  of the components are being adjusted to have external type defns with docstrings

src/js/jobs/components/JobArgs.tsx
ReeceHoffmann: I was wondering... I hadn't heard of this job before. Any chance of it still showing up if the user goes back far enough?

igboyes: It will still be handled by your changes with no loss in functionality. Since job records could be deleted when Update Sample was around, they are probably mostly gone anyway.

src/js/jobs/components/JobArgs.tsx
ReeceHoffmann: Yeah, testing updated to RTL, and exports removed

PR #277, opened 2023-10-20
Ensure analysis list renders correctly with new workflow

ReeceHoffmann:
Deal with analysis list and detail view in same PR:

Unsupported detail view: (pathoscope used as an example, is recognised in pushed version)
image

Unsupported list view: (pathoscope used as an example, is recognised in pushed version)
image


ReeceHoffmann:
Updated screenshots:
image

image

PR #278, opened 2023-10-20
chore: overhaul `CreateUser` form

raastinn:


src/js/users/components/CreateUserForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const [forceReset, setForceReset] = useState(false);
+
+    const {
+        formState: { errors },
+        register,
+        handleSubmit,
+    } = useForm({ defaultValues: { handle, password, forceReset } });
+
+    function handleForceReset() {
+        forceReset ? setForceReset(false) : setForceReset(true);
```
ReeceHoffmann: Nitpick:
```suggestion
       setForceReset(!forceReset);
```
 Also I would probably just attach this an an anonymous function

src/js/users/components/CreateUserForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+    error: string;
+    /** A callback function to be called when the form is submitted */
+    onSubmit: (data: { handle: string; password: string; forceReset: boolean }) => void;
+}
+
+/**
+ * A form component for creating a new user
+ */
+export function CreateUserForm({ handle = "", password = "", error, onSubmit }: CreateUserFormProps) {
+    const [forceReset, setForceReset] = useState(false);
```
ReeceHoffmann: Why separate state for `forceReset`? Why not just let it be managed by `react-hook-form`

src/js/users/components/CreateUserForm.tsx
```diff
@@ -0,0 +1,71 @@
...
+                    <InputLabel htmlFor="handle">Username</InputLabel>
+                    <InputSimple id="handle" {...register("handle", { required: "Please specify a username" })} />
+                    <InputError>{errors.handle?.message}</InputError>
+                </InputGroup>
+                <InputGroup>
+                    <InputLabel htmlFor="password">Password</InputLabel>
+                    <InputSimple
+                        id="password"
+                        type="password"
+                        {...register("password", { required: "Password does not meet minimum length requirement (8)" })}
```
ReeceHoffmann: Will this reject a password that exists but is too short?

src/js/users/components/CreateUser.tsx
```diff
@@ -0,0 +1,56 @@
...
+    password: string;
+    /** Whether the user will be forced to reset their password on next login */
+    forceReset: boolean;
+};
+
+/**
+ * A dialog for creating a new user
+ */
+export default function CreateUser() {
+    const [open, setOpen] = useState(false);
```
ReeceHoffmann: Modal control should be moved to the `state` of the browsers history object

src/js/users/components/CreateUser.tsx
```diff
@@ -0,0 +1,56 @@
...
+    const [error, setError] = useState("");
```
ReeceHoffmann: I think a lot of the error handling logic here is made redundant by the error handling features of `useMutation`. Take a look at the documentation of `useMutation` and see if you can trim down the error handling logic

ReeceHoffmann (changes requested):

src/js/users/components/CreateUser.tsx
```diff
@@ -0,0 +1,54 @@
+import { DialogPortal, DialogTrigger } from "@radix-ui/react-dialog";
+import React from "react";
+import { useMutation } from "react-query";
+import { Dialog, DialogContent, DialogOverlay, DialogTitle, Icon } from "../../base";
+import { StyledButton } from "../../base/styled/StyledButton";
+import { create } from "../api";
+import { CreateUserForm } from "./CreateUserForm";
+
+type NewUserProps = {
```
ReeceHoffmann: Nitpick: Props is generally reserved in the codebase for react components. I would probably name it just straight NewUser since it is the object needed to create a new user, but NewUserParams or some variant of would be fine too imo

src/js/users/components/CreateUser.tsx
```diff
@@ -0,0 +1,54 @@
...
+        mutation.mutate({ handle, password, forceReset });
+    }
+
+    function handleChange() {
+        mutation.reset();
+        history.replaceState({ state: !history.state?.state }, "");
+    }
+
+    return (
+        <Dialog open={history.state?.state} onOpenChange={() => handleChange()}>
```
ReeceHoffmann:
```suggestion
        <Dialog open={history.state?.state} onOpenChange={handleChange}>
```

src/js/users/components/CreateUser.tsx
```diff
@@ -0,0 +1,54 @@
...
+        },
+    });
+
+    function handleSubmit({ handle, password, forceReset }: NewUserProps) {
+        mutation.mutate({ handle, password, forceReset });
+    }
+
+    function handleChange() {
+        mutation.reset();
+        history.replaceState({ state: !history.state?.state }, "");
```
ReeceHoffmann:
```suggestion
        history.replaceState({ createUser: !history.state?.state }, "");
```

I would give the variable in state are more meaningful name than state. Also consider if you want the button needs the ability to close the dialogue or not.

src/js/users/components/CreateUser.tsx
```diff
@@ -0,0 +1,54 @@
...
+    forceReset: boolean;
+};
+
+/**
+ * A dialog for creating a new user
+ */
+export default function CreateUser() {
+    const mutation = useMutation(create, {
+        onSuccess: () => {
+            history.replaceState({ state: !history.state.state }, "");
```
ReeceHoffmann:
```suggestion
            history.replaceState({ state: false }, "");
```
?

src/js/users/components/CreateUser.tsx
```diff
@@ -0,0 +1,54 @@
...
+        },
+    });
+
+    function handleSubmit({ handle, password, forceReset }: NewUserProps) {
+        mutation.mutate({ handle, password, forceReset });
+    }
+
+    function handleChange() {
```
ReeceHoffmann:
```suggestion
    function onOpenChange() {
```

src/js/users/components/CreateUserForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    <InputLabel htmlFor="password">Password</InputLabel>
+                    <InputSimple
+                        id="password"
+                        type="password"
+                        {...register("password", {
+                            required: "Password does not meet minimum length requirement (8)",
+                            minLength: {
+                                value: 8,
+                                message: "Password does not meet minimum length requirement (8)",
+                            },
```
ReeceHoffmann, line 54: Does min-length imply required?, might be able to cut `required`

ReeceHoffmann (changes requested):

src/js/users/components/CreateUserForm.tsx
raastinn: I believe `required` is needed since `minLength` doesn't give a message if the input is empty. Its a bit redundant but we can replace `minLength` with the mutation error instead if that looks cleaner.

src/js/users/components/__tests__/ManageUsers.test.jsx
```diff
@@ -43,7 +43,6 @@ describe("<ManageUsers />", () => {
 
         renderWithRouter(<ManageUsers />, state, history, createReducer);
 
-        expect(await screen.findByLabelText("user-plus")).toBeInTheDocument();
```
ReeceHoffmann, line 46: This should be replaced with something new. This check is needed to ensure the button for creating users appears for admin users

ReeceHoffmann (approved):

PR #280, opened 2023-10-30
chore: use string type for all dates in object types

raastinn:
NOTE: also changed the formatting of docstrings in touched files to stay consistent throughout the code


ReeceHoffmann:
Ian pointed out that we should really be instead casting the data prior to passing it back to the components. Working on an writing up an issue to address this. Closing this and the underlying issue as well while I come up with a soln

PR #283, opened 2023-11-07
refactor: add pagination to samples list view

raastinn:


src/js/analyses/components/Create/Quick.jsx
```diff
@@ -62,41 +56,85 @@ const validationSchema = Yup.object().shape({
...
+    const location = useLocation();
+
+    if (location.state?.quickAnalysis === true) {
+        if (libraryType === "amplicon") {
+            return "barcode";
+        }
+
+        return "genome";
+    }
+    return false;
```
ReeceHoffmann: The return value of this function seems overloaded to me. A single return value should stick to either determining the mode or whether to show the modal. Also not sure if this totally makes sense as a hook, guessing it could be paired down quite a bit.

src/js/analyses/components/Create/Quick.jsx
```diff
@@ -62,41 +56,85 @@ const validationSchema = Yup.object().shape({
...
+
+        return sample.library_type === "normal" || sample.library_type === "srna";
+    });
+}
+
+export function QuickAnalyze({ samples, subtractionOptions, onShortlistSubtractions, onClear }) {
+    const mode = useQuickAnalysisMode(samples.library_type);
+    const compatibleSamples = getCompatibleSamples(mode, samples);
+    const { data: hasHmm } = useFindHmms();
+    const mutation = useMutation(analyze);
```
ReeceHoffmann: Are there any queries that need to update on a successful analysis creation?

src/js/analyses/components/Create/Quick.jsx
```diff
@@ -62,41 +56,85 @@ const validationSchema = Yup.object().shape({
...
             onHide();
         }
     }, [mode]);
 
+    function referenceId(selectedIndexes) {
+        const selectedCompatibleIndexes = indexes.filter(index => selectedIndexes.includes(index.id));
+
+        const referenceIds = selectedCompatibleIndexes.map(index => index.reference.id);
+
+        return Array.from(new Set(referenceIds));
```
ReeceHoffmann: lodash: `uniq(By)` https://lodash.com/docs/4.17.15#uniqBy

src/js/analyses/components/Create/Quick.jsx
```diff
@@ -62,41 +56,85 @@ const validationSchema = Yup.object().shape({
...
+        }
+
+        return sample.library_type === "normal" || sample.library_type === "srna";
+    });
+}
+
+export function QuickAnalyze({ samples, subtractionOptions, onShortlistSubtractions, onClear }) {
+    const mode = useQuickAnalysisMode(samples.library_type);
+    const compatibleSamples = getCompatibleSamples(mode, samples);
+    const { data: hasHmm } = useFindHmms();
```
ReeceHoffmann: Naming here wonky. `hasVar` namin scheme implies a bool, this is an object... also what about the case where the query has not yet resolved?

src/js/hmm/querys.ts
```diff
@@ -0,0 +1,10 @@
+import { useQuery } from "react-query";
+import { Request } from "../app/request";
+
+export function findHmms() {
+    return Request.get("/hmms").then(res => res.body);
+}
+
+export function useFindHmms() {
+    return useQuery("hmms", findHmms);
```
ReeceHoffmann: Typing

src/js/hmm/querys.ts
```diff
@@ -0,0 +1,10 @@
+import { useQuery } from "react-query";
+import { Request } from "../app/request";
+
+export function findHmms() {
+    return Request.get("/hmms").then(res => res.body);
+}
```
ReeceHoffmann: Should be in `api` file

src/js/indexes/querys.ts
```diff
@@ -36,3 +37,13 @@ export function useInfiniteFindIndexes(refId: string, term?: string) {
...
 }
+
+export function listReadyIndexes() {
+    return Request.get("/indexes")
+        .query({ ready: true })
+        .then(res => res.body);
+}
+
+export function useListReadyIndexes() {
+    return useQuery("listReadyIndex", listReadyIndexes);
```
ReeceHoffmann: typing

src/js/indexes/querys.ts
```diff
@@ -36,3 +37,13 @@ export function useInfiniteFindIndexes(refId: string, term?: string) {
         },
     );
 }
+
+export function listReadyIndexes() {
+    return Request.get("/indexes")
+        .query({ ready: true })
+        .then(res => res.body);
+}
```
ReeceHoffmann: api file

src/js/samples/components/Filter/LabelFilter.jsx
```diff
@@ -3,21 +3,31 @@ import React, { useState } from "react";
...
+        labels && Array.isArray(labels)
+            ? labels.map(label => (
+                  <LabelFilterItem
+                      key={label.id}
+                      {...label}
+                      pressed={selected.includes(label.id)}
+                      onClick={handleClick}
+                  />
+              ))
+            : null;
```
ReeceHoffmann: 1. Why not just use `isLoading`
2. How does this look when with higher latency on a fresh load? I have this sneaking suspicion that the labels popping in after the rest of the page loads might look a bit janky. might need to preload labels with the rest of the list

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+const StyledSamplesList = styled.div`
+    display: grid;
+    grid-column-gap: ${props => props.theme.gap.column};
+    grid-template-columns: minmax(auto, 1150px) max(320px, 10%);
+`;
+
+export default function SamplesList() {
+    const location = useLocation();
+    const URLPage = parseInt(new URLSearchParams(location.search).get("page")) || 1;
```
ReeceHoffmann: Why not use the SearchParams hook for getting and setting page?

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const [term, setTerm] = useUrlSearchParams("find");
+    const { data, isLoading } = useFindSamples(URLPage, 4, term);
+    const [selected, setSelected] = useState([]);
+
+    if (isLoading) {
+        return <LoadingPlaceholder />;
+    }
+    const { documents, page, page_count, total_count } = data;
```
ReeceHoffmann:
```suggestion
    }
    
    const { documents, page, page_count, total_count } = data;
```

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+    display: grid;
+    grid-column-gap: ${props => props.theme.gap.column};
+    grid-template-columns: minmax(auto, 1150px) max(320px, 10%);
+`;
+
+export default function SamplesList() {
+    const location = useLocation();
+    const URLPage = parseInt(new URLSearchParams(location.search).get("page")) || 1;
+    const [term, setTerm] = useUrlSearchParams("find");
+    const { data, isLoading } = useFindSamples(URLPage, 4, term);
```
ReeceHoffmann:
```suggestion
    const { data: samples, isLoading } = useFindSamples(URLPage, 4, term);
```

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+    const [selected, setSelected] = useState([]);
+
+    if (isLoading) {
+        return <LoadingPlaceholder />;
+    }
+    const { documents, page, page_count, total_count } = data;
+
+    function onClear() {
+        setSelected([]);
+    }
```
ReeceHoffmann: Nitpick: simple enough I would define as an anon attached to the component using it

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+                library_type={document.library_type}
+                name={document.name}
+                ready={document.ready}
+                workflows={document.workflows}
+                handle={document.user.handle}
+                checked={selected.some(item => item.id === document.id)}
+                selected={selected}
+                onSelect={handleSelect}
+                select={select}
+                document={document}
```
ReeceHoffmann: Redefine `SampleItem` as taking a sample as part of it's props. Way simpler interface

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+                labels={document.labels}
+                library_type={document.library_type}
+                name={document.name}
+                ready={document.ready}
+                workflows={document.workflows}
+                handle={document.user.handle}
+                checked={selected.some(item => item.id === document.id)}
+                selected={selected}
+                onSelect={handleSelect}
+                select={select}
```
ReeceHoffmann: These two seem too similar, might just be a naming problem though

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+        setSelected([]);
+    }
+
+    function renderRow(document) {
+        function handleSelect() {
+            if (!selected.includes(document)) {
+                setSelected(selected => [...selected, document]);
+            } else {
+                setSelected(prevSelected => prevSelected.filter(item => item !== document));
+            }
```
ReeceHoffmann: xor?

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        function select() {
+            if (!selected.includes(document)) {
+                setSelected(selected => [...selected, document]);
+            }
```
ReeceHoffmann: union?

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        return (
+            <SampleItem
+                {...document}
+                key={document.id}
+                id={document.id}
+                index={document.id}
```
ReeceHoffmann: destructure seems duplicate with all the explicit defns

src/js/samples/components/SamplesToolbar.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+const mapDispatchToProps = dispatch => ({
+    onQuickAnalyze: () => {
+        dispatch(pushState({ quickAnalysis: true }));
+    },
+
+    onSelect: sampleId => {
+        // There is something wrong with this...
+        // dispatch(toggleSelectSample(sampleId));
+    },
```
ReeceHoffmann: I don't think this is even used...

src/js/samples/components/Sidebar/ManageLabels.jsx
```diff
@@ -26,14 +27,42 @@ const StyledSideBarSection = styled(SideBarSection)`
...
-export const ManageLabels = ({
-    allLabels,
-    selectedSamples,
-    selectedLabels,
-    onLabelUpdate,
-    partiallySelectedLabels,
-}) => {
-    const onUpdate = label => onLabelUpdate(selectedSamples, selectedLabels, label);
+export function ManageLabels({ selectedSamples, partiallySelectedLabels, documents }) {
+    const { data: allLabels, isLoading } = useFetchLabels();
```
ReeceHoffmann:
```suggestion
    const { data: labels, isLoading } = useFetchLabels();
```

src/js/samples/querys.ts
```diff
@@ -0,0 +1,78 @@
...
+export const getLabelsFromURL = () => {
+    if (window.location.search) {
+        const search = new URLSearchParams(window.location.search);
+        const labels = search.getAll("label");
+
+        return labels;
+    }
+
+    return [];
+};
```
ReeceHoffmann: Not a query. Should be tweaking the custom hook to permit array values

ReeceHoffmann (changes requested):
Good start. A lot of the components currently feel a bit complex. I think a good chunk of the logic can be simplified. Take another run through and see what you can do. `SampleList` and `quickAnalyze` and `ManageLabels` would especially benefit from some extra attention.

Also please start including screenshots for PRs where there are visual changes

src/js/samples/components/SamplesToolbar.tsx
raastinn: Yeah this was there before. Wasnn't sure if you still needed it so I just left it. I'll take it down

src/js/analyses/components/Create/Quick.jsx
raastinn: If you're referring to `samplesQueryKeys.lists()` getting updated, I believe the changes in `websockets.ts` updates the status of the analyses in the sample item

raastinn:
Screenshot from 2023-11-16 11-26-15
_Used 8 per page for screenshot_

src/js/analyses/components/Create/Quick.jsx
ReeceHoffmann: Is is updating automatically for you? In my local instance there is no samples WS message sent

src/js/analyses/components/Create/Quick.jsx
```diff
@@ -62,42 +57,88 @@ const validationSchema = Yup.object().shape({
...
-            "reference.id",
-        );
-        onAnalyze(compatibleSamples, referenceIds, subtractions, accountId, workflows);
-        onUnselect(compatibleSamples.map(sample => sample.id));
-    };
+    if (isLoadingHmms || isLoadingIndexes) {
+        return <LoadingPlaceholder />;
+    }
+
+    function referenceId(selectedIndexes) {
```
ReeceHoffmann: nitpick: function is named like a variable

src/js/analyses/components/Create/Quick.jsx
```diff
@@ -62,42 +57,88 @@ const validationSchema = Yup.object().shape({
...
+    return filter(samples, sample => {
+        if (mode === "barcode") {
+            return sample.library_type === "amplicon";
+        }
+
+        return sample.library_type === "normal" || sample.library_type === "srna";
+    });
+}
+
+export function QuickAnalyze({ samples, subtractionOptions, onShortlistSubtractions, onClear }) {
```
ReeceHoffmann: Typing/docstrings generally missing from this module

src/js/hmm/querys.ts
```diff
@@ -0,0 +1,10 @@
+import { useQuery } from "react-query";
+import { listHmms } from "./api";
+import { HMMSearchResults } from "./types";
+
+/**
+ * Fetches a list of HMM search results from the API
+ */
```
ReeceHoffmann: `@returns`

src/js/indexes/api.ts
```diff
@@ -30,7 +30,9 @@ export function get({ indexId }) {
  * @returns A promise resolving to the API response containing the list of ready indexes
  */
 export function listReady() {
-    return Request.get("/indexes").query({ ready: true });
+    return Request.get("/indexes")
+        .query({ ready: true })
+        .then(res => res.body);
 }
```
ReeceHoffmann: Is this used anywhere else?

src/js/indexes/querys.ts
```diff
@@ -36,3 +36,10 @@ export function useInfiniteFindIndexes(refId: string, term?: string) {
         },
     );
 }
+
+/**
+ * Gets a list of ready indexes
+ */
```
ReeceHoffmann: `@returns`

src/js/samples/components/Filter/LabelFilter.tsx
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Sidebar for filtering samples by labels
+ */
+export default function LabelFilter({ labels, onClick, selected }: LabelFilterProps) {
+    return (
+        <SideBarSection>
+            <SidebarHeader>
+                Labels <Link to="/samples/labels">Manage</Link>
+            </SidebarHeader>
+            {Array.isArray(labels)
```
ReeceHoffmann: You need this check? typing as written suggest it should never be needed

src/js/indexes/querys.ts
```diff
@@ -36,3 +36,10 @@ export function useInfiniteFindIndexes(refId: string, term?: string) {
...
+export function useListReadyIndexes() {
+    return useQuery<IndexSearchResult>("listReadyIndex", listReady);
+}
```
ReeceHoffmann: Should be a general index fetching function that takes query params

src/js/samples/querys.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (labels) {
+        labels.forEach(label => request.query({ label }));
+    }
+
+    if (workflows) {
+        request.query({ workflows });
+    }
+
+    return request.then(res => res.body);
+}
```
ReeceHoffmann: api.ts

src/js/utils/hooks.ts
```diff
@@ -91,3 +92,52 @@ export function useUrlSearchParams(key: string, defaultValue?: string): [string,
...
+export function useUrlSearchParamsList(key: string, defaultValue?: string[]): [string[], (newValue: string) => void] {
+    const history = useHistory();
+    const location = useLocation();
+    const firstRender = useRef(true);
+
+    let value = new URLSearchParams(location.search).getAll(key);
+
+    if (firstRender.current && defaultValue && !value.length) {
+        value = defaultValue;
+        value.forEach(val => updateUrlSearchParamsList(val, key, history));
```
ReeceHoffmann: Why not just pass the whole array? Seems like it would simplify the logic quite a bit and be faster

src/tests/fake/sample.tsx
```diff
@@ -0,0 +1,39 @@
+import nock from "nock";
+
+export function mockCreateSample() {
```
ReeceHoffmann: Does the name match what is being mocked?

src/tests/fake/sample.tsx
```diff
@@ -0,0 +1,39 @@
...
+                    name: "sample2",
+                    created_at: "2023-11-06T18:31:46.469000Z",
+                    host: "",
+                    isolate: "",
+                    labels: [],
+                    library_type: "srna",
+                    ready: true,
+                    user: { handle: "testUser" },
+                    workflows: { aodp: "incompatible", nuvs: "none", pathoscope: "none" },
+                },
```
ReeceHoffmann: Mocked documents should specified when calling. Look at the other mock functions for examples

ReeceHoffmann (changes requested):
Mostly small stuff
 re query naming: I am thinking for consistency sake getting a list of anything should be `useListX` whereas getting single should be `useFetchX`. Open to other ideas as well

src/js/analyses/components/Create/QuickAnalyze.tsx
```diff
@@ -139,16 +227,21 @@ export function QuickAnalyze({
...
                                 as={IndexSelector}
-                                indexes={compatibleIndexes}
+                                indexes={indexes}
                                 selected={values.indexes}
                                 onChange={value => setFieldValue("indexes", value)}
                             />
-                            <QuickAnalyzeError>{touched.indexes && errors.indexes}</QuickAnalyzeError>{" "}
+                            <QuickAnalyzeError>
+                                {touched.indexes && (errors.indexes as ReactNode & (string | string[]))}
+                            </QuickAnalyzeError>{" "}
```
ReeceHoffmann:
```suggestion
                            </QuickAnalyzeError>
```
Unless it is doing something

src/js/samples/api.ts
```diff
@@ -46,3 +46,26 @@ export const update = ({ sampleId, update }) => Request.patch(`/samples/${sample
...
+export function listSamples(page: number, per_page: number, term: string, labels: string[], workflows: string) {
+    const request = Request.get("/samples").query({ page, per_page, find: term });
+
+    if (labels) {
+        labels.forEach(label => request.query({ label }));
+    }
+
+    if (workflows) {
+        request.query({ workflows });
+    }
```
ReeceHoffmann: Do these need to be separate? I think superagent supports being passed an array of values, and undefined parameters should be ignored.

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const { data: labels, isLoading: isLabelsLoading } = useFetchLabels();
+
+    const [selected, setSelected] = useState([]);
+
+    useEffect(() => {
+        setSelected(
+            selected.map(
+                prevSample => samples.documents.find(newSample => newSample.id === prevSample.id) || prevSample,
+            ),
+        );
```
ReeceHoffmann: Why store the whole sample instead of just the samples id? Save having to write updaters like this. Should also let you simplify other logic working with the list

src/js/samples/components/SamplesList.tsx
```diff
@@ -0,0 +1,129 @@
...
+                </SamplesListContent>
+                {selected.length ? (
+                    <SampleLabels labels={labels} selectedSamples={selected} />
+                ) : (
+                    <SampleFilters
+                        labels={labels}
+                        onClick={e => {
+                            !filterLabels.includes(e.toString())
+                                ? setFilterLabels(union(filterLabels, [e]))
+                                : setFilterLabels(without(filterLabels, e.toString()));
```
ReeceHoffmann: Is `xor` an option here?

src/js/samples/components/SamplesToolbar.tsx
```diff
@@ -0,0 +1,34 @@
...
+            )}
+        </Toolbar>
+    );
+}
+
+export default function SampleToolbar({ selected, onClear, onChange, term }) {
+    const history = useHistory();
+    function onQuickAnalyze() {
+        history.push({ state: { quickAnalysis: true } });
+    }
```
ReeceHoffmann: Looks like this is only used by `SampleSelectionToolbar` would define it there instead

src/js/samples/components/__tests__/SamplesList.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { describe, expect, it } from "vitest";
+import { createFakeAccount, mockGetAccountAPI } from "../../../../tests/fake/account";
+import { createFakeSampleMinimal, mockApiGetSamples } from "../../../../tests/fake/sample";
+import { renderWithRouter } from "../../../../tests/setupTests";
+import { AdministratorRoles } from "../../../administration/types";
+import SamplesList from "../SamplesList";
+
+describe("<SamplesList />", () => {
+    const history = createBrowserHistory();
+    const samples = [createFakeSampleMinimal(), createFakeSampleMinimal()];
```
ReeceHoffmann: Careful with this. All tests will share the same variables when running the test suite. This can result in subtle bugs where running as a single test it works but when they are run in a group tests start to fail since the previous tests modified the shared variables. 

Even though deepsource will complain, i generally would just initialise with no value and assign in `beforeEach` to eliminate this problem

src/js/samples/querys.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Fetch a page of samples from the API
+ *
+ * @param page - The page to fetch
+ * @param per_page - The number of samples to fetch per page
+ * @param term - The search term to filter samples by
+ * @param labels - The labels to filter the samples by
+ */
+export function useListSamples(page: number, per_page: number, term?: string, labels?: string[]) {
+    const params = new URLSearchParams(window.location.search);
+    const workflows = params.get("workflows");
```
ReeceHoffmann: Should probably be passed as a regular param

src/js/samples/querys.ts
```diff
@@ -0,0 +1,82 @@
...
+        onSuccess: () => {
+            queryClient.invalidateQueries(samplesQueryKeys.lists());
+        },
+    });
+
+    function onUpdate(label: number) {
+        forEach(selectedSamples, sample => {
+            const sampleLabelIds = map(sample.labels, label => label.id);
+            const labelExists = selectedLabels.some(item => item.id === label);
+            const allLabeled = selectedLabels.every(item => item.allLabeled === true);
```
ReeceHoffmann, line 57: If `SelectedLabels` is supposed to be `Labels[]` then where is `item.allLabeled` coming from. Somethings funky here.

ReeceHoffmann (changes requested):
Overall I think this is a step in the right direction for the codebase. I have two general things I want improved, but they may be more suited to another PR as I suspect they are holdovers from previous implementations:

1. When collecting a list of something (EX. SelectedSamples) I generally prefer to store the ids, then use those ids to lookup the original when needed. Prevents needing to write strange updaters

2. A lot of the logic surrounding labels is pretty tough to grok (EX: useUrlSearchParamsList and `getSelectedLabels`). I think it is a multifaceted typing/logic problem.

Take a look to see if you easily clean these up a bit, otherwise I think it is reasonable to fix these in another PR.

src/js/analyses/components/Create/QuickAnalyze.tsx
```diff
@@ -62,42 +60,130 @@ const validationSchema = Yup.object().shape({
...
-}) {
+/**
+ * Gets the quick analysis mode
+ *
+ * @param libraryType - The library type of the sample
+ * @param history - The history object
+ * @returns The quick analysis mode
+ */
+export function getQuickAnalysisMode(libraryType: string, history: HistoryType) {
+    if (history.location.state?.["quickAnalysis"] === true) {
```
ReeceHoffmann:
```suggestion
    if (history.location.state?.quickAnalysis === true) {
```
?

src/js/samples/components/Filter/SampleFilters.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Label } from "../../../labels/types";
+import LabelFilter from "./LabelFilter";
+import WorkflowFilter from "./WorkflowFilter";
+
+const StyledSampleFilters = styled.div`
+    grid-column: 2;
+    grid-row: 2;
+`;
+
+type LabelFilterProps = {
```
ReeceHoffmann:
```suggestion
type SampleFilterProps = {
```

src/js/samples/querys.ts
```diff
@@ -0,0 +1,80 @@
...
+    color: string;
+    /** The detailed description */
+    description: string;
+    /** The unique identifier */
+    id: number;
+    /** The display name */
+    name: string;
+    /** Whether all selected samples contain the label */
+    allLabeled: boolean;
+};
```
ReeceHoffmann:
```suggestion
type SampleLabel =Label & {
    allLabeled: boolean;
};
```
My suggested name isn't amazing, but makes a bit more clear it is not a standard label.

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

ReeceHoffmann (changes requested):
Noticed bugs in manual testing after initial review:

1. Label filtering doesn't work
2. We are getting multistage loading of the main page

ReeceHoffmann (dismissed):

ReeceHoffmann:
Resolve the conflicts and I am good to merge

ReeceHoffmann (approved):

PR #284, opened 2023-11-09
refactor: paginate files in sample creation view

raastinn:
Screenshot from 2023-11-09 12-01-13

Basically done, just need to add typing to the touched components


src/js/base/InputSelect.tsx
```diff
@@ -10,7 +10,7 @@ type InputSelectProps = {
 
 export const InputSelect = ({ children, name, value, onChange }: InputSelectProps) => {
     return (
-        <Input as="select" name={name} value={value} onChange={onChange}>
+        <Input as="select" name={name} value={value || ""} onChange={onChange}>
```
ReeceHoffmann: I would be reluctant to do this as it forces the Input to always be controlled. May break other components that are intentionally not passing a value

src/js/files/querys.ts
```diff
@@ -1,11 +1,13 @@
...
 import { Request } from "../app/request";
-import { FileType } from "./types";
+import { FileResponse, FileType } from "./types";
 
 export const fileQueryKeys = {
     all: () => ["files"] as const,
     lists: () => ["files", "list"] as const,
     list: (type: string, filters: Array<string | number | boolean>) => ["files", "list", type, ...filters] as const,
+    infiniteList: (type: string, filters: Array<string | number | boolean>) =>
+        ["files", "infiniteList", type, ...filters] as const,
```
ReeceHoffmann: You mighta duped my bad code here... Should be:

```suggestion
    infiniteList: (type: string, filters: Array<string | number | boolean>) =>
        ["files", "list", "infinite" type, ...filters] as const,
```
That way you can reset all types of list at once say when WS updates come in without having to worry about list type

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -104,16 +105,18 @@ const AlertContainer = styled.div`
...
-    const subtractionIds = intersectionWith(
-        values.sidebar.subtractionIds,
-        subtractions,
-        (subtractionId, subtraction) => subtractionId === subtraction.id,
-    );
-    return { ...values, readFiles, sidebar: { labels, subtractionIds } };
-};
+function castValues(subtractions, allLabels) {
+    return function (values) {
+        // const readFiles = intersectionWith(values.readFiles, reads, (readFile, read) => readFile === read.id);
```
ReeceHoffmann:
```suggestion
```
Comments like these + `console.log`s need to be removed before merging

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -126,50 +129,68 @@ type formValues = {
...
+    const { data: allLabels, isLoading: labelsLoading } = useFetchLabels();
+    const { data: groups, isLoading: groupsLoading } = useListGroups();
+    const { data: subtractions, isLoading: subtractionsLoading } = useSubtractionsShortlist();
+    const { data: settings, isLoading: settingsLoading } = useFetchSettings();
+    const {
+        data: readsResponse,
+        isLoading: isReadsLoading,
+        isFetchingNextPage,
+        fetchNextPage,
+    } = useInfiniteFindFiles(FileType.reads, 11);
```
ReeceHoffmann: Why 11 in particular?

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -126,50 +129,68 @@ type formValues = {
...
+        libraryType: "normal",
+        readFiles: [],
+        group: forceGroupChoice ? "none" : null,
+        sidebar: { labels: [], subtractionIds: [] },
+    };
+}
 
-    if (subtractions === null || allLabels === null || isReadsLoading) {
+export default function CreateSample() {
+    const { data: allLabels, isLoading: labelsLoading } = useFetchLabels();
```
ReeceHoffmann: Nitpick: keeps the `is` for booleans convention and sounds more natural imo
```suggestion
    const { data: allLabels, isLoading: isLoadingLabels } = useFetchLabels();
```

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -126,50 +129,68 @@ type formValues = {
...
+            history.push("/samples");
+        },
+    });
+
+    if (isReadsLoading || labelsLoading || subtractionsLoading || settingsLoading || groupsLoading) {
         return <LoadingPlaceholder margin="36px" />;
     }
 
-    const reads = filter(readsResponse.documents, { reserved: false });
+    function onCreate(name, isolate, host, locale, libraryType, subtractionIds, files, labels, group?) {
```
ReeceHoffmann:
```suggestion
    function onCreate(name, isolate, host, locale, libraryType, subtractionIds, files, labels, group) {
```
Why the `?` ?

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -126,50 +129,68 @@ type formValues = {
...
+            files,
+            labels,
+            group,
+        });
+    }
 
-    const autofill = (selected, setFieldValue) => {
+    const forceGroupChoice = settings.sample_group === "force_choice";
+
+    const reads = readsResponse.pages.flatMap(page => page.items, { reserved: false });
```
ReeceHoffmann:
```suggestion
    const reads = readsResponse.pages.flatMap(page => page.items, { reserved: false });
```

Is `{ reserved: false }` doing anything here? Also a plug for lodash, all of these default functions, often with conveniences and many more on top. https://lodash.com/docs/4.17.15#flatMap

src/js/samples/components/Create/ReadSelector.tsx
```diff
@@ -0,0 +1,161 @@
...
+    data: InfiniteData<FileResponse>;
+    isFetchingNextPage: boolean;
+    fetchNextPage: (options?: FetchNextPageOptions) => Promise<InfiniteQueryObserverResult>;
+    isLoading: boolean;
+    onSelect: (selected: number[]) => void;
+    error: string;
+    selected: number[];
+};
+
+let selectedFiles = [];
```
ReeceHoffmann: Mutable global variables are extremely dicey. Every instance of this function now shares the exact same list of selected files. Define it inside the function

ReeceHoffmann (changes requested):
Good work so far! In addition to the other comments please also fix the styling on the read selector, looks like some of that got lost during the update

image

raastinn:
Screenshot from 2023-11-16 12-34-29
Updated styling

raastinn:
Screenshot from 2023-12-21 14-20-45
Updated styling

src/js/base/ScrollList.tsx
```diff
@@ -15,6 +20,54 @@ const StyledScrollList = styled.div`
...
+ *
+ * @param className - The class name used for the scroll list
+ * @param fetchNextPage - A function which initiates fetching the next page
+ * @param isFetchingNextPage - Whether a new page is being fetched
+ * @param isLoading - Whether the first page is being fetched
+ * @param items - The list of items
+ * @param renderRow - A function which accepts an item and returns a react element
+ * @returns An infinitely scrolling list of items
+ */
+export function ScrollListElement({
```
ReeceHoffmann: What about `CompactScrollList`?

src/js/base/ScrollList.tsx
```diff
@@ -15,6 +20,54 @@ const StyledScrollList = styled.div`
...
+    className?: string;
+    fetchNextPage: (options?: FetchNextPageOptions) => Promise<InfiniteQueryObserverResult>;
+    isFetchingNextPage: boolean;
+    isLoading: boolean;
+    items: unknown[];
+    renderRow: (item: unknown) => void;
+};
+
+/**
+ * An infinitely scrolling list of items within an element
```
ReeceHoffmann: I feel like the docstring should more clearly highlight that this is a compact height constrained scrollList. `Element` could be very large

src/js/base/ScrollList.tsx
```diff
@@ -15,6 +20,54 @@ const StyledScrollList = styled.div`
...
+export function ScrollListElement({
+    className,
+    fetchNextPage,
+    isFetchingNextPage,
+    isLoading,
+    items,
+    renderRow,
+}: ScrollListElementProps) {
+    function onScroll() {
+        const scrollListElement = document.getElementById("scroll-list");
```
```suggestion
    function onScroll(e) {
        const scrollListElement = e.target
```
or similar should work. Save having to rely on the id, which could quickly become a problem if more than one scrollList exists in a a single page

src/js/base/ScrollList.tsx
```diff
@@ -15,6 +20,54 @@ const StyledScrollList = styled.div`
...
+        if (getScrollRatio(scrollListElement) > 0.8 && !isFetchingNextPage) {
+            void fetchNextPage();
+        }
+
+        scrollListElement.addEventListener("scroll", onScroll);
+        scrollListElement.removeEventListener("scroll", onScroll);
```
ReeceHoffmann: What's going on here? Seems like this shouldn't be doing anything since it add then instantly remove the listener

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -186,14 +214,17 @@ export function CreateSample({
                 group === "none" ? "" : group,
             );
         } else {
-            onCreate(name, isolate, host, locale, libraryType, subtractionIds, readFiles, labels);
+            onCreate(name, isolate, host, locale, libraryType, subtractionIds, readFiles, labels, group);
```
ReeceHoffmann: Currently sending group is ignored by the backend unless `forceGroupChoice` is sent. That said I am not against moving towards supporting it in the FE. That said couple of notes:

1. the value of group sent to the backend must be either and int, null, or not sent. 
2. The default value in the form for group should be the current users `primary_group` or `None` iff the user has no `primary_group`

With those two changes you can probably condense the two cases in handleSubmit to a single case

src/js/samples/components/Sidebar/Subtractions.jsx
```diff
@@ -21,34 +20,36 @@ const SampleSubtractionFooter = styled.div`
...
-        />
-        {Boolean(subtractionOptions.length) || (
-            <SampleSubtractionFooter>
-                No subtractions found. <Link to="/subtractions">Create one</Link>.
-            </SampleSubtractionFooter>
-        )}
-    </SideBarSection>
-);
+export default function DefaultSubtractions({ defaultSubtractions, onUpdate }) {
+    const { data: subtractionOptions, isLoading } = useSubtractionsShortlist();
```
ReeceHoffmann:
```suggestion
    const { data: subtractionOptions, isLoading } = useFetchSubtractionsShortlist();
```

src/js/subtraction/api.js
```diff
@@ -25,3 +25,7 @@ export function findSubtractions({ page, per_page, term }) {
             return response.body;
         });
 }
+
+export function subtractionShortlist() {
```
ReeceHoffmann:
```suggestion
export function fetchSubtractionShortlist() {
```

src/js/subtraction/querys.ts
```diff
@@ -15,3 +16,13 @@ export function useFindSubtractions(page: number, per_page: number, term: string
         },
     );
 }
+
+type SubtractionShortlist = {
+    name: string;
+    id: string;
+    ready: boolean;
+};
```
ReeceHoffmann: Not an official server type, but should probably be moved to subtraction types

src/js/subtraction/querys.ts
```diff
@@ -15,3 +16,13 @@ export function useFindSubtractions(page: number, per_page: number, term: string
...
+
+export function useSubtractionsShortlist() {
```
ReeceHoffmann:
```suggestion
export function useFetchSubtractionsShortlist() {
```

src/tests/fake/subtractions.ts
```diff
@@ -77,6 +77,14 @@ export function createFakeSubtractionMinimal(overrides?: CreateFakeSubtractionMi
     return assign(defaultSubtractionMinimal, overrides);
 }
 
+export function createFakeSubtractionShortlist() {
```
ReeceHoffmann: Hmm naming seems off here. A subtraction shortList should be a list of these reduced subtraction objects

ReeceHoffmann (changes requested):

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -126,78 +132,98 @@ type formValues = {
...
+            name,
+            isolate,
+            host,
+            locale,
+            libraryType,
+            subtractionIds,
+            readFiles,
+            labels,
+            group === "none" ? "" : group,
+        );
```
ReeceHoffmann: Why is `onCreate` it's own function? Seems like it essentially just passes through to the mutation

src/js/subtraction/querys.ts
```diff
@@ -1,9 +1,10 @@
 import { useQuery } from "react-query";
-import { findSubtractions } from "./api";
-import { SubtractionSearchResult } from "./types";
+import { fetchSubtractionShortlist, findSubtractions } from "./api";
+import { SubtractionSearchResult, SubtractionShortlist } from "./types";
 
 export const subtractionQueryKeys = {
     list: (filters: Array<string | number | boolean>) => ["subtraction", "list", ...filters] as const,
+    shortlist: () => ["subtraction", "shortlist"] as const,
```
ReeceHoffmann:
```suggestion
    shortlist: () => ["subtraction", "list", "short"] as const,
```
I have been moving all lists over to being some variant of `["subtraction", "list", "OtherKey"]` so that invalidation targeting lists can simply invalidate `Thing.List` to invalidate all the lists. For now we are using all instead, but I am hoping to make it more selective down the line.

src/tests/fake/samples.tsx
```diff
@@ -0,0 +1,104 @@
...
+        ...sampleMinimal,
+        locale: faker.random.word(),
+        subtractions: faker.datatype.number(),
+        files: files || [faker.datatype.number(), faker.datatype.number()],
+        labels: [],
+        group: null,
+    };
+
+    return assign(defaultCreateSample, overrides);
+}
```
ReeceHoffmann: Assignment the docstring is on point this might be too specific for the mock file. The specific object for creating a sample should only be used in the sample creation view, so I am not sure the benefit of having to available to all tests.  If it is needed in multiple components happy to leave it here, but it should probably be renamed. Current naming scheme is exactly what I would expect a function that returns a complete `Sample` to return

src/tests/fake/subtractions.ts
```diff
@@ -77,6 +79,13 @@ export function createFakeSubtractionMinimal(overrides?: CreateFakeSubtractionMi
     return assign(defaultSubtractionMinimal, overrides);
 }
 
+/**
+ * Create a fake subtraction shortlist
+ */
+export function createFakeReducedSubtraction(): SubtractionShortlist {
+    return pick(createFakeSubtractionMinimal(), ["id", "name", "ready"]);
+}
```
ReeceHoffmann:
```suggestion
export function createFakeShortlistSubtraction(): SubtractionShortlist {
    return pick(createFakeSubtractionMinimal(), ["id", "name", "ready"]);
}
```
? Not stuck on it, since we have a `ShortlistSubtraction` type now this tracks for me

ReeceHoffmann (changes requested):
Looks really good. Have a few small naming things that I think could be confusing, but I expect to merge next round

ReeceHoffmann (approved):

PR #285, opened 2023-11-16
refactor: add pagination to subtraction creation view

raastinn:


raastinn:
Screenshot from 2023-11-16 15-28-00

raastinn:
My tests pass locally not sure why they are failing here

src/js/subtraction/components/SubtractionFileSelector.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    ${Attribution} {
+        margin-left: auto;
+    }
+`;
+
+export function SubtractionFileItem({ active, onClick, name, uploaded_at, user, id, error }) {
+    function handleClick() {
+        onClick(id);
+    }
```
ReeceHoffmann: Makes more sense as an anonymous function imo

src/js/subtraction/components/SubtractionFileSelector.tsx
```diff
@@ -0,0 +1,114 @@
...
+                error={error}
+            />
+        );
+    }
+
+    useEffect(() => {
+        if (selected) {
+            onClick("");
+        }
+    }, [foundCount]);
```
ReeceHoffmann: What is this doing?

src/js/subtraction/components/CreateSubtraction.tsx
```diff
@@ -13,48 +14,87 @@ import {
...
-    const uploadId = find(files, ["id", values.uploadId]) ? values.uploadId : "";
-    return { ...values, uploadId };
+function castValues(files) {
+    files.forEach(item => {
+        return function (values) {
+            const uploadId = find(item, ["id", values.uploadId]) ? values.uploadId : "";
+            return { ...values, uploadId };
+        };
+    });
+}
```
ReeceHoffmann: This function work? Pretty sure forEach throws away any returned values

src/js/files/querys.ts
```diff
@@ -20,6 +22,27 @@ export function useListFiles(type: FileType, paginate: boolean, page = 1) {
     });
 }
 
+function findFiles(type: FileType, page: number, per_page: number, term?: string) {
+    return Request.get("/uploads")
+        .query({ upload_type: type, page, per_page, ready: true, paginate: true, find: term })
+        .then(response => response.body);
+}
```
ReeceHoffmann: api.ts

src/js/subtraction/components/SubtractionFileSelector.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+const StyledSubtractionFileItem = styled(SelectBoxGroupSection)<StyledSubtractionFileItemProps>`
+    display: flex;
+
+    ${Attribution} {
+        margin-left: auto;
+    }
+`;
+
+export function SubtractionFileItem({ active, onClick, name, uploaded_at, user, id, error }) {
```
ReeceHoffmann: 1 export per file works better with vite

ReeceHoffmann (changes requested):

ReeceHoffmann:
> My tests pass locally not sure why they are failing here

Reasons can vary, but in this case it is test flake.

src/js/subtraction/components/SubtractionFileSelector.tsx
raastinn: This clears the selected file if the selected subtraction file is deleted

src/js/subtraction/components/CreateSubtraction.tsx
raastinn: This allows us to preserve the selected value if it isn't on the first page. It goes through all the pages to preserve the selected file by its uploadId.

src/js/subtraction/components/SubtractionFileSelector.tsx
ReeceHoffmann: What happens if the number of files changes but the file still exists?

raastinn:
Screenshot from 2023-11-21 14-01-37
Changed styling accordingly

src/js/base/ScrollList.tsx
```diff
@@ -31,19 +38,29 @@ type ScrollListProps = {
...
+export const ScrollList = ({
+    fetchNextPage,
+    isFetchingNextPage,
+    isLoading,
+    items,
+    renderRow,
+    elementId,
+}: ScrollListProps) => {
     useEffect(() => {
+        const scrollListElement = elementId ? document.getElementById(elementId) : window;
```
ReeceHoffmann: This works but I don't think having to pass the outer component id is very intuitive as an API. Can you apply the overflow property on the scrollist and using the scrollist itself as the element? Might even make the height determination calculation easier.

src/js/subtraction/components/SubtractionFileSelector.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const {
+        data: subtractions,
+        isLoading: isLoadingSubtractions,
+        fetchNextPage: fetchNextSubtractionsPage,
+        hasNextPage,
+    } = useInfiniteFindFiles(FileType.subtraction, 25);
+
+    useEffect(() => {
+        if (!isLoadingSubtractions && selected) {
+            const allSubtractions = getAllSubtractions(fetchNextSubtractionsPage, hasNextPage, subtractions);
```
ReeceHoffmann: Naming and signature of this function is misleading imo. Since the underlying really just returns the data it has already been passed.

Since you'll need something similar in the samples view it might be worth considering a custom hook along the lines of `useValidateFiles` that is capable of generically handling out of date fiel selections. Not required for merge though

src/js/subtraction/components/CreateSubtraction.tsx
```diff
@@ -13,50 +13,78 @@ import {
...
         nickname: "",
         uploadId: "",
     };
 
-    const handleSubmit = ({ uploadId, name, nickname }) => {
-        onCreate({ uploadId, name, nickname });
-    };
+    function handleSubmit({ uploadId, name, nickname }) {
+        onCreate(uploadId, name, nickname);
+    }
```
ReeceHoffmann: Why have onCreate and handleSubmit as distinct functions?

ReeceHoffmann (changes requested):

src/js/base/ScrollList.tsx
```diff
@@ -34,22 +47,31 @@ type ScrollListProps = {
...
-        return () => window.removeEventListener("scroll", onScroll);
+        scrollListElement.addEventListener("scroll", onScroll);
+        return () => scrollListElement.removeEventListener("scroll", onScroll);
     }, [isFetchingNextPage, fetchNextPage]);
 
     const entries = map(items, item => renderRow(item));
 
     return (
-        <StyledScrollList>
+        <StyledScrollList id={"scroll-list"} className={className}>
```
```suggestion
        <StyledScrollList id={"scroll-list"} className={className} onScroll={onScroll}>
```
I think this should do the same thing as your id based solution in a more reacty way. Worth checking that this soln doesn't mess with implementations where there is no height limit. (if the user scroll the whole window it might not get picked up depending on cursor position)

src/js/base/ScrollList.tsx
```diff
@@ -5,17 +5,29 @@ import styled from "styled-components";
 import { usePrevious } from "./hooks";
 import { LoadingPlaceholder } from "./LoadingPlaceholder";
 
-function getScrollRatio(): number {
+function getScrollRatio(scrollListElement?: HTMLElement | Window): number {
+    if (scrollListElement !== window) {
```
```suggestion
function getScrollRatio(scrollListElement?: HTMLElement ): number {
    if (boolean(scrollListElement)) {
```
Unless there is a case where you expect to pass window

src/js/utils/hooks.ts
```diff
@@ -91,3 +94,39 @@ export function useUrlSearchParams(key: string, defaultValue?: string): [string,
...
+ * @param selected - The selected file id
+ */
+export function useValidateFiles(
+    data: InfiniteData<FileResponse>,
+    fetchNextPage: (options?: FetchNextPageOptions) => Promise<InfiniteQueryObserverResult>,
+    hasNextPage: boolean,
+    isLoading: boolean,
+    onClick: (file: string) => void,
+    selected: string,
+) {
```
ReeceHoffmann: I would redesign the hook around it fetching the list of files for itself. Doing it this way ensures that you get a full list of files even if the user has applied a filter and if there is no filter applied then react-query's caching should mean that the request results are re-used between both component automagically.

signature would look along the lines of:

```suggestion
export function useValidateFiles(
    type: string
    selected: string[]
    setSelected: (selected: string[] => void)
) {
```

src/js/utils/hooks.ts
```diff
@@ -91,3 +94,39 @@ export function useUrlSearchParams(key: string, defaultValue?: string): [string,
...
+ * Hook for validating selected files from paginated data
+ *
+ * @param data - The data files
+ * @param fetchNextPage - Fetches the next page of data
+ * @param hasNextPage - Whether the search result has a next page
+ * @param isLoading - Whether the data is loading
+ * @param onClick - A callback function to handle file selection
+ * @param selected - The selected file id
+ */
+export function useValidateFiles(
```
ReeceHoffmann: Move to `hooks.ts` in`/files` since this is so resource specific.

ReeceHoffmann (changes requested):

src/js/files/hooks.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param setSelected - A callback function to handle file selection
+ */
+export function useValidateFiles(type: FileType, selected: string[], setSelected: (selected: string[]) => void) {
+    const { data, isLoading, fetchNextPage, hasNextPage } = useInfiniteFindFiles(type, 25);
+
+    function getData() {
+        if (hasNextPage) {
+            void fetchNextPage();
+        }
+        return data.pages.flatMap(page => page.items);
```
ReeceHoffmann: Why is this a separate function?

src/js/subtraction/components/CreateSubtraction.tsx
```diff
@@ -13,50 +13,74 @@ import {
...
+    uploadId: Yup.array().min(1, "Please select a file"),
 });
 
-const castValues = files => values => {
-    const uploadId = find(files, ["id", values.uploadId]) ? values.uploadId : "";
-    return { ...values, uploadId };
+type formValues = {
+    name: "";
+    nickname: "";
+    uploadId: string[];
```
ReeceHoffmann, line 30: I would probably keep it as `string` and adjust the parameters you pass to the hook.

src/js/subtraction/components/CreateSubtraction.tsx
```diff
@@ -13,50 +13,74 @@ import {
...
+    nickname: "",
+    uploadId: [],
+};
+
+/**
+ * A form for creating a subtraction
+ */
+export default function CreateSubtraction() {
+    const {
+        data: filesResponse,
```
ReeceHoffmann:
```suggestion
        data: files,
```
?

ReeceHoffmann (changes requested):
Looking close, have a couple things I am uncertain about with `useValidateFiles` though.

1. I noticed that it retries validation on every update to data, seems like there is real potential for major API spam here. Can check this and make sure it is within reason
2. As written I think this will always load all pages, but if we find the file in the first page then realistically we could stop there.

ReeceHoffmann (approved):

ReeceHoffmann (changes requested):
Code looks good but look like there is a css problem for the last item in the list: 
image

src/js/base/ScrollList.tsx
```diff
@@ -103,10 +118,12 @@ export const LegacyScrollList = ({ page, documents, pageCount, onLoadNextPage, r
     }
 
     return (
-        <StyledScrollList>
-            {entries}
-            {loading}
-        </StyledScrollList>
+        <ThemeProvider theme={theme}>
```
ReeceHoffmann: Were you having problems without `ThemeProvider`? You shouldn't need it here

src/js/subtraction/components/SubtractionFileSelector.tsx
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    return foundCount === 0 ? (
+        <NoneFoundBox noun="files">
+            <Link to="/subtractions/files">Upload some</Link>
+        </NoneFoundBox>
+    ) : (
+        <>
+            <ScrollList
+                className="border"
```
ReeceHoffmann: https://styled-components.com/docs/basics#styling-any-component

ReeceHoffmann (changes requested):

src/js/base/ScrollList.tsx
raastinn: Yeah since I added the border class, but with the styled method you sent it doesn't give this error anymore. Removed the `ThemeProvider` in the latest push

ReeceHoffmann (approved):

PR #286, opened 2023-11-23
refactor: add pagination to hmm list view

raastinn:


src/js/hmm/api.ts
```diff
@@ -34,3 +34,17 @@ export function install(): Promise<any> {
...
+
+/**
+ * Fetch a page of hmm search results
+ *
+ * @param page - The page to fetch
+ * @param per_page -The number of hmms to fetch per page
+ * @param term - The search term to filter the hmms by
+ * @returns A promise resolving to a page of hmm search results
+ */
+export function listHmms(page: number, per_page: number, term: string) {
```
ReeceHoffmann:
```suggestion
export function listHmms(page: number, per_page: number, term: string): Promise<HMMSearchResult> {
```

src/js/hmm/components/HMMInstaller.tsx
```diff
@@ -38,14 +37,25 @@ const StyledHMMInstaller = styled(Box)`
...
+export default function HMMInstaller() {
+    const { data, isLoading } = useListHmms(1, 25);
+    const queryClient = useQueryClient();
+
+    if (isLoading) {
+        return <LoadingPlaceholder />;
+    }
+
+    const { status } = data;
+    const { task } = status;
```
ReeceHoffmann:
```suggestion
    const { status: { task, installed} } = data;
```

src/js/hmm/components/HMMInstaller.tsx
```diff
@@ -38,14 +37,25 @@ const StyledHMMInstaller = styled(Box)`
...
+
+    if (task && !status.installed) {
         const progress = task.progress;
         const step = replace(task.step, "_", " ");
 
-        if (task.complete) {
-            onComplete();
-        }
+        queryClient.invalidateQueries(hmmQueryKeys.lists());
```
ReeceHoffmann: Does this result in api request spam from the list being invalidated repeatedly?

src/js/hmm/components/HMMList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    if (isLoading) {
+        return <LoadingPlaceholder />;
+    }
+
+    const { documents, page, page_count, found_count, total_count } = data;
+
+    function renderRow(document: HMMMinimal) {
+        return <HMMItem key={document.id} hmm={document} />;
+    }
```
ReeceHoffmann: Nitpick: Looks like a pure component so you can define this outside of the component to save redefining it on each render

src/js/hmm/components/HMMList.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    return (
+        <div>
+            <ViewHeader title="HMMs">
+                <ViewHeaderTitle>HMMs {total_count > 0 && <Badge>{found_count}</Badge>}</ViewHeaderTitle>
+            </ViewHeader>
+
+            {total_count ? (
```
ReeceHoffmann, line 34: I would use status instead. should be packaged into the response from your existing hmm query already. Shouldn't be the case, but technically you could have a valid install of 0 HMMs

src/tests/fake/hmm.ts
```diff
@@ -0,0 +1,75 @@
...
+        total_count: 0,
+    };
+}
+
+/**
+ * Create a fake HMM search result
+ *
+ * @param hasDocument - Whether the search result has documents
+ */
+export function createFakeHMMSearchResults(hasDocument = true): HMMSearchResults {
```
ReeceHoffmann: For faking functions I like to allow overriding of any top level object generated by the function so that the faker can be used to test any UI state.

So for example the user can pass a value for `documents` which fully overrides the underlying.

src/tests/fake/hmm.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    const documents = times(5, () => ({
+        cluster: faker.datatype.number(),
+        count: faker.datatype.number(),
+        families: {
+            None: faker.datatype.number(),
+            Papillomaviridae: faker.datatype.number(),
+        },
+        id: faker.random.alphaNumeric(9, { casing: "lower" }),
+        names: [faker.name.lastName()],
```
ReeceHoffmann: I would factor this out into a separate faker function. Then just call the new function x times

src/tests/fake/hmm.ts
```diff
@@ -0,0 +1,75 @@
...
+    };
+}
+
+/**
+ * Sets up a mocked API route for fetching a list of HMMs
+ *
+ * @param hmm - The hmm search results to be returned from the mocked API call
+ * @returns The nock scope for the mocked API call
+ */
+export function mockApiGetHmms(hmm: HMMSearchResults) {
```
ReeceHoffmann:
```suggestion
export function mockApiGetHmms(hmmSearchResults: HMMSearchResults) {
```

ReeceHoffmann (changes requested):

src/js/hmm/components/__tests__/HMMList.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+        const scope = mockApiGetHmms(fakeHMMData);
+        renderWithRouter(<HMMList />, {}, history);
+
+        expect(await screen.findByText("HMMs")).toBeInTheDocument();
+        expect(screen.getByPlaceholderText("Definition")).toBeInTheDocument();
+
+        expect(screen.getByText(fakeHMMData.documents[0].cluster)).toBeInTheDocument();
+        expect(screen.getByText(fakeHMMData.documents[0].names[0])).toBeInTheDocument();
+
+        scope.isDone();
```
ReeceHoffmann:
```suggestion
        scope.done();
```
? I think `scope.isDone()` just returns a bool, whereas `scope.done()` should fail the test if not mocks haven't been called

src/tests/fake/hmm.ts
```diff
@@ -0,0 +1,68 @@
...
+        },
+        task: {
+            complete: true,
+        },
+    };
+
+    const { documents, status, total_count } = overrides || {};
+
+    const defaultHMMSearchResult = {
+        documents: documents || times(5, () => createFakeHMMMinimal()),
```
ReeceHoffmann: Careful with using `||` statements here. Falsey overrides will erroneously  be ignored. (ex. You want to return `total_count==0`) If you know the keys between the overrides and your dict are identical I would just let merge combine the two and save the headache of this bug

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #287, opened 2023-11-27
chore: update user detailed view

raastinn:
I believe that the group list is needed since it handles the selection of the groups associated with the user.


src/js/administration/querys.ts
```diff
@@ -87,6 +87,30 @@ export const useInfiniteFindUsers = (per_page: number, term: string, administrat
...
 
+/**
+ * Fetches a single user
+ *
+ * @param userId - The id of the user to fetch
+ * @returns A single user
+ */
+export function useGetUser(userId: string) {
+    return useQuery<User>(userQueryKeys.detail(userId), () => getUser(userId));
+}
```
ReeceHoffmann:
```suggestion
export function useFetchUser(userId: string) {
    return useQuery<User>(userQueryKeys.detail(userId), () => getUser(userId));
}
```

src/js/users/components/UserPermissions.tsx
```diff
@@ -17,28 +16,30 @@ const UserPermissionsHeader = styled.div`
         margin-left: auto;
     }
 `;
-export const UserPermissions = ({ permissions }) => {
```
ReeceHoffmann, line 20: Just an FYI for the future, you can also just remove the `export` and that's okay for vite. I think this component work either way so no need to change it back.

src/tests/fake/user.tsx
```diff
@@ -101,3 +102,18 @@ export function mockApiFindUsers(users: Array<User>, query?: Query) {
...
+ *
+ * @param userId - The users unique id
+ * @param userDetail - The details of the user
+ * @returns A nock scope for the mocked API call
+ */
+export function mockApiGetUser(userId: string, userDetail: User) {
+    return nock("http://localhost").get(`/api/users/${userId}`).reply(200, userDetail);
+}
+
+export function mockApiEditUser(userId: string, statisCode: number, update: any) {
```
ReeceHoffmann:
```suggestion
export function mockApiEditUser(userId: string, statusCode: number, update: any) {
```
? Also I think the return type on editing a user might be UserDetailed. We don't leverage that in the UI, but migt as well reflect what the API does

src/js/users/components/__tests__/UserDetail.test.tsx
```diff
@@ -0,0 +1,280 @@
...
+            renderWithRouter(<UserDetail {...props} />, {}, history);
+
+            expect(await screen.findByText("Change Password")).toBeInTheDocument();
+
+            expect(screen.queryByLabelText("admin")).not.toBeInTheDocument();
+            expect(screen.queryByText("User Role")).not.toBeInTheDocument();
+            expect(screen.getByText("Group4")).toBeInTheDocument();
+            expect(screen.getByText("create_sample")).toBeInTheDocument();
+
+            scope.isDone();
```
ReeceHoffmann:
```suggestion
            scope.done();
```
I think `scope.isDone()` just returns a bool indicating if all mocked scopes were called as expected. `scope.done()` raises an error if not everything has been called. Only essential if you expecting a mutation to have fired

ReeceHoffmann (changes requested):

src/tests/fake/user.tsx
```diff
@@ -101,3 +103,29 @@ export function mockApiFindUsers(users: Array<User>, query?: Query) {
...
 }
+
+/**
+ * Mocks an API call for getting the user details
+ *
+ * @param userId - The users unique id
+ * @param userDetail - The details of the user
+ * @returns A nock scope for the mocked API call
+ */
+export function mockApiGetUser(userId: string, userDetail: User) {
```
ReeceHoffmann:
```suggestion
export function mockApiGetUser(userId: string, user: User) {
```
The way we have named our types `userDetailed` is the default state whereas the reduced cases are the special cases

ReeceHoffmann (commented):
This looks good to me. I had a naming nitpick but other than that I am happy to merge. Nice work!

ReeceHoffmann (approved):

PR #289, opened 2023-12-07
chore: use react query for reference list view

raastinn:


src/js/base/IconLink.tsx
```diff
@@ -11,7 +11,7 @@ type IconLinkProps = {
 };
 
 export const IconLink = ({ color, name, replace, tip, to }: IconLinkProps) => (
-    <Link replace={replace} to={to}>
+    <Link replace={replace} to={to} aria-label={name}>
```
ReeceHoffmann: Careful what you pass to aria-label: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label

src/js/references/hooks.ts
```diff
@@ -8,9 +8,7 @@ import { Request } from "../app/request";
 import { Reference } from "./types";
 
 function getReference(refId: string): Reference {
-    return Request.get(`/refs/${refId}`).then(response => {
-        return new Reference(response.body);
-    });
+    return Request.get(`/refs/${refId}`).then(response => response.body);
 }
```
ReeceHoffmann: Should be in the api.ts file

src/js/references/types.ts
```diff
@@ -81,56 +82,56 @@ export type ReferenceNested = {
...
+    task: Task;
+    unbuilt_change_count: number;
+    updating: boolean | null;
     user: UserNested;
-    users: ReferenceUser[];
-
-    constructor(data) {
-        this.clonedFrom = data.cloned_from;
-        this.contributors = data.contributors;
-        this.createdAt = new Date(data.created_at);
```
ReeceHoffmann, line 113: Make sure that the components that made use of this can handle the change. This class was converting python `snake case` variables to more java-script idiomatic `pascal case` naming.

ReeceHoffmann (changes requested):

ReeceHoffmann:
Deal with the conflicts then I think this is g2g

ReeceHoffmann (approved):

PR #290, opened 2023-12-07
chore: use react-query for hmm detail view

raastinn:


src/js/hmm/querys.ts
```diff
@@ -23,3 +25,15 @@ export function useListHmms(page: number, per_page: number, term?: string) {
...
+
+/**
+ * Fetches a single HMM
+ *
+ * @param hmmId - The id of the hmm to fetch
+ * @returns A single HMM
+ */
+export function useFetchHmm(hmmId: string) {
+    return useQuery<HMM>(hmmQueryKeys.detail(hmmId), () => fetchHmm(hmmId), {
+        retry: 0,
```
ReeceHoffmann: I assume this is to handle a 404? I would pass a callback instead that looks for the correct error and bails.

ReeceHoffmann (changes requested):

src/js/hmm/querys.ts
```diff
@@ -23,3 +25,26 @@ export function useListHmms(page: number, per_page: number, term?: string) {
         keepPreviousData: true,
     });
 }
+
+type Error = {
+    response: {
+        status: number;
+    };
+};
```
ReeceHoffmann:
```suggestion
type ErrorResponse = {
    response: {
        status: number;
    };
};
```

I would be tempted to fill this out a little more with anything else that came in the response and bump it into `js/types/ts`

ReeceHoffmann (approved):

PR #294, opened 2023-12-13
fix: use latest indices for quick analysis view

raastinn:


src/js/samples/components/SamplesList.tsx
```diff
@@ -55,6 +55,14 @@ export default function SamplesList() {
         return <LoadingPlaceholder />;
     }
 
+    const filteredIndexes = indexes.reduce((acc, current) => {
+        const existingIndex = acc.find(item => item.reference.id === current.reference.id);
+        if (!existingIndex || current.version > existingIndex.version) {
+            acc.splice(existingIndex ? acc.indexOf(existingIndex) : acc.length, 1, current);
+        }
+        return acc;
+    }, []);
```
ReeceHoffmann: I think it is worth taking another crack at the logic here. Your doing two iterations over acc per index,(once per find and once per indexOf) which could potentially be a lot of checks. My first instinct would to instead use an object keyed by `ref_id`, but lodash may offer other functions to help simplify.

ReeceHoffmann (approved):
Looks great!

PR #297, opened 2023-12-15
fix: excessive space in user creation dialog

raastinn:


src/js/users/components/CreateUser.tsx
```diff
@@ -15,40 +14,39 @@ type NewUser = {
...
+
 /**
  * A dialog for creating a new user
  */
-export default function CreateUser() {
+export default function CreateUser({ show, history }: CreateUserProps) {
     const mutation = useMutation(create, {
         onSuccess: () => {
-            history.replaceState({ createUser: false }, "");
+            history.push({ state: { createUser: false } });
```
ReeceHoffmann, line 30: Hmm why the change here? I feel like the use of replace state may have been deliberate to prevent having the modal on the history stack

src/js/users/components/ManageUsers.tsx
```diff
@@ -11,6 +13,7 @@ import { UsersList } from "./UsersList";
  * @returns The user management view
  */
 export function ManageUsers() {
+    const history = useHistory<{ createUser: boolean }>();
```
ReeceHoffmann, line 16: I think this still makes more sense as part of `CreateUser`. The strange desync issue experienced in testing seems worth confronting now, imo as we will likely have to deal with this down the road. If you haven't already, take some time to confirm the source of the problem and hunt for some other solutions. From my reading we should be using `MemoryRouter` more than we are in testing so we need to figure this out anyway. Happy to talk or help out if you get stuck.

ReeceHoffmann (changes requested):

ReeceHoffmann:
Given that `reach` seems to be the cause of our test flake issues I am proposing we close and start fresh with a new issue that just cleans up the look of the current modal rather than reverting to the old one. Happy to leave open if you think this is an easier starting point than fresh

igboyes:
@ReeceHoffmann @raastinn Is this being addressed?

PR #307, opened 2024-01-03
chore: update reference creation dialouges

raastinn:


src/js/references/components/CreateReference.tsx
```diff
@@ -1,28 +1,30 @@
...
         <ContainerNarrow>
             <ViewHeader title="Create Reference">
                 <ViewHeaderTitle>Create Reference</ViewHeaderTitle>
             </ViewHeader>
             <Tabs>
-                <TabsLink
-                    to={{ state: { emptyReference: true } }}
-                    isActive={(_, location) => location.state.emptyReference}
-                >
+                <TabsLink to={{ state: { emptyReference: true } }} isActive={() => location.state.emptyReference}>
```
ReeceHoffmann, line 24:
```suggestion
                <TabsLink to={{ state: { emptyReference: true } }} isActive={location.state.emptyReference}>
```
Can you just exclude the arrow functions? Seems like the kind of thing you should be able to do, but depends on TabsLink api.

src/tests/fake/references.ts
```diff
@@ -125,3 +126,23 @@ export function mockApiCloneReference(name: string, description: string, referen
...
+ */
+export function mockApiCreateReference(
+    name: string,
+    description: string,
+    data_type: ReferenceDataType,
+    organism: string,
+) {
+    const reference = createFakeReference({ name, description, data_type, organism });
+
+    return nock("http://localhost").post("/api/refs", { name, description, data_type, organism }).reply(201, reference);
```
ReeceHoffmann, line 147: Does the API return the entire ref? Your typing with the mutation suggests your only getting some value back, but here your returning the entire reference. Update whichever one is incorrect

ReeceHoffmann (changes requested):

src/js/references/components/CreateReference.tsx
raastinn: No it seem as if the TabsLink api requires an arrow function

ReeceHoffmann (approved):

PR #308, opened 2024-01-04
fix: resolve visual bugs in sequence form

raastinn:


src/js/sequences/components/Field.jsx
```diff
@@ -1,29 +1,27 @@
 import PropTypes from "prop-types";
 import React from "react";
 import styled from "styled-components";
-import { Badge, InputContainer, InputError, InputGroup, InputLabel, TextArea } from "../../base";
+import { Badge, InputError, InputGroup, InputLabel, TextArea } from "../../base";
 export const SequenceFieldTextArea = styled(TextArea)`
     font-family: ${props => props.theme.fontFamily.monospace};
-    transform-text: uppercase;
+    text-transform: uppercase;
```
ReeceHoffmann, line 7: This only gives the appearance of uppercase without actually upper casing the letters in state. So the server still gets lower case letters which it won't accept

src/js/base/DialogContent.tsx
```diff
@@ -22,7 +22,7 @@ export const DialogContent = styled(DialogPrimitive.Content)<DialogContentProps>
     border-radius: ${props => props.theme.borderRadius.md};
     box-shadow: ${props => props.theme.boxShadow.lg};
     position: fixed;
-    top: 40%;
+    top: 50%;
```
ReeceHoffmann: Hmm this works, but the original looks better with some of the other dialogues. (create user for example) Take a look to see if you can find a way to have the position adjust with the size of the window. (vh + vw could be handy) Tricky thing here could be dealing with the fact that the base component doesn't know what height the final modal will be

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #309, opened 2024-01-05
chore: modernise subtractions detail view

raastinn:


src/js/subtraction/querys.ts
```diff
@@ -15,3 +31,42 @@ export function useFindSubtractions(page: number, per_page: number, term: string
...
+ */
+export function useFetchSubtraction(subtractionId: string) {
+    return useQuery<Subtraction, ErrorResponse>(
+        subtractionQueryKeys.detail(subtractionId),
+        () => getSubtraction(subtractionId),
+        {
+            retry: (failureCount, error) => {
+                if (error.response?.status === 404) {
+                    return false;
+                }
```
ReeceHoffmann: I think It might actually make more sense to bump this into the queryClient definition in `app/App`. 
I can't see a broad benefit to re-trying requests that got 404'ed

src/js/subtraction/reducer.js
```diff
@@ -21,35 +12,14 @@ export const initialState = {
...
-        })
-        .addCase(WS_UPDATE_SUBTRACTION, (state, action) => {
-            return update(state, action.payload, "id");
-        })
         .addCase(WS_REMOVE_SUBTRACTION, (state, action) => {
             return remove(state, action.payload);
         })
         .addCase(WS_UPDATE_JOB, (state, action) => {
             return updateJobs(state, action.payload);
         })
```
ReeceHoffmann: Do either of these do anything if we aren't populating the main list of subtractions?

src/js/subtraction/reducer.js
```diff
@@ -1,15 +1,6 @@
...
-    WS_REMOVE_SUBTRACTION,
-    WS_UPDATE_JOB,
-    WS_UPDATE_SUBTRACTION,
-} from "../app/actionTypes";
-import { insert, remove, update, updateDocuments, updateJobs } from "../utils/reducers";
+import { SHORTLIST_SUBTRACTIONS, WS_REMOVE_SUBTRACTION, WS_UPDATE_JOB } from "../app/actionTypes";
+import { remove, updateJobs } from "../utils/reducers";
 
 export const initialState = {
     detail: null,
```
ReeceHoffmann, line 5: This can probably be cut way down

src/js/subtraction/reducer.js
```diff
@@ -21,35 +12,14 @@ export const initialState = {
...
             return updateJobs(state, action.payload);
         })
-        .addCase(FIND_SUBTRACTIONS.REQUESTED, (state, action) => {
-            state.term = action.payload.term;
-        })
-        .addCase(FIND_SUBTRACTIONS.SUCCEEDED, (state, action) => {
-            return updateDocuments(state, action.payload, "name");
-        })
         .addCase(SHORTLIST_SUBTRACTIONS.SUCCEEDED, (state, action) => {
             state.shortlist = action.payload;
```
ReeceHoffmann: How many components are still using the redux implementation of the subtraction shortlist at this point? Can we fully remove this if sample creation is merger first?

src/js/subtraction/components/__tests__/Subtraction.test.tsx
```diff
@@ -86,9 +87,10 @@ describe("<Subtraction />", () => {
     });
 
     it("should render /subtractions/:subtractionId route", async () => {
-        history.push(`/subtractions/${subtractionMinimal.id}`);
+        const subtraction = createFakeSubtraction();
+        history.push(`/subtractions/${subtraction.id}`);
```
ReeceHoffmann, line 91: Not a change you made, but looks like the same history object is being used in every test in this module which is bad from a test isolation perspective. A new history object should be made before every test via `beforeEach`

ReeceHoffmann (changes requested):

src/js/subtraction/reducer.js
raastinn: `CreateAnalysis`, `QuickAnalyze`, and `SampleDetail` are still using subtraction shortlist

ReeceHoffmann (approved):

PR #310, opened 2024-01-08
chore: rework`EmptyReference` to use react-hook-form

raastinn:


src/js/references/components/ReferenceForm.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const StyledInputGroup = styled(InputGroup)`
+    padding-bottom: 0px;
+`;
+
+type ReferenceFormProps = {
+    /** Form validation errors */
+    errors: FieldErrors<FormValues>;
+    /** The mode of the reference form */
+    mode: string;
```
ReeceHoffmann: Should be an enum dictating what the valid states are. Look for usages and create an enum around expected values

src/js/references/components/ReferenceForm.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * Form input fields for organism, name and description
+ */
+export function ReferenceForm({ errors, mode, register }: ReferenceFormProps) {
+    let organismComponent;
+
+    if (mode === "empty" || mode === "edit") {
+        organismComponent = (
+            <InputGroup>
+                <InputLabel htmlFor="organism">Organism</InputLabel>
+                <InputSimple as="textarea" id="organism" {...register("organism")} />
```
ReeceHoffmann: Why the change from a regular input to a textarea?

src/js/references/components/ReferenceForm.tsx
```diff
@@ -0,0 +1,51 @@
...
+                <InputLabel htmlFor="name">Name</InputLabel>
+                <InputSimple id="name" {...register("name", { required: "Required Field" })} />
+                <InputError>{errors.name?.message}</InputError>
+            </StyledInputGroup>
+
+            {organismComponent}
+
+            <InputGroup>
+                <InputLabel htmlFor="description">Description</InputLabel>
+                <InputSimple as="textarea" id="description" {...register("description")} />
```
ReeceHoffmann: Why are you using passing `textarea` vs simply using `TextArea`? This will apply the styles from `InputSimple` which may not be what you want when creating a larger text input

src/js/references/components/__tests__/EmptyReference.test.tsx
```diff
@@ -52,4 +52,14 @@ describe("<EmptyReference />", () => {
...
+
+    describe("<ReferenceForm />", () => {
+        it("should render", async () => {
+            renderWithProviders(<EmptyReference />);
+
+            await userEvent.click(screen.getByRole("button", { name: "Save" }));
+
+            expect(screen.getByText("Required Field")).toBeInTheDocument();
+        });
+    });
```
ReeceHoffmann: Looks like a dupe of an above test

src/js/references/components/ReferenceForm.tsx
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+import { FieldErrors, UseFormRegister } from "react-hook-form";
+import styled from "styled-components";
+import { InputError, InputGroup, InputLabel, InputSimple } from "../../base";
+import { FormValues } from "./EmptyReference";
```
ReeceHoffmann: imo it is a bit weird for `FormValues` to be defined outside of the actual form module. Esp when `ReferenceForm` is used by other components

ReeceHoffmann (changes requested):
Code looks overall really close to what we want. Keep a close eye out for styling changes and consider if they both look better and/or provide a better user experience than the previous version.

src/js/base/TextArea.tsx
```diff
@@ -8,6 +8,6 @@ const StyledTextArea = styled(Input)`
     overflow-y: scroll;
 `;
 
-export function TextArea(props) {
-    return <StyledTextArea as="textarea" {...props} />;
+export function TextArea({ register, name, ...props }) {
+    return <StyledTextArea as="textarea" {...(register && register(name))} {...props} />;
```
ReeceHoffmann: This because of `ref`? `React.forwardRef` might be a cleaner soln here

src/js/references/components/ReferenceForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { FieldErrors, UseFormRegister } from "react-hook-form";
+import styled from "styled-components";
+import { InputError, InputGroup, InputLabel, InputSimple, TextArea } from "../../base";
+import { ReferenceDataType } from "../types";
+
+const StyledInputGroup = styled(InputGroup)`
+    padding-bottom: 0px;
+`;
+
+export enum Mode {
```
ReeceHoffmann: Will this name make sense in the importing component?

src/js/references/components/ReferenceForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+    register: UseFormRegister<FormValues>;
+};
+
+/**
+ * Form input fields for organism, name and description
+ */
+export function ReferenceForm({ errors, mode, register }: ReferenceFormProps) {
+    let organismComponent;
+
+    if (mode === "empty" || mode === "edit") {
```
ReeceHoffmann: Since your defining an enum you can do the comparison with the enum rather than the raw string

ReeceHoffmann (changes requested):
Small stuff. Would be nice to clean up the API `TextArea`. Okay to merge either way on the `Mode` comments

raastinn:
Updated `TextArea` to use `React.forwardRef` rather than the method before. Imo the previous one looks more cleaned up, let me know what you think.

ReeceHoffmann (approved):

PR #311, opened 2024-01-09
refactor: update barcode sequence addition and editing to current libraries

raastinn:
Screenshot from 2024-01-09 16-53-31
Updated from reach-ui to radix-ui by changing `Modal` to `Dialog` and `ListBox` to `Select`


src/js/sequences/components/Barcode/AddBarcodeSequence.tsx
```diff
@@ -0,0 +1,145 @@
...
+export function castValues(targets: ReferenceTarget[], defaultTarget: string) {
+    return function (values: formValues) {
+        const targetName = find(targets, { name: values.targetName }) ? values.targetName : defaultTarget;
+        return { ...values, targetName };
+    };
+}
+
+export const StyledContent = styled(DialogContent)`
+    top: 50%;
+`;
```
ReeceHoffmann: Couple things:

 1. best to avoid more than one component export per file
 2. Naming convention thing. `styled` naming convention is information poor and generally reserved for cases where your wrapping the current component in a div for styling. For example you could define a `styledAddBarcodeSequenceProps` that wraps the JSX returned from `AddBarcodeSeqeunceProps` and be within convention

src/js/sequences/components/Barcode/SequenceTarget.tsx
```diff
@@ -0,0 +1,57 @@
+import * as RadixSelect from "@radix-ui/react-select";
+import React from "react";
+import styled from "styled-components";
+import { fontWeight, getColor, getFontSize, getFontWeight } from "../../../app/theme";
+
+const StyledSelectItem = styled(RadixSelect.Item)`
```
ReeceHoffmann: Same comment as above re `styled` naming usage

src/js/sequences/components/Barcode/SequenceTarget.tsx
```diff
@@ -0,0 +1,57 @@
...
+    position: relative;
+    user-select: none;
+    margin-bottom: 5px;
+    text-transform: capitalize;
+
+    :hover {
+        background-color: ${({ theme }) => getColor({ color: "greyHover", theme })};
+        border: 0;
+    }
+`;
```
ReeceHoffmann: Is this copied from the base component equivalent? If so, it might make sense to tweak the base component, as I suspect the main problem would be naming one

ReeceHoffmann (changes requested):
Code overall looks good and I like the visual look of the new component. Few small convention tweaks then lgtm!

src/js/base/select/SelectItem.tsx
```diff
@@ -21,19 +21,10 @@ const StyledSelectItem = styled(RadixSelect.Item)`
...
-`;
-
-export const SelectItem = ({ value, text, description }) => {
+export function SelectItem({ value, children }) {
     return (
         <StyledSelectItem value={value} key={value}>
-            <RadixSelect.ItemText>{text}</RadixSelect.ItemText>
-            <Description>{description}</Description>
+            {children}
         </StyledSelectItem>
```
ReeceHoffmann: How about something like this?
```
export function SelectItem({ value, children, description }) {
  <StyledSelectItem value={value} key={value} >
            <RadixSelect.ItemText>{children}</RadixSelect.ItemText>
            {description && <Description>{description}</Description>}
           
  </StyledSelectItem>
```
This way we can keep the abstraction on `RadixSelect.ItemText` and keep consistency between components about how the description would be rendered.

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #312, opened 2024-01-10
chore: rework `SampleRights` & update related components

raastinn:


src/js/samples/components/SampleRights.tsx
```diff
@@ -63,49 +78,36 @@ export const SampleRights = props => {
...
                     </InputSelect>
                 </InputGroup>
 
                 <InputGroup>
                     <InputLabel>All Users' Rights</InputLabel>
                     <InputSelect
-                        name="all"
-                        value={props.all}
-                        onChange={e => props.onChangeRights("all", e.target.value)}
+                        aria-label="all"
```
ReeceHoffmann: Ideally you would get the label based on the content of the `InputLabel` above. Failing that, the `aria-label` should make sense for anyone using accessibility software

ReeceHoffmann (commented):
Looks good! just a tweak to the usage of `aria-label`

ReeceHoffmann (approved):

PR #320, opened 2024-01-17
fix: password reset workflow

raastinn:


src/js/wall/Reset.jsx
```diff
@@ -14,13 +14,22 @@ export class Reset extends React.Component {
...
         this.setState({ password: e.target.value });
     };
 
     handleSubmit = e => {
         e.preventDefault();
-        this.props.onReset(this.state.password, this.props.resetCode);
+
+        if (!this.state.password.length < 8) {
+            this.props.onReset(this.state.password, this.state.resetCode);
+        }
```
ReeceHoffmann: Careful! This conditional always resolves to true as written
```suggestion
        if (!(this.state.password.length < 8)) {
            this.props.onReset(this.state.password, this.state.resetCode);
        }
```

Also: Will this render an error to the user if the request is blocked? I don't see it in the code, and I would hate to block the user but not say why

src/js/wall/Reset.jsx
```diff
@@ -14,13 +14,22 @@ export class Reset extends React.Component {
         };
     }
 
+    componentDidMount() {
+        this.setState({
+            resetCode: this.props.resetCode,
```
ReeceHoffmann: Did you figure out why the reset code get forgotten after the first request? Not totally against this soln if it makes the most sense, just seems strange that the code persists for only one request

raastinn:
The reason why `resetCode` was being set to `undefined` when reset_password failed was the reducer changing the reset code I believe.

ReeceHoffmann (approved):

PR #321, opened 2024-01-19
chore: modernise subtractions `Remove` and related components

raastinn:


src/js/subtraction/components/Detail/EditSubtraction.tsx
```diff
@@ -24,7 +24,7 @@ type EditSubtractionProps = {
 };
 
 /**
- * A modal for editing the subtraction
+ * Displays a modal for editing a subtraction
```
ReeceHoffmann: Nit: The current goal for component docstrings is to describe what they are rather than being action focused like a normal function. 

```suggestion
 * Dialog for editing an existing subtraction
```
or something like that.

src/js/subtraction/components/Detail/RemoveSubtraction.tsx
```diff
@@ -0,0 +1,30 @@
...
+    /** The subtraction data */
+    subtraction: Subtraction;
+    /** Indicates whether the modal for removing a subtraction is visible */
+    show: boolean;
+    /** A callback function to hide the modal */
+    onHide: () => void;
+};
+
+/**
+ * Displays a modal for removing a subtraction
```
ReeceHoffmann: Same comment on component docstrings

src/js/subtraction/querys.ts
```diff
@@ -15,6 +23,23 @@ export const subtractionQueryKeys = {
...
+ * @returns A mutator for creating a subtraction
+ */
+export function useCreateSubtraction() {
+    const history = useHistory();
+    return useMutation<Subtraction, unknown, { name: string; nickname: string; uploadId: string }>(
+        ({ name, nickname, uploadId }) => createSubtraction(name, nickname, uploadId),
+        {
+            onSuccess: () => {
+                history.push("/subtractions");
+            },
```
ReeceHoffmann: You can and should specify navigation in the component instead of with the mutation.  You can specify `onSuccess` when you call `mutation.mutate` in the the component.

check out this post: https://tkdodo.eu/blog/mastering-mutations-in-react-query Covers the difference between specifying `onSuccess` on the `useMutation` vs `mutation.mutate` level

src/js/subtraction/querys.ts
```diff
@@ -63,6 +88,23 @@ export function useUpdateSubtraction(subtractionId: string) {
...
+ * @returns A mutator for removing a subtraction
+ */
+export function useRemoveSubtraction() {
+    const history = useHistory();
+    return useMutation<Response, unknown, { subtractionId: string }>(
+        ({ subtractionId }) => removeSubtraction(subtractionId),
+        {
+            onSuccess: () => {
+                history.push("/subtractions");
+            },
```
ReeceHoffmann: Same deal here, should be in the component

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #324, opened 2024-01-22
chore: update `EditSample` component

raastinn:
Screenshot from 2024-01-22 15-00-15
Updated styling from `Modal` to `Dialog`


src/js/samples/components/Detail/General.jsx
```diff
@@ -21,117 +22,109 @@ const StyledSampleDetailGeneral = styled.div`
...
-                        <tr>
-                            <th>Locale</th>
-                            <td>{locale}</td>
-                        </tr>
-                    </tbody>
-                </Table>
-            </BoxGroup>
+export function SampleDetailGeneral({ count, encoding, gc, lengthRange, libraryType, sample }) {
+    const { name, host, isolate, locale, paired, notes } = sample;
+    const history = useHistory();
```
ReeceHoffmann: Why no ts + docstring?

src/js/samples/components/Detail/FileSizeWarning.jsx
```diff
@@ -28,7 +28,6 @@ export const SampleFileSizeWarning = ({ sampleId, show, showLink }) => {
 };
 
 export const mapStateToProps = state => ({
-    sampleId: state.samples.detail.id,
```
ReeceHoffmann: Why? seems like sampleId is being used in the component to create a link

src/js/samples/components/Detail/__tests__/General.test.jsx
```diff
@@ -1,22 +1,21 @@
...
             encoding: "Sanger / Illumina 2.1",
             gc: "42.3%",
-            host: "Malus domestica",
-            isolate: "Isolate Foo",
             lengthRange: "41 - 76",
-            locale: "Bar",
-            name: "Foo",
-            notes: "Example notes",
-            paired: false,
+            sample: {},
```
ReeceHoffmann: Hmm seems like you just removed a lot of data. Is none of this getting used by the component being tested?

src/js/samples/components/Detail/__tests__/General.test.jsx
```diff
@@ -25,12 +24,23 @@ describe("<SampleDetailGeneral />", () => {
...
-    it.each([true, false])("should render with [paired=%p]", paired => {
-        props.paired = paired;
-        const wrapper = shallow(<SampleDetailGeneral {...props} />);
-        expect(wrapper).toMatchSnapshot();
+    it("should render with [paired=true]", () => {
+        props.sample = createFakeSample({ paired: true });
+        renderWithRouter(<SampleDetailGeneral {...props} />, {}, history);
+
+        expect(screen.getByText("Paired")).toBeInTheDocument();
+        expect(screen.getByText("Yes")).toBeInTheDocument();
```
ReeceHoffmann: Does this cover all the fields that should render?

src/js/samples/components/EditSample.tsx
```diff
@@ -0,0 +1,86 @@
...
+export default function EditSample({ sample, show, onHide }: EditSampleProps) {
+    const mutation = useUpdateSample(sample.id);
+
+    const { register, handleSubmit } = useForm({
+        defaultValues: {
+            name: sample.name || "",
+            isolate: sample.isolate || "",
+            host: sample.host || "",
+            locale: sample.locale || "",
+            notes: sample.notes || "",
```
ReeceHoffmann: Don't think it will make a big difference in this case, but I think for checking existence `??` is more strictly correct

src/js/samples/components/__tests__/EditSample.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+        const isolateInput = screen.getByLabelText("Isolate");
+        expect(isolateInput).toBeInTheDocument();
+        expect(isolateInput).toHaveValue(sample.isolate);
+
+        await userEvent.clear(isolateInput);
+        expect(isolateInput).toHaveValue("");
+
+        await userEvent.type(isolateInput, "test");
+        expect(isolateInput).toHaveValue("test");
```
ReeceHoffmann: I think there is room for improving these ` should render after x has changed` tests. Couple of ways I could see it going though.

Option 1:  Do the initial value check for all of them as part of a `should render` test then check the typing in the final `should submit` test. Would reduced test redundancy and having to render the dom as many times.

Option 2: you can probably use it.each since the tests are so similar.

src/js/samples/querys.ts
```diff
@@ -37,6 +41,23 @@ export function useListSamples(page: number, per_page: number, term?: string, la
...
+ * @returns A mutator for updating a sample
+ */
+export function useUpdateSample(sampleId: string) {
+    const history = useHistory();
+    return useMutation<Sample, ErrorResponse, { update: SampleUpdate }>(
+        ({ update }) => updateSample(sampleId, update),
+        {
+            onSuccess: () => {
+                history.replace({ state: { editSample: false } });
+            },
```
ReeceHoffmann: I would keep navigation out of the hook defn. do it in the component

ReeceHoffmann (commented):
Some tweaks to the code. Also looks like there is spacing differences between the inputs, please resolve.

src/js/samples/components/Detail/__tests__/General.test.jsx
raastinn: I redefine the prop in the tests that use it using `createFakeSample`. Didn't declare it in the props because the tests that use it need different properties so I just defined it in the test directly.

ReeceHoffmann (commented):
Looks like merging the sample typescript PR caused some conflicts. fix then LGTM!

ReeceHoffmann (approved):

PR #331, opened 2024-01-29
refactor: simplify `ReferenceList` items

raastinn:
Screenshot from 2024-01-29 11-17-10
Also replaced the task bar with a progress circle. Its currently functional, but it might be better to create an issue later on to make a progress circle component for tasks.


igboyes (changes requested):
* Fix inconsistent vertical position of user and timestamp.
* Make sure the icons on the right are centred on the same vertical axis.

raastinn:
Screenshot from 2024-01-29 15-59-11

igboyes (approved):

PR #336, opened 2024-01-31
refactor: update styling for upload items

raastinn:
Screenshot from 2024-01-30 17-08-03


src/js/base/Pagination.tsx
```diff
@@ -63,6 +65,7 @@ interface PaginationProps {
 }
 
 export const Pagination = ({
+    boxGroup,
```
igboyes: This isn't a good solution. Flags are almost always bad.

What about just putting a `margin-top` on `PaginationBox` instead and removing `PaginationContainer` entirely?

igboyes (changes requested):

igboyes (approved):

PR #346, opened 2024-02-05
chore: rework `AddIsolate` and update the form & source types

raastinn:
Screenshot from 2024-02-05 13-54-31
- Updated from `Modal` to the radix `Dialog`
- Moved the `SourceType` file into the Isolates folder since its only being used within that folder
- Removed the `SourceType` tests since its redundant and already being tested in the `AddIsolates` test


src/js/otus/components/Detail/Isolates/EditIsolate.jsx
```diff
@@ -0,0 +1,49 @@
...
+                        sourceName={props.sourceName}
+                        allowedSourceTypes={props.allowedSourceTypes}
+                        restrictSourceTypes={props.restrictSourceTypes}
+                        onSubmit={handleSubmit}
+                    />
+                </DialogContent>
+            </DialogPortal>
+        </Dialog>
+    );
+}
```
ReeceHoffmann: I feel like complete conversion would be in scope for this PR. Not blocking, will merge either way. Also happy with an issue made and merging as is

src/js/otus/components/Detail/Isolates/SourceType.tsx
```diff
@@ -0,0 +1,65 @@
...
+        ));
+
+        return (
+            <InputGroup>
+                <InputLabel htmlFor="sourceType">Source Type</InputLabel>
+                <Controller
+                    name="sourceType"
+                    control={control}
+                    render={({ field: { onChange, value } }) => (
+                        <InputSimple
```
ReeceHoffmann: I think the styled for `InputSimple` are really intended to cover the just a regular text input. I would lean towards using the `Select` base component unless there is a good reason not to

src/js/otus/components/Detail/Isolates/SourceType.tsx
```diff
@@ -0,0 +1,65 @@
...
+        const optionComponents = map(allowedSourceTypes, sourceType => (
+            <option key={sourceType} value={capitalize(sourceType)}>
+                {capitalize(sourceType)}
+            </option>
+        ));
+
+        return (
+            <InputGroup>
+                <InputLabel htmlFor="sourceType">Source Type</InputLabel>
+                <Controller
```
ReeceHoffmann: Does react-hook-form not support select lists more directly than this? Select lists are a normal HTML pattern, and I get the impression from the react-hook-form docs that `Controller` is more of a backup escape hatch than meant to be used normally

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #348, opened 2024-02-05
refactor: flatten subtraction items

raastinn:
Screenshot from 2024-02-05 15-19-45

### DIFF
--- a/src/js/subtraction/components/SubtractionItem.tsx
+++ b/src/js/subtraction/components/SubtractionItem.tsx
@@ -1,25 +1,41 @@
 import React from "react";
+import { Link } from "react-router-dom";
 import styled from "styled-components";
 import { getFontSize, getFontWeight, sizes } from "../../app/theme";
-import { BoxLink } from "../../base";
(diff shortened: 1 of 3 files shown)


igboyes (changes requested):
Make visual changes discussed in call.

raastinn:
Screenshot from 2024-02-07 12-23-07
Made the attribution left-aligned. IMO this looks a bit weird if there are subtractions being created. Maybe we could put like an edit icon or something on the left of all of them and have 4 columns? Could look better if we only keep the progress circle and not display the status text

ReeceHoffmann:
Not privy to what your guys discussion, but this seems stalled out so i'll comment. I would try the 4 column solution, but if creation is complete, have nothing in that new 4th column.  The 4th column should be just big enough for the circle + margin, the text should be removed for this to look good. I would also render the regardless of progress.

raastinn:
Screenshot from 2024-02-21 14-49-09

igboyes (approved):

PR #353, opened 2024-02-07
fix: user detail groups not ordered alphabetically

raastinn:


src/js/users/components/UserGroups.tsx
```diff
@@ -39,7 +39,7 @@ export default function UserGroups({ memberGroups, userId }: UserGroupsType) {
             <label>Groups</label>
             <UserGroupsList>
                 {documents.length ? (
-                    map(documents, ({ id, name }) => (
+                    map(sortBy(documents, [document => document.name.toLowerCase()]), ({ id, name }) => (
```
ReeceHoffmann, line 42:
```suggestion
                    map(sortBy(documents, document => document.name.toLowerCase()), ({ id, name }) => (
```
?

igboyes:
Let's hold off on this one. I think it may be better to rely on sorting on the API which is not working at the moment.

igboyes:
Closing because these are sorted in the API response now.

PR #355, opened 2024-02-09
fix: `ScrollList` to trigger re-request in absense of scrolling

raastinn:


src/js/base/ScrollList.tsx
```diff
@@ -39,8 +41,15 @@ export function ScrollList({
...
     items,
     renderRow,
+    hasNextPage,
 }: ScrollListProps) {
     useEffect(() => {
+        const isScrollEnabled = document.documentElement.scrollHeight > window.innerHeight;
+
+        if (!isScrollEnabled && hasNextPage) {
+            void fetchNextPage();
+        }
```
ReeceHoffmann: Would it make more sense to simply call onScroll? Seems like it would be equivalent or even better since it would also load for cases where the user is nearly at the bottom as well. Might need a rename since it will be invoked outside the context of scrolling

src/js/base/ScrollList.tsx
```diff
@@ -39,16 +41,23 @@ export function ScrollList({
...
     renderRow,
+    hasNextPage,
 }: ScrollListProps) {
     useEffect(() => {
-        function onScroll() {
+        const isScrollEnabled = document.documentElement.scrollHeight > window.innerHeight;
+
+        if (!isScrollEnabled && hasNextPage) {
+            handleFetchNextPage();
+        }
```
```suggestion
        handleFetchNextPage()
```
? (probably also a good idea to move calling the function below the functions defn, java-script will hoist in many cases, but I don't like relying on it)

raastinn: Removing the if statement and just having the `handleFetchNextPage() function called loads all pages on initial render even if scroll is enabled

ReeceHoffmann: Ah interesting, do you know why? I would have thought that as more pages are loaded the scroll percentage would drop preventing further loading.

raastinn: Seems like the `getScrollRatio` consistently returns 1 so it basically spams the `fetchNextPage`.

ReeceHoffmann (approved):

PR #357, opened 2024-02-12
chore: update `SampleRights` to use react-query in detailed view

raastinn:
**_NOTE:_** Updating the group and user rights gives a 500 error but updating the group works.


ReeceHoffmann (commented):
Took a look and it seems like the 500 error occurs with the old version as well. that said not pumped to merge with a known 500 error. I'll take a look at the backend, it may be expecting a different format or simply be bugged.

Might end up needing to update the mutation if the back-end expects a different format, otherwise code looks good to me

ReeceHoffmann:
Backend is bugged, working on getting a fixed release out then this is good to merge imo

ReeceHoffmann (approved):

PR #358, opened 2024-02-14
chore: rework `AddReferenceMember` in reference detail view to use react-query

raastinn:
Screenshot from 2024-02-14 11-36-38
Screenshot from 2024-02-14 11-36-46


src/js/groups/api.ts
```diff
@@ -1,12 +1,31 @@
...
 
+/**
+ * Fetch a page of group search results
+ *
+ * @param page - The page to fetch
+ * @param per_page - The number of groups to fetch per page
+ * @param term - The search term to filter groups by
+ * @returns A promise resolving to a page of group search results
+ */
+export function findGroups(page?: number, per_page?: number, term?: string): Promise<GroupSearchResults> {
```
ReeceHoffmann: Naming/functionality of these two is too close. I would probably merge them. If there is a good reason to have them separate then rename them and for the paginated one force the calling function to pass the current page

src/js/groups/querys.ts
```diff
@@ -9,10 +9,35 @@ export const groupQueryKeys = {
     all: () => ["groups"] as const,
     lists: () => ["groups", "list"] as const,
     list: filters => ["groups", "list", ...filters] as const,
+    infiniteLists: () => ["groups", "infiniteList"] as const,
+    infiniteList: (filters: Array<string | number | boolean>) => ["groups", "infiniteList", ...filters] as const,
```
ReeceHoffmann:
```suggestion
    infiniteLists: () => ["groups", "list", "infinite"] as const,
    infiniteList: (filters: Array<string | number | boolean>) => ["groups", "list", "infinite", ...filters] as const,
```
I have been leaning towards this key structure so that you only have to do one invalidation to hit both lists

src/js/references/components/Detail/ReferenceSettings.tsx
```diff
@@ -12,23 +14,59 @@ type ReferenceSettingsProps = {
...
+            <ReferenceMembers
+                data={groups}
+                noun="group"
+                members={sortBy(reference.groups, "id")}
+                refId={refId}
+                term={term}
+                setTerm={setTerm}
+                isFetchingNextPage={isFetchingGroupsNextPage}
+                fetchNextPage={fetchGroupsNextPage}
+                isLoading={isLoadingGroups}
```
ReeceHoffmann: To me the state seems to far up. The data seems to only be used by components further down in the tree. I think it would be wroth taking the time to investigate a solution that localises data + state where it is getting used rather than so far up

src/js/users/querys.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Factory object for generating user query keys
+ */
+export const userQueryKeys = {
+    all: () => ["users"] as const,
+    lists: () => ["users", "list"] as const,
+    list: (filters: Array<string | number | boolean>) => ["users", "list", ...filters] as const,
+    infiniteLists: () => ["users", "infiniteList"] as const,
+    infiniteList: (filters: Array<string | number | boolean>) => ["users", "infiniteList", ...filters] as const,
```
ReeceHoffmann: Same comment as other key section

ReeceHoffmann (changes requested):

ReeceHoffmann (approved):

PR #359, opened 2024-02-14
chore: rework OTU create to use react-query and typescript

raastinn:
Screenshot from 2024-02-14 14-28-05


src/js/otus/components/OTUList.tsx
```diff
@@ -27,6 +27,9 @@ type OTUListProps = {
  */
 export default function OTUList({ match }: OTUListProps) {
     const { refId } = match.params;
+    const history = useHistory();
+    const location = useLocation<{ createOTU: boolean }>();
```
ReeceHoffmann: Why handle this logic here instead of in CreateOtu? Both should have access to the same history object

src/tests/fake/otus.ts
```diff
@@ -1,15 +1,71 @@
...
 
 type CreateFakeOTUMinimalProps = {
+    abbreviation?: string;
+    name?: string;
     verified?: boolean;
 };
 
 /**
- * Create a fake HMM minimal
+ * Create a fake OTU minimal
```
ReeceHoffmann:
```suggestion
 * Create a fake OTUMinimal
```
or
```suggestion
 * Create a fake minimal OTU
```

ReeceHoffmann (changes requested):
Couple small things overall looks good

ReeceHoffmann (dismissed):

ReeceHoffmann (approved):

PR #361, opened 2024-02-16
fix: check correct rights for `OTUToolbar`

raastinn:


src/js/references/hooks.ts
```diff
@@ -128,3 +130,55 @@ export function useUpdateSourceTypes(
...
+    if (!groups) {
+        return false;
+    }
+
+    return some(groups, { [right]: true });
+}
+
+/**
+ * Get a boolean indicating whether the logged in account can modify the OTUs of the
+ * reference detail is loaded for.
```
ReeceHoffmann:
```suggestion
 * Get a boolean indicating whether the current user can modify the OTUs of a reference.
```
Careful on the wording, any reference id can be passed here

src/js/references/hooks.ts
```diff
@@ -128,3 +130,55 @@ export function useUpdateSourceTypes(
...
+ *
+ * The result depends on the account's rights on the reference. It also depends on whether the reference is a remote
+ * reference. Remote references cannot be modified by any user.
```
ReeceHoffmann:
```suggestion
 * Ensures the user has sufficient rights on the reference and that reference is not defined by an external resource
 
```
? Aiming for reduced verbosity

src/js/references/hooks.ts
```diff
@@ -128,3 +130,55 @@ export function useUpdateSourceTypes(
...
+ *
+ * @param reference - The reference to check
+ * @returns The OTU modification right of the account
+ */
+export function useCanModifyReferenceOTU(reference: Reference) {
+    const canModifyOTU = useCheckReferenceRight(reference, "modify_otu");
+    return !reference.remotes_from && canModifyOTU;
```
ReeceHoffmann: Imo these hooks should return their loading status via `isLoading`. Otherwise there isn't a great way for the consuming component to know if the result is because the data hasn't loaded or if the user really doesn't have permission.

src/js/references/hooks.ts
```diff
@@ -128,3 +130,55 @@ export function useUpdateSourceTypes(
...
 }
+
+/**
+ * Check if the logged in account has the passed `right` on the reference detail is loaded for.
+ *
+ * @param reference - The reference to check
+ * @param right - The right to check for (eg. modify_otu)
+ * @returns Whether the right is possessed by the account
+ */
+export function useCheckReferenceRight(reference: Reference, right: string) {
```
ReeceHoffmann:
```suggestion
export function useCheckReferenceRight(referenceId: string, right: ReferenceRight) {
```
Define an enum for this, prevents typo errors. I would also just pass the ref string. 

I am also personally leaning towards just passing the id in these requests, keeps the data flowing one way and imo makes it easier to reason through what happens at given load steps.

src/js/references/hooks.ts
```diff
@@ -128,3 +130,55 @@ export function useUpdateSourceTypes(
...
+    const { data: account, isLoading } = useFetchAccount();
+
+    if (isLoading || reference === null) {
+        return;
```
ReeceHoffmann:
```suggestion
        return {hasPermission: false, isLoading: true};
```
Same format for other return statements as well

src/js/references/hooks.ts
```diff
@@ -128,3 +130,55 @@ export function useUpdateSourceTypes(
...
+ * Get a boolean indicating whether the logged in account can modify the OTUs of the
+ * reference detail is loaded for.
+ *
+ * The result depends on the account's rights on the reference. It also depends on whether the reference is a remote
+ * reference. Remote references cannot be modified by any user.
+ *
+ * @param reference - The reference to check
+ * @returns The OTU modification right of the account
+ */
+export function useCanModifyReferenceOTU(reference: Reference) {
```
ReeceHoffmann: Looking at what is here, I wonder if it would make more sense to eliminate this hook and just use the enum with `useCheckReferenceRight`.

ReeceHoffmann (changes requested):

src/js/references/hooks.ts
```diff
@@ -128,3 +131,44 @@ export function useUpdateSourceTypes(
...
+ *
+ * @param referenceId - The id of the reference to check
+ * @param right - The right to check for (eg. modify_otu)
+ * @returns Whether the right is possessed by the account
+ */
+export function useCheckReferenceRight(referenceId: string, right: ReferenceRight) {
+    const queryClient = useQueryClient();
+    const { data: account, isLoading: isLoadingAccount } = useFetchAccount();
+
+    const reference: Reference = queryClient.getQueryData([...referenceQueryKeys.all(), referenceId]);
```
ReeceHoffmann: why not `UseGetReference`? Prevents implicit dependency on an element up the tree having already populated the cache, and will still pull from the cache if that's an option

src/js/references/hooks.ts
```diff
@@ -128,3 +131,44 @@ export function useUpdateSourceTypes(
...
+
+    if (isLoadingAccount || !reference) {
```
ReeceHoffmann:
```suggestion
    if (isLoadingAccount || isLoadingReference) {
```
With the tweak above

src/js/references/hooks.ts
```diff
@@ -128,3 +131,44 @@ export function useUpdateSourceTypes(
...
     };
 }
+
+/**
+ * All reference rights
+ */
+export enum ReferenceRight {
+    modify = "modify",
+    modify_otu = "modify_otu",
+}
```
ReeceHoffmann: build and remove?

ReeceHoffmann (changes requested):
Is this new version accounting for `remotes_from`? Looking at the rights list I think remotes_from would block `modify` and `modify_otu` but not `remove` or `build`.

ReeceHoffmann (approved):

PR #365, opened 2024-02-19
Chore: Rework `RemoveIsolate` component to current standards & create `RemoveDIalog`

raastinn:
Screenshot from 2024-02-19 13-20-56


ReeceHoffmann (commented):
Code looks good, that said the `undefined` in your screenshot is a bit concerning. Does that make sense in your local instance?

raastinn:
> Code looks good, that said the `undefined` in your screenshot is a bit concerning. Does that make sense in your local instance?

Yeah. When you create an isolate without a source type it sets it to undefined. This happened before the conversion to react-query as well I believe.

ReeceHoffmann:
Hmm something is going on here, on main when I make an isolate with no name it is assigned unknown: 
image

And when I go to delete it I get: 
image

Which is definitely better. Problem might be from work in another PR as well if the OTU your `deleteing` was made while working on something else

raastinn:
> Hmm something is going on here, on main when I make an isolate with no name it is assigned unknown: image
> 
> And when I go to delete it I get: image
> 
> Which is definitely better. Problem might be from work in another PR as well if the OTU your `deleteing` was made while working on something else

Screenshot from 2024-02-21 14-05-23
This works fine for me too.

Sorry, i meant when you try to create an isolate with a type but no name that this happens
Screenshot from 2024-02-21 14-05-34

ReeceHoffmann (approved):

PR #369, opened 2024-02-21
chore: Rework `RemoveReference` to use react-query & visually update deletion section

raastinn:
Screenshot from 2024-02-21 11-23-38


ReeceHoffmann (commented):
Code changes look good. That said testing locally it looks like clicking the delete button immediately deletes the reference with no warning. Given how big of a deal it is to replace a custom reference, it needs to pop a confirmation banner before sending the deletion request

raastinn:
Screenshot from 2024-02-21 13-14-10

ReeceHoffmann (approved):

PR #373, opened 2024-03-04
chore: rework `OTURemove` to use react-query

raastinn:
Screenshot from 2024-03-04 12-14-40


src/js/otus/components/Detail/RemoveOTU.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Displays a dialog for removing an OTU
+ */
+export default function RemoveOTU({ id, name, refId }: RemoveOTUProps) {
+    const history = useHistory();
+    const location = useLocation<{ removeOTU: boolean }>();
+    const mutation = useRemoveOTU();
+
+    const handleConfirm = useCallback(() => {
```
ReeceHoffmann: Why the usage of `useCallback`?

ReeceHoffmann (approved):

PR #374, opened 2024-03-04
fix: prevent crushed install button for reference installation

raastinn:
Screenshot from 2024-03-04 12-40-44

### DIFF
--- a/src/js/references/components/Official.tsx
+++ b/src/js/references/components/Official.tsx
@@ -41,7 +41,6 @@ export const ReferenceOfficial = ({ officialInstalled, onRemote }: ReferenceOffi
                             plant virus reference
                         </ExternalLink>
                         <span>
-                            {" "}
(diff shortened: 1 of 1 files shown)


src/js/references/components/Official.tsx
```diff
@@ -20,6 +20,10 @@ const StyledReferenceOfficial = styled(Box)`
     }
 `;
 
+const StyledButton = styled(Button)`
+    min-width: 83px;
+`;
```
ReeceHoffmann: Why not just include this in `StyledReferenceOfficial`? Also naming nitpick, should be something like `InstallButton`, generally we avoid `styled` unless it is the top level div for an otherwise usefully named component. Ex.`StyledReferenceOfficial` is a case where it is fine.

ReeceHoffmann (approved):

PR #379, opened 2024-03-06
chore: update `SampleDetailFiles` components

raastinn:


src/js/samples/components/Detail/SampleFileSizeWarning.tsx
```diff
@@ -0,0 +1,34 @@
...
+type SampleFileSizeWarningProps = {
+    sampleId: string;
+    show: boolean;
+    showLink: boolean;
+};
+
+/**
+ * Displays a warning if any sample files are under 10 megabytes
+ */
+export default function SampleFileSizeWarning({ sampleId, show, showLink }: SampleFileSizeWarningProps) {
```
ReeceHoffmann: I would be tempted to rethink the interface here. I think it is within the scope of `SampleFileSizeWarning` to be passed the file size and determine internally if it is large enough to render the warning. Also you can probably also handle showLink internally as well given that it just uses `location`

src/js/samples/components/Files/SampleReads.tsx
```diff
@@ -29,15 +35,3 @@ function SampleReads({ reads, prefix }) {
...
     );
 }
-
-function mapStateToProps(state) {
-    const { id, reads, name } = state.samples.detail;
-
-    return {
-        reads,
-        id,
-        prefix: snakeCase(name),
```
ReeceHoffmann, line 39: Was prefix never used before?

src/js/samples/components/Files/SampleReads.tsx
```diff
@@ -14,10 +14,16 @@ const SampleReadsTitle = styled.h2`
...
+type SampleReadsProps = {
+    /** A list of reads used to create the sample */
+    reads: Read[];
+};
+
+/**
+ * Displays a list of reads used to create the sample
+ */
+export default function SampleReads({ reads }: SampleReadsProps) {
+    const fileComponents = map(reads, file => <ReadItem key={file.name} {...file} />);
```
ReeceHoffmann:
```suggestion
    const fileComponents = map(reads, file => <ReadItem key={file.name} name={file.name} />);
```
I generally avoid spread syntax as it can make what is getting passed a bit less obvious. I tend to either pass only the needed props or pass the whole reference

src/js/samples/components/SampleFilesMessage.tsx
```diff
@@ -11,28 +9,14 @@ type SampleFilesMessageProps = {
...
     return showLegacy ? (
         <Alert color="orange" block>
             <p>
                 <strong>
                     Virtool now retains raw data for newly created samples instead of trimming during sample creation.
                 </strong>
             </p>
-            <p>
-                Because this is an older sample, only trimmed data is available. You can upload the original sample
-                files by dragging them onto the trimmed files they should replace.
```
ReeceHoffmann, line 24: What about something like: `
Because this is an older sample, only trimmed data is available. Recreate the sample to run analysis with untrimmed reads`

@igboyes thoughts?

ReeceHoffmann (changes requested):

src/js/samples/components/SampleFilesMessage.tsx
raastinn: Changed it to this for now

src/js/samples/components/Files/SampleReads.tsx
raastinn: No I believe it originally got passed to `ReadItem` but was never actually used

ReeceHoffmann (approved):

PR #380, opened 2024-03-06
chore: Rework `OTUGeneral` to use react-query

raastinn:


src/js/otus/components/Detail/OTUDetail.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+const OTUDetailSubtitle = styled.p`
+    font-size: ${props => props.theme.fontSize.md};
+    margin-top: 5px;
+
+    strong {
+        font-weight: ${getFontWeight("thick")};
+    }
+`;
+
```
ReeceHoffmann, line 45: docstring

src/js/otus/components/Detail/OTUDetail.tsx
```diff
@@ -0,0 +1,164 @@
...
+export function OTUDetail(props) {
```
ReeceHoffmann: Destructure props. 

Also consider adding typing. Might not make sense to spend much time on complex typing for redux code that we are going to bail on. Not blocking, will merge without typing.

src/js/otus/components/Detail/OTUDetail.tsx
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    if (props.detail === null || props.detail.id !== props.match.params.otuId) {
+        return <LoadingPlaceholder />;
+    }
+
+    const refId = props.detail.reference.id;
+    const { id, name, abbreviation } = props.detail;
+
+    let modifyOTUComponents;
```
ReeceHoffmann: bump this closer to usage. (next to `IconButtons)

src/js/otus/components/Detail/OTUDetail.tsx
```diff
@@ -0,0 +1,164 @@
...
+                    onClick={props.showEdit}
+                />
+                <IconLink
+                    key="remove-icon"
+                    color="red"
+                    name="trash"
+                    tip="Remove OTU"
+                    to={{ state: { removeOTU: true } }}
+                />
+            </>
```
ReeceHoffmann: Would this make sense as it's own component? Could factor out perms checking with the buttons as well

src/js/otus/types.ts
```diff
@@ -84,7 +84,7 @@ export type OTUMinimal = {
 /** A complete OTU */
 export type OTU = OTUMinimal & {
     isolates: Array<OTUIsolate>;
-    issues?: { [key: string]: any } | boolean;
```
ReeceHoffmann: Sadly the old typing is correct and the UI needs to be able to handle it atm:

https://github.com/virtool/virtool-core/blob/d4acce2e9c1fe5c6b9421854702bb1911b746d1a/virtool_core/models/otu.py

ReeceHoffmann (changes requested):

src/js/otus/components/Detail/OTUDetail.tsx
raastinn: Done. I think i'll save the typing for the conversion issue.

src/js/base/IconLink.tsx
```diff
@@ -14,7 +14,7 @@ type IconLinkProps = {
 export function IconLink({ color, name, replace, tip, to, ...props }: IconLinkProps) {
     return (
         <Link replace={replace} to={to} aria-label={props["aria-label"]}>
-            <Icon color={color} name={name} tip={tip} hoverable />
+            <Icon as="button" color={color} name={name} tip={tip} hoverable />
```
ReeceHoffmann: Why the switch up here? The icon has no inherent functionality here, so I my instinct here is that it is semantically incorrect to render it as a button. If your going for a look it is probably best to take advantage of classname and apply additional styles

raastinn:
Screenshot from 2024-03-11 12-03-18

src/js/otus/components/Detail/OTUHeaderEndIcons.tsx
```diff
@@ -11,6 +12,14 @@ type OTUHeaderEndIconsProps = {
     abbreviation: string;
 };
 
+const EditOTUIcon = styled(IconLink)`
+    margin: 0 5px;
+`;
+
+const RemoveOTUIcon = styled(IconLink)`
```
ReeceHoffmann: Nitpick: I would just have a single `OTUIcon` since the style changes your applying are identical

ReeceHoffmann (approved):

PR #384, opened 2024-03-07
Support path aliases

ReeceHoffmann:
Change:
 - adds path aliases for all folders in src directory to `tsconfig`, `vite.config` and `vitest.config` supporting type aliases in real and testing environments
 - Converts files is `src/account` to use type aliases


igboyes (approved):
cool. lgtm.

PR #386, opened 2024-03-12
feat(testing)!: remove Enzyme, rework `RemoveSequence` to current standards

raastinn:
Screenshot from 2024-03-12 14-33-19
Screenshot from 2024-03-12 14-27-56

Previous version showed sequence id rather than the accession


igboyes (changes requested):
Please add back the explicit imports of `screen` and remove the global defintiion. I'd like to get rid of the testing globals completely.

igboyes (approved):
LGTM, but I'll leave final say to @ReeceHoffmann since he's the expert.

ReeceHoffmann (approved):

PR #387, opened 2024-03-14
chore: Rework `IsolateEditor` to use typescript & strip redux

raastinn:


src/js/otus/hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @param otuId - The otu which the isolate belongs to
+ * @param sequences - The sequences in the active isolate
+ */
+export default function useGetSequences(otuId: string, sequences: OTUSequence) {
+    const { data, isLoading } = useFetchOTU(otuId);
+
+    if (isLoading) {
+        return { sequences: [], isLoading: true };
+    }
```
ReeceHoffmann: Couple generic comments that I think could help clean this up a lot: 

1. There isn't really a need to manually define the value for isLoading since what the query returns is always correct
2. I would treat the API of this hook as if it was a more general query. ie before data is loaded, data can be undefined a per a normal query

Combing these two I am guessing you should be able to get it down to one return, assuming that post query success sequences is at least an empty array

src/js/otus/types.ts
```diff
@@ -87,7 +87,7 @@ export type OTU = OTUMinimal & {
     issues?: { [key: string]: any } | boolean;
     last_indexed_version?: number;
     most_recent_change: HistoryNested;
-    otu_schema: Array<OTUSegment>;
+    schema: Array<OTUSegment>;
```
ReeceHoffmann: https://github.com/virtool/virtool-core/blob/main/virtool_core/models/otu.py

You check that this change makes sense? It conflicts with the models in core.

ReeceHoffmann (changes requested):

src/js/otus/hooks.ts
```diff
@@ -0,0 +1,35 @@
...
+import { indexOf, map, sortBy } from "lodash-es/lodash";
+
+/**
+ * A hook for sorting the sequences for the active isolate
+ *
+ * @param otuId - The otu which the isolate belongs to
+ * @param sequences - The sequences in the active isolate
+ */
+export default function useGetSequences(otuId: string, sequences: OTUSequence) {
+    const { data, isLoading } = useFetchOTU(otuId);
```
ReeceHoffmann, line 12: Why call `useFetchOtu`? looks to me like the data returned from it isn't getting used

ReeceHoffmann: I think it's sketchy to use a query and not expose the loading status to the calling component. the calling component has no way of knowing if there is no sequences, or if the query simply isn't cached. For this hook I would go one of two directions:

1. lean into it being a query. Take only the Ids you need and fetch the rest from the API. From the perspective of the calling component the interface of `useGetSequences` matches as closely as possible the interface of a standard query. 

2. Explicitly pass the data via props, context, ect and turn this into a simple util function that does the sorting

src/js/otus/hooks.ts
```diff
@@ -0,0 +1,38 @@
...
+                const index = indexOf(segmentNames, entry.segment);
+                return index !== -1 ? index : segmentNames.length;
+            },
+        ]);
+
+        return { data: { ...isolate, sequences: updatedSequences }, isLoading: isLoading };
+    }
+
+    return { data: { ...isolate, sequences: [] }, isLoading: isLoading };
+}
```
ReeceHoffmann: I think we need a rethink on how this is being done. While this version is better since it no longer has the implicit dependency, I think as a long term design pattern having to include extra loaders is sub-optimal. 

Taking a look at the code, I think this could be good use case for using context. This way we can avoid the excess loaders and keep any dependencies explicit.

I think the template show here is a good jumping off point for getting this done nicely: https://tkdodo.eu/blog/react-query-and-react-context

Then for the actual sorting we can just use a helper called something along the lines of `sortSequnecesBySegment(sequences, segments): sequences`

Happy to talk over a call if you want additional guidance on this one

ReeceHoffmann (approved):

PR #389, opened 2024-03-18
refactor: Reimplement `Dropdown` base components with Radix

raastinn:
Screenshot from 2024-03-18 13-54-27
Screenshot from 2024-03-18 13-54-33
Screenshot from 2024-03-18 13-54-39


ReeceHoffmann:
Dropdown for the login looks good, the others could use some tweaks:

I would personally: 

 - Have the dropdown size match the content size or the button used to open them (+ appropriate margin ofc)
 - Align the dropdown so that it lines up with the triggers left hand side

Alternatively/along with those you could also style it with a little arrow like how [radix](https://www.radix-ui.com/primitives/docs/components/dropdown-menu) does it. Could make the size difference feel a bit more natural. Not required though

ReeceHoffmann (commented):
Code wise this looks good. Worth taking a look if you haven't already for an alternative to using z-index on the dropdown content. Might not be avoidable, but nice to doge if possible. Address the visual style comment above and we can merge

ReeceHoffmann (dismissed):

ReeceHoffmann (approved):

PR #395, opened 2024-03-22
chore: upgrade react-dropzone

raastinn:


ReeceHoffmann (commented):
Latest is version 14, https://www.npmjs.com/package/react-dropzone. NPM will only update to latest version that is not a major version by default

ReeceHoffmann:
Did you verify that everything works as expected post update?

raastinn:
> Did you verify that everything works as expected post update?

Yeah the file uploads seem to be working as expected with drag and drop.

ReeceHoffmann (approved):

PR #398, opened 2024-03-27
Prevent crash when untrustworthy regions overlap

ReeceHoffmann:
Changes:
 - Prevents crash causing attempted mutation of read-only source array when untrustworthy regions overlap


igboyes (commented):
Why is the `cloneDeep` required?

igboyes (approved):

PR #399, opened 2024-03-27
refactor: remove react scroll sync

raastinn:
Screenshot from 2024-03-27 16-29-39


src/js/analyses/components/Pathoscope/PathoscopeDetail.tsx
```diff
@@ -19,20 +18,25 @@ export function PathoscopeDetail({ hit, mappedCount }: PathoscopeDetailProps) {
...
-    return (
-        <div>
-            <ScrollSync>
-                <div>{isolateComponents}</div>
-            </ScrollSync>
-        </div>
-    );
+    useEffect(() => {
+        isolateRefs.current = map(filtered, (_filter, index) => isolateRefs.current[index]);
+    }, [filtered]);
```
ReeceHoffmann: Seems like it is just going to truncate `isolateRefs` down to the length of `filtered`. Would be much simpler to do via slice if that's the goal

Related, it would be a huge boon for a usability of the hook to abstract needing to manage the refs entirely out of the calling component

src/js/analyses/components/Pathoscope/PathoscopeDetail.tsx
```diff
@@ -19,20 +18,25 @@ export function PathoscopeDetail({ hit, mappedCount }: PathoscopeDetailProps) {
...
+                key={isolate.id}
+                {...isolate}
+                reads={Math.round(isolate.pi * mappedCount)}
+                showPathoscopeReads={showReads}
+                isolateRef={element => (isolateRefs.current[index] = element)}
+            />
+        );
+    });
+
+    useScrollSync(isolateRefs.current);
```
ReeceHoffmann: nitpick, but I would bump this call up. During the first render isolateRefs will always be empty anyway

src/js/utils/hooks.ts
```diff
@@ -135,3 +135,32 @@ export function useUrlSearchParamsList(key: string, defaultValue?: string[]): [s
...
+                element.scrollTop = scrollTop;
+            });
+        }
+
+        elements.map(element => {
+            element.addEventListener("scroll", e => handleScroll(e));
+            return () => {
+                element.removeEventListener("scroll", e => handleScroll(e));
+            };
+        });
```
ReeceHoffmann: Couple things:

 1. Nothing is getting returned to useEffect, so no removal of listeners is occuring
 2. I think useEffect only supports getting a single function as the cleanup callback

src/js/utils/hooks.ts
```diff
@@ -135,3 +135,32 @@ export function useUrlSearchParamsList(key: string, defaultValue?: string[]): [s
...
+/**
+ * Hook for synchronizing scroll between components
+ *
+ * @param elements - A list of elements to apply synchronous scroll to
+ */
+export function useScrollSync(elements: HTMLElement[]) {
+    useEffect(() => {
+        if (!elements[1]) {
+            return;
+        }
```
ReeceHoffmann: `elements.length`?

src/js/analyses/components/Pathoscope/Isolate.tsx
```diff
@@ -1,13 +1,11 @@
 import { useUrlSearchParams } from "@utils/hooks";
 import { map } from "lodash-es";
 import React from "react";
-import { ScrollSyncPane } from "react-scroll-sync";
 import styled from "styled-components";
 import { toScientificNotation } from "../../../utils/utils";
 import Coverage from "./Coverage";
 
 const PathoscopeChartRibbon = styled.div`
-    white-space: nowrap;
```
ReeceHoffmann, line 10: The graphs should stay in a single line

ReeceHoffmann (changes requested):
Decent start, but the custom hooks could use a few tweaks for sure. The biggest issue for me is currently the hook feels clunky to use. Having to manage the refs should be entirely abstracted into the hook. The only thing the calling component should really have to do is pass a subscription function as `ref` to the underlying dom node. (like how `react-hook-form` works)

Happy to talk this over as it could get quite technical

ReeceHoffmann:
The current hook seems iffy. My main concern here is that as written it is trying to manage the refs without having a clean mechanism for registering and de-registering the components. There are a lot of ways we could handle this, but I think in this case the cleanest might just be to switch over to using a couple of components instead of the hook. 

One way you could do this is the following:

1. `ScrollSyncContext` Mananges context. Provides two varaibles: 1. the current scroll percentage 2. A function to call on scroll to change the current percentage

2. `ScrollSync` subscribes to context and scrolls the underlying component/calls the adjustment function as needed


```javascript

const ScrollContenxt = createContext(null)

export function ScrollSyncContext({children)){
 const [scroll, setScrollPercentage] = useState()

 function setScrollPercentage(percentage){
   //takes the scroll percentage and updates
  }


 return <ScrollContext.provider value={handlescroll, scrollPercentage}>children </ScrollContext.provider>
}


export function ScrollSync({children}){
 const {onScroll, scrollPercentage} = useContext(ScrollContext)
 const ref = useRef()

// One use effect with [] deps that sets up the onScroll

//another useeffect with [scrollPercentage] that scrolls the element when `scrollpercentage` changes

return <div ref={ref}>children</div>

}



 return children
}

```

Then you make sure that context appears somewhere above the components in the DOM tree and wrap each synced component with `ScrollSync`

The advantage of doing it this way is the refs are managed locally by the component, so registering and de-registering the components is essentially gotten for free as the components are loaded and unloaded. 

Not hard stuck on this method, but the way that refs are managed does need to be cleaned up.

ReeceHoffmann (approved):

PR #403, opened 2024-04-02
refactor: remove `react-beautiful-dnd` & reorder schema list with arrows

raastinn:
Screenshot from 2024-04-02 15-37-46
Screenshot from 2024-04-02 15-38-18

**_Note:_** removed alphabetical sorting of the schema's using `sortBy` since it messes up with the index order


src/js/otus/components/Detail/Schema/Schema.tsx
```diff
@@ -30,15 +31,46 @@ export default function Schema({ match }: SchemaProps) {
...
         return <LoadingPlaceholder />;
     }
 
     const { abbreviation, name, schema } = data;
 
+    function handleMoveUp(index: number) {
+        const updatedSchema = [...data.schema];
+        const temp = updatedSchema[index];
+        updatedSchema[index] = updatedSchema[index - 1];
+        updatedSchema[index - 1] = temp;
```
ReeceHoffmann: nitpick: I like the shorter es6 syntax,  both are fine though tbh
```suggestion
        const updatedSchema = data.schema.slice();
        [updatedSchema[index], updatedSchema[index-1]] = [updatedSchema[index-1], updatedSchema[index]] 
```

src/js/otus/components/Detail/Schema/Schema.tsx
```diff
@@ -30,15 +31,46 @@ export default function Schema({ match }: SchemaProps) {
...
+        mutation.mutate(
+            { otuId, name, abbreviation, schema: updatedSchema },
```
ReeceHoffmann: Can you get away with updating only the schema?

Also worth considering factoring the mutation out rather then duplicating it

ReeceHoffmann (changes requested):
Visually I would make the carets bigger, right now they are too small to make for nice buttons. I would also consider stacking them vertically with a bit of space between them

raastinn:
Screenshot from 2024-04-08 12-46-42

ReeceHoffmann:
Style review round 2:
 - At least double the actual icon size, they should be a comfortable size to quickly click
 - The area between the icons should not be clickable, active area should be the icon itself or a tight square around the icon
 - Should probably have a hover color to match the style of other buttons in the UI
 - consider offsetting the icons for the top and bottom item so they are in the same position they otherwise would be in

raastinn:
Screenshot from 2024-04-22 16-16-30

igboyes:
I think the side-by-side design was better.

The carets now look too similar to graphics that would normally indicate the elements can be reordered vertically by drag-drop.

That the first and last elements only have one icon is less obvious when they are closely stacked in a vertical line. Horizontally, it is obvious, and a good clue to the user for how the interface works.

raastinn:
Screenshot from 2024-04-23 11-56-31

ReeceHoffmann:
I am happy with this version. @igboyes?

igboyes:
@raastinn @ReeceHoffmann Needs conflicts resolved, then you should merge

PR #416, opened 2024-04-17
refactor: rework samples sidebar directory to use typescript

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [X] All changes are tested
 - [X] All touched code documentation is updated
 - [X] Debugging logging and commented out code has been removed

 
## Screenshots:


src/js/samples/components/Sidebar/SampleSidebarSelector.tsx
```diff
@@ -23,33 +24,53 @@ export const SampleSidebarSelectorButton = styled.div`
...
+        const { id } = item;
+
         return (
             <SampleSidebarSelectorItem
-                key={result.id}
-                selected={selectedItems.includes(result.id)}
-                partiallySelected={partiallySelectedItems.includes(result.id)}
-                {...result}
+                key={item.id}
+                selected={selectedItems.includes(id as never)}
```
ReeceHoffmann: Hmm why the usage of `never`? Either the check or typing doesn't make much sense here

ReeceHoffmann:
Current status? tsc check is still looks like it is having problems

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -219,7 +219,7 @@ export default function CreateSample() {
                     values,
                 }: {
                     errors: FormikErrors<formValues>;
-                    setFieldValue: (field: string, value: string) => void;
+                    setFieldValue: (field: string, value: string | string[] | number[]) => void;
```
ReeceHoffmann, line 222:
```suggestion
                    setFieldValue: (field: string, value:  string[] | number[]) => void;
```
Why allow a single string on it's own?

src/js/samples/components/Sidebar/SampleSidebarSelector.tsx
```diff
@@ -0,0 +1,95 @@
...
+    manageLink,
+}: SampleSidebarSelectorProps) {
+    const [results, term, setTerm] = useFuse(sampleItems, ["name"], []);
+    const sampleItemComponents = results.map((item: Label | SubtractionShortlist) => (
+        <SampleSidebarSelectorItem
+            key={item.id}
+            selected={
+                selectionType === "labels"
+                    ? (selectedItems as number[]).includes(item.id as number)
+                    : (selectedItems as string[]).includes(item.id as string)
```
ReeceHoffmann: I would give this a rework using generics. You should be able to declare a type such that is either a string or number but only one of the two. As written this introduces an`if` into the runtime,  while the types are removed at compile time.

ReeceHoffmann (changes requested):

src/js/samples/components/Create/CreateSample.tsx
raastinn: Some of the other form values like `group` use strings

ReeceHoffmann (approved):

PR #417, opened 2024-04-17
Check the api version prior to serving the client

ReeceHoffmann:
## Changes

 - Production server will now check the API version before loading and exit with code of 1 if the server can't be reached or the version is lower than required in the config file.
 - Light refactor to the API server code structure
 - Updates the PR template to request reflect the need to update the API version with most breaking changes (Updated wording)
 

## Compatiblity

changes.

Common breaking changes:

Choose one:
 - [x]  The changes break compatibility and the API version has been updated if required

PR is breaking in two distinct ways:

 1. It mandates the use of the current latest API version at time of writing (23.4.1)
 2. it requires the production server be provided a correct server URL to function

Both have been addressed in dev by: https://github.com/virtool/dev/pull/25

## Pre-Review Checklist
 - [ ] ~~All changes are tested~~ No, presently we have no testing framework for the express server. Has been tested manually

 
## Screenshots:
N/A - no visual changes

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,7 +16,7 @@ Common breaking changes:
 
 Choose one:
  - [ ] The changes do not break compatibility
- - [ ] The changes potentially break compatibility
+ - [ ] The changes break compatibility and the API version has been updated if required
 
 
 ## Pre-Review Checklist
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -16,7 +16,7 @@ Common breaking changes:
 
 Choose one:
  - [ ] The changes do not break compatibility
- - [ ] The changes break compatibility and the API version has been updated if required
+ - [ ] The changes potentially break compatibility
 
 
 ## Pre-Review Checklist
--- a/server/config.js
+++ b/server/config.js
@@ -1,4 +1,0 @@
-/** Configuration object for a production virtool UI server*/
-module.exports = {
-    minAPIVersion: "23.4.0",
-};
--- a/server/config.js
+++ b/server/config.js
@@ -1,3 +1,4 @@
+/** Configuration object for a production virtool UI server*/
 module.exports = {
     minAPIVersion: "23.4.0",
 };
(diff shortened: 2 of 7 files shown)


.github/pull_request_template.md
```diff
@@ -16,7 +16,7 @@ Common breaking changes:
 
 Choose one:
  - [ ] The changes do not break compatibility
- - [ ] The changes potentially break compatibility
+ - [ ] The changes break compatibility and the API version has been updated if required
```
igboyes: I would prefer to leave this as is and put a link a link below to docs on what to do for breaking changes in frontend.

server/config.js
```diff
@@ -0,0 +1,4 @@
+/** Configuration object for a production virtool UI server*/
+module.exports = {
+    minAPIVersion: "23.4.0",
+};
```
igboyes: * I think this should go in `package.json` under a `virtool` section.
* Change to `minimumApiVersion`

server/version.js
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Verify the running API version is compatible with the UI
+ *
+ * @func
+ * @param {string[]} minVersion - the minimum API version required
+ * @param {string} apiUrl - the base url for the API
+ * @returns {N/A}
+ */
+async function verifyAPIVersion(minVersion, apiUrl) {
```
igboyes:
```suggestion
async function verifyApiVersion(minVersion, apiUrl) {
```

server/version.js
```diff
@@ -0,0 +1,25 @@
...
+    const response = await superagent.get(apiUrl).then(res => res.body);
```
igboyes, line 18: I think you should load the `minimumApiVersion` here instead of `run.js`.

igboyes (changes requested):

server/version.js
ReeceHoffmann: Yeah that's fair. My thought process was towards loading all options in the main file, but as of now that is the only option passed via file like that. Switched loading it in `version.js`

server/config.js
ReeceHoffmann: Done, but it requires copying `package.json` into the production build image. I don't think that's a big deal, but flagging it just in case

igboyes (approved):

PR #422, opened 2024-04-29
chore: update detail directory in samples & remove all unused redux

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/samples/components/Detail/Sidebar.tsx
```diff
@@ -0,0 +1,61 @@
...
+                sampleLabels={sampleLabels.map(label => label.id)}
+            />
+            <DefaultSubtractions
+                onUpdate={subtractions => {
+                    mutation.mutate(
+                        { update: { subtractions } },
+                        {
+                            onSuccess: () => {
+                                queryClient.invalidateQueries(samplesQueryKeys.detail(sampleId));
+                            },
```
ReeceHoffmann: I think this might make more sense to be a part of the query defn in `queries.ts` since the invalidation won't happen if the user navigates away as written. Also might be worth looking if there is a WS message for this already, and if that is causing request duplication

ReeceHoffmann (approved):

PR #424, opened 2024-04-30
chore: update `RebuildAlert` & `RebuildIndex` to use react-query

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/indexes/components/RebuildIndex.tsx
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    function handleSubmit(e) {
+        e.preventDefault();
+        mutation.mutate(
+            { refId },
+            {
+                onSuccess: () => {
+                    history.push({ state: { rebuild: false } });
+                    queryClient.invalidateQueries(indexQueryKeys.infiniteLists());
```
ReeceHoffmann: I think the query key invalidation belongs as part of the `useCreateIndex` defn.

ReeceHoffmann (changes requested):
One small tweak and I think it's good!

What were your concerns regarding a compatibility break?

ReeceHoffmann (approved):

PR #428, opened 2024-05-01
fix: attribution time bug

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:
Screenshot from 2024-05-01 11-46-32

### DIFF
--- a/src/js/base/RelativeTime.tsx
+++ b/src/js/base/RelativeTime.tsx
@@ -15,12 +15,12 @@ import React, { useEffect, useState } from "react";
 function createTimeString(time) {
     const now = Date.now();
     const timeString = formatDistanceStrict(new Date(time), now, { addSuffix: true });
-    return includes(timeString, "in a") || includes(timeString, "a few") ? "just now" : timeString;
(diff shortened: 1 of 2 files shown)


src/js/base/RelativeTime.tsx
```diff
@@ -14,8 +14,9 @@ import React, { useEffect, useState } from "react";
  */
 function createTimeString(time) {
     const now = Date.now();
-    const timeString = formatDistanceStrict(new Date(time), now, { addSuffix: true });
-    return includes(timeString, "1 minute") || includes(timeString, "seconds") ? "just now" : timeString;
+    const currentTime = isAfter(new Date(time), new Date()) ? now : time;
+    const timeString = formatDistanceStrict(new Date(currentTime), now, { addSuffix: true });
+    return includes(timeString, "in a") || includes(timeString, "a few") ? "just now" : timeString;
```
igboyes: It seems like you could cut down on the number of `new Date()` calls.

igboyes (changes requested):

igboyes (approved):

PR #429, opened 2024-05-01
fix: prevent retries by default when user lacks permission

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:

### DIFF
--- a/src/js/references/components/Detail/AddReferenceGroup.tsx
+++ b/src/js/references/components/Detail/AddReferenceGroup.tsx
@@ -55,9 +55,9 @@ type AddReferenceGroupProps = {
 export default function AddReferenceGroup({ groups, onHide, refId, show }: AddReferenceGroupProps) {
     const mutation = useAddReferenceMember(refId, "group");
     const [term, setTerm] = useState("");
-    const { data, isLoading, isError, isFetchingNextPage, fetchNextPage } = useInfiniteFindGroups(25, term);
+    const { data, isLoading, isFetchingNextPage, fetchNextPage } = useInfiniteFindGroups(25, term);
 
-    if (isLoading || isError) {
+    if (isLoading) {
         return null;
     }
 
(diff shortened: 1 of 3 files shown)


src/js/references/components/Detail/AddReferenceUser.tsx
```diff
@@ -55,9 +55,9 @@ type AddReferenceUserProps = {
 export default function AddReferenceUser({ users, onHide, refId, show }: AddReferenceUserProps) {
     const mutation = useAddReferenceMember(refId, "user");
     const [term, setTerm] = useState("");
-    const { data, isLoading, isFetchingNextPage, fetchNextPage } = useInfiniteFindUsers(25, term);
+    const { data, isLoading, isError, isFetchingNextPage, fetchNextPage } = useInfiniteFindUsers(25, term);
 
-    if (isLoading) {
+    if (isLoading || isError) {
```
ReeceHoffmann: ~~Are you running into permission problems with fetching users?~~

Figured out what was going on. The fix here is going to be opening up the API to allow all users to hit that API. (fix is in CI atm)

I would actually remove this change since the 403 in this case represents a mistake in permission management rather than a real lack of permission. In general I would be careful with `isError`, as it will be true if any error occurs which could end up masking a major error

src/js/references/components/Detail/AddReferenceGroup.tsx
```diff
@@ -55,9 +55,9 @@ type AddReferenceGroupProps = {
 export default function AddReferenceGroup({ groups, onHide, refId, show }: AddReferenceGroupProps) {
     const mutation = useAddReferenceMember(refId, "group");
     const [term, setTerm] = useState("");
-    const { data, isLoading, isFetchingNextPage, fetchNextPage } = useInfiniteFindGroups(25, term);
+    const { data, isLoading, isError, isFetchingNextPage, fetchNextPage } = useInfiniteFindGroups(25, term);
```
ReeceHoffmann: I don't think I would expect this API to get 403'd either. were you running into problems?

ReeceHoffmann (approved):

PR #432, opened 2024-05-08
chore: rework API directory to use react-query

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:
Screenshot from 2024-05-14 14-40-09
Screenshot from 2024-05-14 14-40-23

Also visually updated the API keys list to match the current standards as well


ReeceHoffmann:
Just checked out a copy to see how it works. I would suggest switching the sections over to regular `BoxGroupSections`. The all blue highlighting doesn't make much sense now imo given that only the checkbox is interactable.

src/js/base/BoxGroupSectionSelect.ts
```diff
@@ -1,17 +1,16 @@
...
-type SelectBoxGroupSectionProps = {
+type SelectBoxGroupSectionProps = BoxGroupSectionProps & {
     active?: boolean;
     theme: DefaultTheme;
 };
 
 export const SelectBoxGroupSection = styled(BoxGroupSection)<SelectBoxGroupSectionProps>`
     background-color: ${props => (props.active ? props.theme.color.blue : "transparent")};
     color: ${props => (props.active ? props.theme.color.white : "inherit")};
-    cursor: pointer;
```
ReeceHoffmann: Why this change?

src/js/base/BoxGroupSectionSelect.ts
```diff
@@ -1,17 +1,16 @@
 import styled, { DefaultTheme } from "styled-components";
-import { BoxGroupSection } from "./BoxGroupSection";
+import { BoxGroupSection, BoxGroupSectionProps } from "./BoxGroupSection";
 import { CheckboxLabel } from "./CheckboxLabel";
 import { StyledCheckbox } from "./styled/StyledCheckbox";
 
-type SelectBoxGroupSectionProps = {
+type SelectBoxGroupSectionProps = BoxGroupSectionProps & {
```
ReeceHoffmann: Hmm feel like it makes more sense to either keep them separate or fuse them fully. Everything in `SelectBoxGroupsSecctionProps` is already in `BoxGroupSectionProps`

ReeceHoffmann (changes requested):

src/js/base/BoxGroupSectionSelect.ts
raastinn: Initially made this change because `BoxGroupSection` already makes the cursor a pointer when there is an `onClick` and all of the `SelectBoxGroupSection` usages had an `onClick` I believe, so I found it redundant and since I changed the `onClick` to `Checkbox` I thought t shouldn't show a pointer for the whole item just the clickable area. I'll revert the changes made to this component since we don't use it in this PR anymore

ReeceHoffmann (approved):

PR #433, opened 2024-05-09
chore: convert OTU history module to use react-query

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/otus/queries.tsx
```diff
@@ -61,7 +63,7 @@ export function useInfiniteFindOTUS(refId: string, term: string, verified?: bool
  * @returns A single OTU
  */
 export function useFetchOTU(otuId: string) {
-    return useQuery<OTU, ErrorResponse>(OTUQueryKeys.detail(otuId), () => getOTU(otuId), {
+    return useQuery<OTU, ErrorResponse>(OTUQueryKeys.details(), () => getOTU(otuId), {
```
ReeceHoffmann: This is likely to cause problems, how does the query tell the difference between request wanting different OTUs when hitting the cache?

src/js/otus/queries.tsx
```diff
@@ -71,6 +73,16 @@ export function useFetchOTU(otuId: string) {
...
 
+/**
+ * Fetches the history of changes for a single OTU
+ *
+ * @param otuId - The id of the OTU to fetch
+ * @returns A history list of changes for a single OTU
+ */
+export function useFetchOTUHistory(otuId: string) {
+    return useQuery<OTUHistory[], ErrorResponse>(OTUQueryKeys.detail(otuId), () => getOTUHistory(otuId));
+}
```
ReeceHoffmann: Should be a new query key defined to cover the case of history. I think linking it with detailed is fair enough from a cache invalidation perspective. That said you can't use the exact same key because of conflicts. 

 Consider:

```typescript
histories: () => ["OTU", "details", "history"] as const,
history: (id: string) => ["OTU", "detail", "history", id] as const,
```

ReeceHoffmann (changes requested):

src/js/otus/queries.tsx
```diff
@@ -28,6 +30,8 @@ export const OTUQueryKeys = {
     infiniteList: (filters: Array<string | number | boolean>) => ["OTU", "list", "infinite", ...filters] as const,
     details: () => ["OTU", "details"] as const,
     detail: (id: string) => ["OTU", "detail", id] as const,
+    histories: () => ["OTU", "details", "history"] as const,
+    history: (id: string) => ["OTU", "detail", "history", id] as const,
```
ReeceHoffmann: Hmm I think there is bug here: 
```suggestion
    details: () => ["OTU", "detail"] as const,
    detail: (id: string) => ["OTU", "detail", id] as const,
    histories: () => ["OTU", "detail", "history"] as const,
    history: (id: string) => ["OTU", "detail", "history", id] as const,
```
I think the pluralized keys would break the query partial query matching during invalidation

ReeceHoffmann (approved):

PR #434, opened 2024-05-10
refactor: add typing + docstrings for file upload item

jasont7:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:

### DIFF
--- a/src/js/files/components/UploadItem.tsx
+++ b/src/js/files/components/UploadItem.tsx
@@ -1,11 +1,11 @@
 import React from "react";
 import styled from "styled-components";
-import { getBorder, getColor, getFontWeight } from "../../app/theme";
+import { getBorder, getColor, getFontWeight, theme } from "../../app/theme";
 import { Icon, Loader, ProgressBarAffixed } from "../../base";
 import { byteSize } from "../../utils/utils";
 
 /**
- * Styled component for the upload item container.
+ * Container for the upload item.
  */
 const StyledUploadItem = styled.div`
     padding: 0;
@@ -17,14 +17,14 @@ const StyledUploadItem = styled.div`
 `;
 
 type UploadItemTitleProps = {
-     /** whether the upload failed. */
-     failed: boolean
+    /** Whether the upload failed */
+    failed: boolean
 }
 
 /** 
-Styled component for the title section of the upload item. 
-*/
-const UploadItemTitle = styled.div<{ failed: boolean }>`
+ * Container for the title section of the upload item.
+ */
+const UploadItemTitle = styled.div<UploadItemTitleProps>`
     justify-content: space-between;
     display: flex;
     padding: 15px 15px 10px;
@@ -35,7 +35,7 @@ const UploadItemTitle = styled.div<{ failed: boolean }>`
     }
     span:last-child {
         margin-left: auto;
-        color: ${props => (props.failed ? getColor({ theme: props.theme, color: "red" }) : "inherit")};
+        color: ${props => (props.failed ? getColor({ theme: theme, color: "red" }) : "inherit")};
     }
     i.fa-times {
         font-size: 20px;
--- a/src/js/files/components/UploadItem.tsx
+++ b/src/js/files/components/UploadItem.tsx
@@ -51,6 +51,12 @@ const UploadItemName = styled.span`
 
 /**
  * Props definition for the UploadItem component.
+ * @param name - the name of the file being uploaded.
+ * @param progress - the progress of the upload in percentage.
+ * @param size - the size of the file being uploaded.
+ * @param failed - whether the upload failed.
+ * @param localId - the local id of the file being uploaded.
+ * @param onRemove - function to remove the file from the upload list.
  */
 interface UploadItemProps {
     name: string;
@@ -63,7 +69,7 @@ interface UploadItemProps {
 
 /**
  * Component representing an individual upload item.
- * @param props The properties passed to the component.
+ * @param props - defined in the UploadItemProps interface.
  */
 export const UploadItem: React.FC<UploadItemProps> = ({ name, progress, size, failed, localId, onRemove }) => {
     let uploadIcon = progress === 100 ? <Loader size="14px" /> : <Icon name="upload" />;
--- a/src/js/files/components/UploadItem.tsx
+++ b/src/js/files/components/UploadItem.tsx
@@ -51,12 +51,12 @@ const UploadItemName = styled.span`
 
 /**
  * Props definition for the UploadItem component.
- * @param name - the name of the file being uploaded.
- * @param progress - the progress of the upload in percentage.
- * @param size - the size of the file being uploaded.
- * @param failed - whether the upload failed.
- * @param localId - the local id of the file being uploaded.
- * @param onRemove - function to remove the file from the upload list.
+ * @prop name - the name of the file being uploaded.
+ * @prop progress - the progress of the upload in percentage.
+ * @prop size - the size of the file being uploaded.
+ * @prop failed - whether the upload failed.
+ * @prop localId - the local id of the file being uploaded.
+ * @prop onRemove - function to remove the file from the upload list.
  */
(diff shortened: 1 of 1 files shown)


src/js/files/components/UploadItem.tsx
```diff
@@ -37,16 +42,35 @@ const UploadItemTitle = styled.div`
...
+}
+
+/**
+ * Component representing an individual upload item.
+ * @param props The properties passed to the component.
+ */
+export const UploadItem: React.FC<UploadItemProps> = ({ name, progress, size, failed, localId, onRemove }) => {
     let uploadIcon = progress === 100 ? <Loader size="14px" /> : <Icon name="upload" />;
-    let uploadBookend = byteSize(size, true);
+    let uploadBookend: React.ReactNode = byteSize(size, true);
```
ReeceHoffmann, line 76: Is that the most specific type that fits what you expect to be returned?

src/js/files/components/UploadItem.tsx
```diff
@@ -37,16 +42,35 @@ const UploadItemTitle = styled.div`
...
+ * Styled component for the name of the upload item.
+ */
 const UploadItemName = styled.span`
     font-weight: ${getFontWeight("thick")};
 `;
 
-export const UploadItem = ({ name, progress, size, failed, localId, onRemove }) => {
+/**
+ * Props definition for the UploadItem component.
+ */
```
ReeceHoffmann: No need for this docstring imo, the purpose quickly discernible from the name of the type in this case

src/js/files/components/UploadItem.tsx
```diff
@@ -37,16 +42,35 @@ const UploadItemTitle = styled.div`
...
+interface UploadItemProps {
+    name: string;
+    progress: number;
+    size: number;
+    failed: boolean;
+    localId: string;
+    onRemove: (localId: string) => void;
+}
```
ReeceHoffmann:
```suggestion
type UploadItemProps = {
    name: string;
    progress: number;
    size: number;
    failed: boolean;
    localId: string;
    onRemove: (localId: string) => void;
}
```

Differences in functionality between type and interface are minimal, but the two don't work together well

~~Not blocking (aka will merge without): Try adding docstrings for the props themselves. Not all of them need description, but the use case for something like `localId` may not be obvious at first look.~~

Outdated since last push. Thanks for jumping on that! That said the format we use is typically todo the docs right next to the type:

interface UploadItemProps {
/** <Relevant docstring>*/
    name: string;
    ...ect
}

src/js/files/components/UploadItem.tsx
```diff
@@ -37,16 +42,35 @@ const UploadItemTitle = styled.div`
     }
 `;
 
+/**
+ * Styled component for the name of the upload item.
+ */
```
ReeceHoffmann:
```suggestion
/**
 * Emphasized name of the upload item.
 */
```

src/js/files/components/UploadItem.tsx
```diff
@@ -15,7 +16,11 @@ const StyledUploadItem = styled.div`
     }
 `;
 
-const UploadItemTitle = styled.div`
+/**
+ * Styled component for the title section of the upload item.
+ * @param failed - whether the upload failed.
```
ReeceHoffmann: Props for styled components need to be documented in the same way as other components.  What is being passed is `props`, or similar, where failed is a member of that object. Unfortunately it makes documenting this way incorrect or very cumbersome

src/js/files/components/UploadItem.tsx
```diff
@@ -37,16 +42,35 @@ const UploadItemTitle = styled.div`
...
+    size: number;
+    failed: boolean;
+    localId: string;
+    onRemove: (localId: string) => void;
+}
+
+/**
+ * Component representing an individual upload item.
+ * @param props The properties passed to the component.
+ */
```
ReeceHoffmann:
```suggestion
/**
 * Progress tracker for a single uploaded file
 */
```
Wording is just a suggestion, two major things being done:

1. The passing of a props object is assumed, so we only really care about documenting component specific elements of props.
2. It should be clear from context that the item is a component, so the docstring should focus on what the component is/does for the user.

src/js/files/components/UploadItem.tsx
```diff
@@ -37,16 +42,35 @@ const UploadItemTitle = styled.div`
...
+export const UploadItem: React.FC<UploadItemProps> = ({ name, progress, size, failed, localId, onRemove }) => {
```
ReeceHoffmann:
```suggestion
export function UploadItem({ name, progress, size, failed, localId, onRemove }: UploadItemProps) {
```

ReeceHoffmann (changes requested):
First PR! 

Overall your usage of TS looks good! There are quite a few comments for bringing things in line with the rest of the project. 

I didn't comment on all instances, but for many of the docstring you can tighten them up significantly. For documenting components focus on what the component is used for. Implementation details like something being a `styled component` are often quickly inferred from the code itself while the use case of the component may not be. 

Same idea goes for documenting elements of props. Focus on information that is non-obvious from a quick glance at the code.

For example my rule of thumb for props is to never include the `type` in a docstring since it should already be present in the TS

ReeceHoffmann:
Looks like a push and my review happened around the same time! I updated the comments which seemed out of date, but there may still be other strangeness.

src/js/files/components/UploadItem.tsx
jasont7: Since `uploadBookend` is initially set to a string and then conditionally updated to a react element if `failed`, `React.ReactNode` captures both of these cases without specifying a new union type. Do you think a union type like `string | React.ReactElement` would be better suited here?

ReeceHoffmann: Ah fair point! I missed the re-assignment. I think what you did already is best. The new version is more specific, but not that much more and probably more difficult to grok.

src/js/files/components/UploadItem.tsx
```diff
@@ -15,7 +16,11 @@ const StyledUploadItem = styled.div`
     }
 `;
 
-const UploadItemTitle = styled.div`
+/**
+ * Styled component for the title section of the upload item.
+ * @prop failed - whether the upload failed.
+ */
+const UploadItemTitle = styled.div<{ failed: boolean }>`
```
ReeceHoffmann:
```suggestion
type UploadItemTitleProps = {
     /** whether the upload failed. */
     failed: boolean
}

/** 
Styled component for the title section of the upload item. 
*/
const UploadItemTitle = styled.div<{ failed: boolean }>`
```
Document the type, `@props` is not a part of `tsdoc` standard. https://tsdoc.org/ (afaik it is a part of jsdoc, which may be part of the confusion)

src/js/files/components/UploadItem.tsx
```diff
@@ -15,7 +16,11 @@ const StyledUploadItem = styled.div`
     }
 `;
 
-const UploadItemTitle = styled.div`
+/**
+ * Styled component for the title section of the upload item.
```
ReeceHoffmann: Please take a crack at updating these as mentioned in the main comment. You can use the new `UploadItemName` as an example. Happy to provide guidance

ReeceHoffmann (changes requested):

src/js/files/components/UploadItem.tsx
jasont7: The updated docstrings should be more aligned now, but let me know what you think

ReeceHoffmann (dismissed):

src/js/files/components/UploadItem.tsx
ReeceHoffmann: Looks good, looks like your running into a `prettier` CI problem though. Run prettier on the file loaded and it should be good to go

ReeceHoffmann (approved):

PR #436, opened 2024-05-10
refactor: add typing + docstrings for user permissions

jasont7:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/users/components/Permission.tsx
```diff
@@ -0,0 +1,26 @@
...
+    margin-right: 15px;
+`;
+
+const StyledPermissionItem = styled(BoxGroupSection)`
+    align-items: center;
+    display: flex;
+`;
+
+/**
+ * Permission item with an icon indicating its status.
```
raastinn:
```suggestion
A condensed permission item for use in a list of user permissions
```

src/js/users/components/Permission.tsx
```diff
@@ -0,0 +1,26 @@
...
+ * @prop permission - name of the permission.
+ * @prop value - indicates if the permission is granted.
+ */
+export function PermissionItem({ permission, value }: { permission: string; value: boolean }) {
```
raastinn: To keep it more consistent with the other components, define a type  `PermissionItemProps` with the permission item types and doc strings.

src/js/users/components/Permission.tsx
```diff
@@ -0,0 +1,33 @@
...
+    /* Name of the permission */
+    permission: string;
+    /* Indicates if the permission is granted */
+    value: boolean;
+};
+
+/**
+ * Permission item with an icon indicating its status.
+ * @prop permission - name of the permission.
+ * @prop value - indicates if the permission is granted.
```
ReeceHoffmann: I don't think `@props` is part of the tsdoc standard. Also information here is redundant with the information the the type

src/js/users/components/Permission.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+type PermissionItemProps = {
+    /* Name of the permission */
+    permission: string;
+    /* Indicates if the permission is granted */
+    value: boolean;
+};
+
+/**
+ * Permission item with an icon indicating its status.
```
raastinn:
```suggestion
 * A condensed permission item for use in a list of user permissions
```

Not sure why but the commit on this didnt apply

ReeceHoffmann (approved):

PR #437, opened 2024-05-10
refactor: cleaup UploadOverlay

jasont7:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/files/components/UploadTime.tsx
```diff
@@ -0,0 +1,43 @@
...
+function isDurationZero(duration: Duration): boolean {
+    return Object.values(duration).every(value => value === 0 || value === undefined);
+}
+
+/**
+ * Displays remaining time and upload speed.
+ * @prop remaining - remaining time in seconds
+ * @prop uploadSpeed - upload speed in bytes per second
+ */
+export function UploadTime({ remaining, uploadSpeed }: { remaining: number; uploadSpeed: number }): JSX.Element {
```
raastinn: I would define the type `UploadTimeProps` with the typings and docstrings to keep it consistent with the codebase

src/js/files/components/UploadOverlay.tsx
```diff
@@ -0,0 +1,103 @@
...
+    name: string;
+    /* Progress of the upload in percentage */
+    progress: number;
+    /* Size of the file in bytes */
+    size: number;
+    /* Remaining time in seconds */
+    remaining: number;
+    /* Upload speed in bytes per second */
+    uploadSpeed: number;
+};
```
raastinn: Screenshot from 2024-05-10 15-15-38
Maybe @ReeceHoffmann can confirm but this is the return type I'm getting from uploads. `overallUploadSpeed` and `totalRemainingUploadTime` return 0 or undefined for me. Aside from this, i would probably define `Uploads` in `types.ts` and define a type `UploadOverlayProps` with `uploads: Uploads[]`

jasont7: You're right, good catch. I also notice that the values for the `remaining` and `uploadSpeed` props of `UploadTime` are both undefined causing the site to break when uploading a file. If we don't need these two props then the `UploadTime` component is obsolete. @ReeceHoffmann thoughts?

ReeceHoffmann: Component works perfectly for me on main.

@raastinn Did you try using a large file or a throttled connection? If the files uploads too fast then it may never update the estimate. Here are the props passed to `UploadOverlay` for me: 
image
Could be dependent on upload progress

@jasont7 If it is always undefined that makes me thing there may be a mistake in the refactor. component works perfectly on main. Try checking out to main to see if your still seeing a crash. The component is not obsolete as it has a use case for the end user, it may be bugged though.

jasont7: I think I might know what the problem is here - I was uploading a very small file that seems to not cause the estimate to update, thus not having real values for `remaining` and `uploadSpeed`. However, in the refactor these two props are required as `number` types (as can be seen in `Upload` in `types.ts`).

src/js/files/components/__tests__/UploadOverlay.test.jsx
```diff
@@ -63,17 +69,13 @@ describe("<UploadOverlay />", () => {
...
-        expect(screen.getByText("0 MB/s")).toBeInTheDocument();
         expect(screen.queryByText("Finishing uploads")).not.toBeInTheDocument();
     });
 
     it("should render null if no uploads in progress", () => {
         props.uploads = [];
         renderWithProviders(<UploadOverlay {...props} />);
         expect(screen.queryByText("Uploads")).not.toBeInTheDocument();
-        expect(screen.queryByText("NaN min remaining")).not.toBeInTheDocument();
-        expect(screen.queryByText("0.0 MB/s")).not.toBeInTheDocument();
```
ReeceHoffmann: imo the checks for correct timing should stay in for all the tests. It ensures the components are working together as intended

src/js/files/components/UploadTime.tsx
```diff
@@ -0,0 +1,48 @@
...
+const StyledUploadInformation = styled.div`
+    display: flex;
+    justify-content: space-between;
+    font-size: ${getFontSize("md")};
+    font-weight: ${getFontWeight("normal")};
+    margin-top: 5px;
+`;
+
+function isDurationZero(duration: Duration): boolean {
+    return Object.values(duration).every(value => value === 0 || value === undefined);
```
ReeceHoffmann:
```suggestion
    return every(duration, (value) => value === 0 || value === undefined);
```
cleaner syntax and no need to iterate the duration twice via lodash

Also is undefined a valid value for members of a `Duration` object? seems like it shouldn't be

ReeceHoffmann (changes requested):
Looks mostly good! just a couple small comments

src/js/files/components/UploadTime.tsx
jasont7: Since each property of `Duration` is optional, they are implicitly undefined if they are not provided

jasont7: Hmm, the tests and prettier seem to fail after implementing this change - was there an import you added for this?

ReeceHoffmann (approved):

PR #438, opened 2024-05-13
chore: rework `ReferenceOfficial` to use react query

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/references/api.ts
```diff
@@ -34,8 +34,16 @@ export function cloneReference({ name, description, refId }): Promise<ReferenceM
         .then(res => res.body);
 }
 
-export function remoteReference({ remote_from }) {
-    return Request.post("/refs").send({ remote_from });
+/**
+ * Remotely installs a reference from the Virtool repository
+ *
+ * @param remote_from - The repository address
+ * @returns A promise resolving to remotely installing a reference
```
ReeceHoffmann: Language here feels a bit off. the `@returns` should just be a description of the returned object.

```suggestion
 * @returns A promise resolving to the installing remote reference
```

Might be an option. Make sure the ref is actually returned though. I am just assuming based of what is written here

ReeceHoffmann (commented):
One docstring comment, otherwise LGTM

ReeceHoffmann (approved):

PR #440, opened 2024-05-14
chore: update `AccountProfile` and related components

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/account/api.ts
```diff
@@ -26,8 +27,10 @@ export function get(): Promise<Response> {
  * @returns A promise resolving to a response containing the updated
  * user's account data
  */
-export function update({ update }): Promise<Response> {
-    return Request.patch("/account").send(update);
+export function updateAccount(update: string): Promise<User> {
```
ReeceHoffmann: Is that typing for update correct? I would assume that any update needs to be a key:value pair

src/js/account/components/AccountProfile.tsx
```diff
@@ -39,47 +36,37 @@ const AdministratorTag = styled(Label)`
...
  */
-function AccountProfile({ administratorRole, groups, handle }: AccountProfileProps) {
+export default function AccountProfile() {
+    const { data, isLoading } = useFetchAccount();
+
+    if (isLoading) {
+        return <LoadingPlaceholder />;
+    }
+
+    const { administrator_role, groups, handle } = data;
```
ReeceHoffmann: total nit:

Stylistic choice, pretty arguable, but i would just extract all the variables you need here. Saves having to do the `data.whatever` syntax below.

src/js/base/InputPassword.tsx
```diff
@@ -1,15 +1,20 @@
...
+};
+
+export const InputPassword = React.forwardRef<HTMLInputElement, InputPasswordProps>((props, ref) => {
     const [show, setShow] = useState(false);
 
     return (
-        <InputContainer align="right">
-            <Input name="password" aria-label="password" type={show ? "text" : "password"} {...props} />
+        <InputContainer>
+            <Input as="input" {...props} ref={ref} type={show ? "" : "password"} />
```
ReeceHoffmann, line 16: Why the `as`? Shouldn't the underlying type of `Input` be `input`?

ReeceHoffmann (changes requested):

src/js/base/InputPassword.tsx
raastinn: Not sure why this was happening, but without the `as="input"`, the input wouldn't register the text from react-hook-form.

ReeceHoffmann: Hmm interesting, is `Input` maybe not forwarding the ref to the html `input`? I think this is worth fixing, having to overwrite with `as` seems kinda brutal

ReeceHoffmann (approved):

PR #441, opened 2024-05-16
refactor: upgrade UploadToolbar

jasont7:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/files/components/Toolbar.tsx
```diff
@@ -0,0 +1,65 @@
...
+    message?: React.ReactNode;
+    /* For validating file type */
+    validationRegex?: RegExp;
+};
+
+/*
+ * Renders an UploadBar if the user has permission to upload files.
+ */
+export function UploadToolbar({ onDrop, fileType, message, validationRegex }: UploadToolbarProps) {
+    const { hasPermission: canCreate } = useCheckAdminRoleOrPermission(Permission.create_ref);
```
ReeceHoffmann: You sure it makes sense to check the ability to create a reference? I would assume the checking for upload permissions makes much more sense here

src/js/base/UploadBar.tsx
```diff
@@ -90,7 +99,7 @@ export const UploadBar = ({
     });
 
     return (
-        <StyledUploadBar active={isDragAccept} {...rootProps}>
+        <UploadBarContainer active={isDragAccept} {...rootProps}>
```
ReeceHoffmann: nitpick:

This usage of `styled` in a component name is actually okay!

Generally we avoid it but the exception is for cases like this where you have a top level container in a another component where it is used to style the main component. 

For non top-level components it's best to avoid styled because there is likely a much more descriptive name that can be used. In this case it is hard to really come up with a name that really describes what the component is doing other than styling the uploadBar.

src/js/files/components/Toolbar.tsx
```diff
@@ -0,0 +1,65 @@
...
+            </Alert>
+        );
+    }
+
+    function handleDrop(acceptedFiles: File[]) {
+        onDrop(fileType, acceptedFiles);
+    }
+
+    function validateExtensions(file: File) {
+        return validationRegex.test(file.name) ? null : { code: "Invalid file type", message: "File type not allowed" };
```
ReeceHoffmann, line 42: Is the new message getting rendered somewhere/being used? If so please upload an image of how the UI is rendering it

src/js/files/components/__tests__/Manager.test.jsx
```diff
@@ -118,7 +120,7 @@ describe("<FileManager>", () => {
                     preloadedState: state,
                 }),
         );
-        const invalidFile = new File(["test"], "test_invalid_file.gz", { type: "application/gzip" });
+        const invalidFile = new File(["test"], "test_invalid_file.gzz", { type: "application/gzip" });
```
ReeceHoffmann: Was the original not working? I think the expectation was that .gz would fail since the file type needs to have `.fa` or `.fasta` before the compression extension

src/js/files/components/__tests__/Manager.test.jsx
```diff
@@ -78,7 +78,7 @@ describe("<FileManager>", () => {
     });
 
     it("should change message if passed", async () => {
-        const account = createFakeAccount({ administrator_role: null });
+        const account = createFakeAccount({ administrator_role: "full" });
```
ReeceHoffmann: Couple notes:

1. Use the `AdministratorRoles` enum. Enums are a useful way to limit string inputs to a set number of valid options and prevent typos from creating tough to debug errors
2. Make sure there are some tests where the user is not an administrator, but still has the upload permission. It is important to make sure that non-admin users that have upload permissions are still able to upload

ReeceHoffmann (changes requested):

ReeceHoffmann:
Forgot to mention it as part of the main review but comments prefaced with `nitpick` or `nit` are non-blocking. aka I will still merge if unaddressed. Generally they will be really small things that are not a big deal, but are (in theory) minor improvements to code style

src/js/files/components/__tests__/Manager.test.jsx
jasont7: No you are right, this was left over from one of my previous attempts of fixing a failure

src/js/files/components/Toolbar.tsx
jasont7: Yes, good catch - I was basing this off of code from other components

src/js/files/components/Toolbar.tsx
jasont7: I don't think it is being used anywhere, but the following error occurs when the `message` attribute is not present in the `validator` prop of `UploadBar`:

`Property 'message' is missing in type '{ code: string; }' but required in type 'FileError'`

ReeceHoffmann (approved):

PR #445, opened 2024-05-23
chore: update sequences to use react-query

raastinn:
## Changes

- Removed all otu redux
- All components in the sequences directory are redux-free
- Created hooks with respect to deleted selectors for OTUs and Sequences
- Updated barcode and genome add/edit components (and related components) to use react-query
- Converted most touched components to typescript

## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/sequences/hooks.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const activeIsolate = useGetActiveIsolate(otu);
+    const activeSequenceId = location.state?.editSequence || undefined;
+    const sequences = sortSequencesBySegment(activeIsolate.sequences, otu.schema);
+
+    const inactiveSequences = reject(sequences, { id: activeSequenceId });
+
+    const referencedSegmentNames = compact(map(inactiveSequences, "segment"));
+    return otu.schema.filter(segment => !referencedSegmentNames.includes(segment.name));
+}
```
ReeceHoffmann: Seems like there is quite a bit of code duplication between some of these functions. Is there is a nice way to reduce this I would. Non-blocking, will merge without update.

ReeceHoffmann (approved):

PR #447, opened 2024-05-24
refactor: upgrade initial user creation form

jasont7:
## Changes

- Replace the usage of Formik with react-hook-form
- Change the file type to .tsx and update as appropriate with typing. (may be minimal to no changes)
- Update doc string to match how other components are documented

## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/wall/FirstUser.tsx
```diff
@@ -0,0 +1,95 @@
...
+                <WallLoginContainer>
+                    <WallTitle />
+                    <WallHeader>Setup First User</WallHeader>
+                    <WallSubheader>
+                        Create an administrative user that can be used to configure your new Virtool instance.
+                    </WallSubheader>
+
+                    <form onSubmit={handleSubmit(onSubmit)}>
+                        <InputGroup>
+                            <InputSimple aria-label="username" id="username" type="username" defaultValue="" />
```
raastinn: The `InputSimple` for the username doesn't register anything so you may want to double check on that one. I believe you can remove the `type`. You should also add the `InputLabel` back as it gives context for the users. To access the input in test you can just add `htmlFor` on `InputLabel`

Also you should add error handling for the register.

src/js/wall/FirstUser.tsx
```diff
@@ -0,0 +1,95 @@
...
+                                id="password"
+                                type="password"
+                                defaultValue=""
+                                {...register("password", {
+                                    required: "Password does not meet minimum length requirement (8)",
+                                    minLength: {
+                                        value: 8,
+                                        message: "Password does not meet minimum length requirement (8)",
+                                    },
+                                })}
```
raastinn: Add the `InputLabel` for context

src/js/wall/FirstUser.tsx
jasont7: I added the `InputLabel` back and updated the test, but I'm not quite sure how to add error handling for the register. Could you point me in the right direction for that?

src/js/wall/FirstUser.tsx
```diff
@@ -0,0 +1,101 @@
...
+                            />
+                            <InputError>{errors.password?.message}</InputError>
+                        </InputGroup>
+
+                        <WallButton type="submit" icon="user-plus" color="blue">
+                            Create User
+                        </WallButton>
+                        {mutation.isError && <InputError>{mutation.error.response?.body.message}</InputError>}
+                    </form>
```
raastinn, line 69: I think this part actually handles the error-handling for username and password. You can maybe double check if it does and remove the error-handling on password & username.

src/js/wall/FirstUser.tsx
```diff
@@ -0,0 +1,101 @@
...
+                                {...register("password", {
+                                    required: "Password does not meet minimum length requirement (8)",
+                                    minLength: {
+                                        value: 8,
+                                        message: "Password does not meet minimum length requirement (8)",
+                                    },
+                                })}
+                            />
+                            <InputError>{errors.password?.message}</InputError>
+                        </InputGroup>
```
raastinn: If the comment above is true, this `InputError` can probably be removed.

src/js/wall/FirstUser.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * A form for creating the first instance user
+ */
+export default function FirstUser() {
+    const mutation = useMutation<User, ErrorResponse, NewUser>(createFirst, {
+        onSuccess: () => {
+            window.location.reload();
+        },
+    });
```
raastinn: You can move this into the queries component in `src/js/users/queries`. We actually made an `ErrorResponse` in `src/js/types/types` so you can just use that for the type.

src/js/wall/FirstUser.tsx
raastinn: Yeah, I just added some comments to see if that alternative would work instead. Other than that, code is looking pretty good!

src/js/wall/FirstUser.tsx
jasont7: I don't see a queries component in `src/js/users/queries`, did you mean `useInfiniteFindUsers`?

src/js/wall/FirstUser.tsx
jasont7: Yup, you are right

src/js/wall/FirstUser.tsx
raastinn: Sorry, I meant the `queries` file. You can just move the mutation in there with all the other react-query stuff and name the function `useCreateFirstUser` and have it return the mutation. The error response can just be imported from our types as well

src/js/users/queries.ts
```diff
@@ -37,3 +39,11 @@ export function useInfiniteFindUsers(per_page: number, term: string) {
         },
     );
 }
+
+export function useCreateFirstUser() {
+    return useMutation<User, ErrorResponse, NewUser>(createFirst, {
+        onSuccess: () => {
```
raastinn:
```suggestion
export function useCreateFirstUser() {
    return useMutation<User, ErrorResponse, {handle: string, password: string, forceReset: boolean}>(({handle, password, forceReset}) => createFirst(handle, password, forceReset), {
        onSuccess: () => {
```

Everything looks good but just a few nitpicks: 
- I would do this so it can stay consistent with the other mutations we have. 
- Also you may want to destructure the props for createFirst from `({handle, password, forceReset})` to `(handle: string, password: string, forceReset: boolean)`
- Then you can probably delete the `NewUser` type from types

src/js/wall/FirstUser.tsx
```diff
@@ -0,0 +1,74 @@
+import { useCreateFirstUser } from "@/users/queries";
+import React from "react";
+import { useForm } from "react-hook-form";
+import { InputError, InputGroup, InputLabel, InputSimple } from "../base";
+import { WallButton, WallContainer, WallDialog, WallHeader, WallLoginContainer, WallSubheader } from "./Container";
+import { WallTitle } from "./WallTitle";
+
+type FormData = {
```
raastinn:
```suggestion
type FormValues = {
```

Nitpick

src/js/users/queries.ts
jasont7: I made these changes, but I also had to add the `force_reset` param to the api call in `createFirst`. Tests still pass but let me know if that won't work.

src/js/users/api.ts
```diff
@@ -6,12 +6,14 @@ import { UserResponse } from "./types";
  *
  * @param handle - The handle of the user
  * @param password - The password of the user
+ * @param forceReset - Whether the user should be forced to reset their password on next login
  * @returns A promise resolving to creating the first user
  */
-export function createFirst({ handle, password }) {
+export function createFirst({ handle, password, forceReset }) {
```
raastinn:
```suggestion
export function createFirst(handle: string, password: string, forceReset: boolean) {
```

src/js/users/queries.ts
```diff
@@ -37,3 +39,14 @@ export function useInfiniteFindUsers(per_page: number, term: string) {
         },
     );
 }
+
+export function useCreateFirstUser() {
+    return useMutation<User, ErrorResponse, { handle: string; password: string; forceReset: boolean }>(
+        ({ handle, password, forceReset }) => createFirst({ handle, password, forceReset }),
```
raastinn:
```suggestion
        ({ handle, password, forceReset }) => createFirst(handle, password, forceReset),
```

src/js/users/queries.ts
raastinn: Yeah good call.

raastinn:
Looks good to me!

PR #455, opened 2024-06-04
fix: hide reference clone button if user does not have create_ref permissions

jasont7:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [ ] All changes are tested
 - [ ] All touched code documentation is updated
 - [ ] All tests pass in your local environment
 - [ ] Deepsource issues have been reviewed and addressed
 - [ ] Debugging logging and commented out code has been removed

 
## Screenshots:

### DIFF
--- a/src/js/references/components/__tests__/ReferenceList.test.tsx
+++ b/src/js/references/components/__tests__/ReferenceList.test.tsx
@@ -27,6 +27,9 @@ describe("<ReferenceList />", () => {
     afterEach(() => nock.cleanAll());
 
     it("should render correctly", async () => {
+        const permissions = createFakePermissions({ create_ref: true });
+        const account = createFakeAccount({ permissions: permissions });
+        mockAPIGetAccount(account);
         const scope = mockApiGetReferences([references]);
         renderWithRouter(<ReferenceList />, state, history);
 
@@ -34,6 +37,8 @@ describe("<ReferenceList />", () => {
         expect(screen.getByText(references.name)).toBeInTheDocument();
         expect(screen.getByText(`${references.user.handle} created`)).toBeInTheDocument();
         expect(screen.getByText(`${references.organism} ${references.data_type}s`)).toBeInTheDocument();
+
+        expect(await screen.findByRole("link", { name: "clone" })).toBeInTheDocument();
         expect(screen.getByLabelText("clone")).toBeInTheDocument();
 
         scope.done();
@@ -113,8 +118,8 @@ describe("<ReferenceList />", () => {
             renderWithRouter(<ReferenceList />, state, history);
 
             expect(await screen.findByText("References")).toBeInTheDocument();
+            expect(await screen.findByRole("link", { name: "clone" })).toBeInTheDocument();
             await userEvent.click(screen.getByRole("link", { name: "clone" }));
-
             await userEvent.click(screen.getByRole("button", { name: "Clone" }));
 
             getReferencesScope.done();
@@ -130,8 +135,8 @@ describe("<ReferenceList />", () => {
             renderWithRouter(<ReferenceList />, state, history);
 
             expect(await screen.findByText("References")).toBeInTheDocument();
+            expect(await screen.findByRole("link", { name: "clone" })).toBeInTheDocument();
             await userEvent.click(screen.getByRole("link", { name: "clone" }));
-
             await userEvent.clear(screen.getByRole("textbox"));
             await userEvent.type(screen.getByRole("textbox"), "newName");
             await userEvent.click(screen.getByRole("button", { name: "Clone" }));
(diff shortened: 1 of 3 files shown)


src/js/references/components/__tests__/ReferenceList.test.tsx
```diff
@@ -57,9 +62,6 @@ describe("<ReferenceList />", () => {
         });
 
         it("should not render creation button when [canCreate=false]", async () => {
-            const permissions = createFakePermissions({ create_ref: true });
-            const account = createFakeAccount({ permissions: permissions });
-            mockAPIGetAccount(account);
```
ReeceHoffmann: I think we still want to fake the account, just make sure the faked acc doesn't have the request perm. 

**Non-Blocking**: As written this test is a bit iffy since it can't tell the difference between the data not having loaded and the user not having perms. This might be fixable quickly be ensuring the API mocks have been called before checking for the button. Might still end up with a race condition though. Will merge without this being resolved as it is a pre-existing issue

src/js/references/components/__tests__/ReferenceList.test.tsx
```diff
@@ -109,22 +115,25 @@ describe("<ReferenceList />", () => {
...
+            const permissions = createFakePermissions({ create_ref: true });
+            const account = createFakeAccount({ permissions: permissions });
+            mockAPIGetAccount(account);
             const getReferencesScope = mockApiGetReferences([references]);
             const cloneReferenceScope = mockApiCloneReference("newName", `Cloned from ${references.name}`, references);
             renderWithRouter(<ReferenceList />, state, history);
 
             expect(await screen.findByText("References")).toBeInTheDocument();
+            expect(await screen.findByRole("link", { name: "clone" })).toBeInTheDocument();
             await userEvent.click(screen.getByRole("link", { name: "clone" }));
```
ReeceHoffmann:
```suggestion
            await userEvent.click(await screen.findByRole("link", { name: "clone" }));
```
afaik you can merge these two. If the component does not exists then this should throw an error

src/js/references/components/__tests__/ReferenceList.test.tsx
```diff
@@ -134,15 +143,17 @@ describe("<ReferenceList />", () => {
...
         it("should display an error when name input is cleared", async () => {
+            const permissions = createFakePermissions({ create_ref: true });
+            const account = createFakeAccount({ permissions: permissions });
+            mockAPIGetAccount(account);
             const scope = mockApiGetReferences([references]);
             renderWithRouter(<ReferenceList />, state, history);
 
             expect(await screen.findByText("References")).toBeInTheDocument();
+            expect(await screen.findByRole("link", { name: "clone" })).toBeInTheDocument();
             await userEvent.click(screen.getByRole("link", { name: "clone" }));
```
ReeceHoffmann: merge if possible

src/js/references/components/__tests__/ReferenceList.test.tsx
```diff
@@ -134,15 +143,17 @@ describe("<ReferenceList />", () => {
...
-
-            await userEvent.clear(screen.getByRole("textbox"));
             await userEvent.click(screen.getByRole("button", { name: "Clone" }));
-
+            await userEvent.clear(screen.getByRole("textbox"));
             expect(screen.getByText("Required Field")).toBeInTheDocument();
```
ReeceHoffmann: Hmmmmm something seems strange here. In theory the textbox clear should have to come before the clone button is clicked. Otherwise I would expect the component happily accept the default name and never render an error. This combined with the other comment might make for a good new issue.

src/js/references/components/__tests__/ReferenceList.test.tsx
```diff
@@ -109,22 +115,25 @@ describe("<ReferenceList />", () => {
             renderWithRouter(<ReferenceList />, state, history);
 
             expect(await screen.findByText("References")).toBeInTheDocument();
+            expect(await screen.findByRole("link", { name: "clone" })).toBeInTheDocument();
             await userEvent.click(screen.getByRole("link", { name: "clone" }));
```
ReeceHoffmann: merge

src/js/references/components/__tests__/ReferenceList.test.tsx
```diff
@@ -27,13 +27,18 @@ describe("<ReferenceList />", () => {
     afterEach(() => nock.cleanAll());
 
     it("should render correctly", async () => {
+        const permissions = createFakePermissions({ create_ref: true });
+        const account = createFakeAccount({ permissions: permissions });
+        mockAPIGetAccount(account);
```
ReeceHoffmann, line 31: I think adding the account here might be making the: `should not render creation button when [canCreate=true]` test redundant. I would remove the other test unless I am missing something here.

ReeceHoffmann (changes requested):
Core code changes are good. Some nitpicks in the testing. There are some pre-existing broader problems with a couple of the tests that should be addressed, but may make more sense in a later PR depending on the difficulty of getting it right.

src/js/references/components/__tests__/ReferenceList.test.tsx
jasont7: Yes, good idea

src/js/references/components/__tests__/ReferenceList.test.tsx
jasont7: You are right, the order of these two lines should be swapped. This was my mistake in the refactor. Good catch!

src/js/references/components/__tests__/ReferenceList.test.tsx
jasont7: Yes, you are right

src/js/references/components/__tests__/ReferenceList.test.tsx
jasont7: I agree; added back the fake acc with no create perms

igboyes (approved):

PR #457, opened 2024-06-07
fix: convert 'active' to transient prop in WorkflowFilter

jasont7:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [ ] All changes are tested
 - [ ] All touched code documentation is updated
 - [ ] All tests pass in your local environment
 - [ ] Deepsource issues have been reviewed and addressed
 - [ ] Debugging logging and commented out code has been removed

 
## Screenshots:


igboyes (approved):

igboyes (changes requested):
An idea:
* Don't pass `active` to the style component at all.
* Rely only on `aria-pressed` for variable styling.
---
Please also add a type for the props for this component.

igboyes (approved):

PR #463, opened 2024-06-12
refactor: replace formik with rhf

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


igboyes (dismissed):

src/js/references/components/CloneReference.tsx
```diff
@@ -54,57 +48,57 @@ type CloneReferenceProps = {
...
+    const reference = find(references, { id: locationState?.cloneReference || "" });
 
-    const reference = find(references, { id: history.location.state?.["id"] || "" });
-
-    function onHide() {
-        history.push({ state: { cloneReference: false } });
-    }
+    useEffect(() => {
+        reference && setValue("name", `Clone of ${reference.name}`);
+    }, [reference]);
```
ReeceHoffmann, line 63: Can you use `DefaultValues` param instead? Might not work if the form never fully unloads between different references

raastinn: That doesn't work since the default value isn't set until clone reference is clicked. So its initial default value would just be blank

ReeceHoffmann (approved):

PR #464, opened 2024-06-12
chore: update label to use tailwind css

raastinn:
## Changes

- updated `Label` component to use Tailwind
- Removed `theme` prop from `getBadgeOrLabelColor` since it wasn't necessary
- Fixed tailwind styling not being applied
- Removed spacing prop and did the styling in parent component

## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/index.html
```diff
@@ -30,6 +30,7 @@
 
         <link rel="shortcut icon" href="/images/favicon.ico" type="image/x-icon" />
         <script type="module" src="./js/index.tsx"></script>
+        <script src="https://cdn.tailwindcss.com"></script>
```
igboyes: This shouldn't be necessary.

The styles and logic should be bundled in the app based off your Tailwind installation and config.

Did you follow the Vite guide? https://tailwindcss.com/docs/guides/vite

src/js/account/components/AccountProfile.tsx
```diff
@@ -56,9 +48,13 @@ export default function AccountProfile() {
                     <h3>
                         {handle}
                         {administrator_role && (
-                            <AdministratorTag key="administrator" color="purple">
+                            <Label
+                                className={`capitalize text-[${theme.fontSize.md}] ml-auto`}
```
igboyes: Tailwind is the theme.
```suggestion
                                className="capitalize text-md ml-auto"
```

src/js/base/Label.tsx
```diff
@@ -0,0 +1,26 @@
...
+    return (
+        <span
+            className={`text-[${theme.color.white}] rounded-[${theme.borderRadius.sm}] bg-[${getBadgeOrLabelColor({
+                color,
+            })}] inline whitespace-nowrap text-center font-bold px-2 py-1 text-[${
+                theme.fontSize.sm
+            }] ${className} align-baseline last-of-type:m-0 `}
+        >
+            {children}
+        </span>
```
igboyes: * Just use the Tailwind classes like `text-white`, `rounded-sm` directly.
* Use something like `cslx` or `classnames` packages for conditional styles. Try `cslx` first if one isn't already installed.

src/js/base/NotFound.tsx
```diff
@@ -27,7 +24,9 @@ type NotFoundProps = {
 export function NotFound({ status = 404, message = "Not found" }: NotFoundProps) {
     return (
         <StyledNotFound>
-            <Label color="red">{status}</Label>
+            <Label className={`text-[${theme.fontSize.xxl}]`} color="red">
```
igboyes: Tailwind is (will be) our theme.

Make sure the `text-` suffix here is similar to what it was before in appearance. Like, you might have to use `xl` instead.
```suggestion
            <Label className="text-xxl" color="red">
```

igboyes (changes requested):

src/index.html
raastinn: Yeah, I followed the guide but the styling wouldn't apply. It turns out the create sample file selector styling was because of the styled version of `CompactScrollList`. I tried some other methods too but this one was the only one that seemed to apply the styling changes

raastinn:
- I moved our colours to the tailwind theme configs since its much more simplified than the theme used by tailwind (they use bg-red-50, bg-red-100, bg-red 200, etc) whereas now its just simplified to `bg-red`, `bg-red-light`.

- Also removed the `color` prop from the `Label` component if we're going full tailwind.

Let me know your thoughts

igboyes:
@raastinn

* Restore `color` prop for `Label`. It will set `className` based on the color variant (eg. "red") provided.
* Reset Tailwind theming to default.
* You can use my changes or drop them and work off yours. I started doing this because I needed it done for something else I was doing. Then, I got pulled into something else and this is stagnant.

Thanks!

src/js/analyses/components/Iimi/IimiDetectionTag.tsx
```diff
@@ -1,6 +1,6 @@
+import { Label } from "@base";
 import React from "react";
-import { Label } from "../../../base";
 
 export function IimiDetectionTag({ result }) {
-    return result ? <Label color="red">Detected</Label> : <Label color="grey">Undetected</Label>;
+    return result ? <Label color="red">Detected</Label> : <Label className="bg-grey">Undetected</Label>;
```
igboyes: Can't this just be like below since grey is the default?:
```suggestion
    return result ? <Label color="red">Detected</Label> : <Label>Undetected</Label>;
```

src/js/hmm/components/HMMDetail.tsx
```diff
@@ -120,7 +116,9 @@ export default function HMMDetail({ match }: HMMDetailProps) {
 
                         <tr>
                             <th>Best Definitions</th>
-                            <td>{names}</td>
+                            <td>
+                                <div className="flex gap-1 leading-none">{names}</div>
+                            </td>
```
igboyes: Does this not work?
```suggestion
                            <td className="flex gap-1 leading-none">
                                {names}
                            </td>
```

igboyes (changes requested):
Very close.

src/js/hmm/components/HMMDetail.tsx
raastinn: That doesn't work but I just applied the styling to the label instead

igboyes (approved):

PR #467, opened 2024-06-17
Add rhf persistence

ReeceHoffmann:
## Changes

 - Add new `createPersistentForm` hook
   - New hook is essentially a wrapper around RHF `createForm` and should be sued in place of it when persistence is desired
 - adjusts RestoredAlert component to be more independent
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


igboyes (approved):
LGTM

PR #477, opened 2024-06-28
refactor: replace formik with rhf in accession

raastinn:
## Changes

- Used `RHF` context in order to avoid prop drilling
- Converted all sequence components to typescript
- Used the new `usePersistForm` hook for forms
- Updated testing & deleted redundant tests

## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/sequences/components/Accession.tsx
```diff
@@ -50,19 +57,18 @@ export const Accession = ({ error }) => {
...
-    };
-
     return (
         <InputGroup>
             <InputLabel htmlFor="accession">Accession (ID)</InputLabel>
             <InputContainer align="right">
-                <Field name="accession" as={Input} error={error} id="accession" onChange={onChange} />
+                <InputSimple
+                    id="accession"
+                    onChange={() => setNotFound(false)}
```
ReeceHoffmann: Does this work as expected? `register` returns an `onChange` which I suspect would override the one manually defined. Looks like register does allow you to pass an onChange to be invoked https://react-hook-form.com/docs/useform/register. Just need to ensure that it is called alongside rather than instead of the built in `onChange`

src/js/sequences/components/Barcode/BarcodeSequenceForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+    /** A list of unreferenced targets */
+    targets: ReferenceTarget[];
+};
+
+/**
+ * A form for creating or editing a barcode sequence
+ */
+export default function BarcodeSequenceForm({ activeSequence, noun, onSubmit, targets }: BarcodeSequenceProps) {
+    const { accession, definition, host, id, sequence, target } = activeSequence || {};
+    const methods = usePersistentForm<FormValues>({
```
ReeceHoffmann, line 35: Are you accounting for the possibility of a target no longer being valid between loads somewhere?

src/js/sequences/components/Barcode/BarcodeSequenceForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+        formName: `${noun}BarcodeSequence${id}`,
+        defaultValues: {
+            target: noun === "add" ? targets[0]?.name || null : target || null,
```
ReeceHoffmann:
```suggestion
            target: target || targets[0]?.name || null 
```
I am making the assumption here that `target` will always be null if `noun == "add"`

src/js/sequences/components/IsolateSequences.tsx
```diff
@@ -35,7 +45,12 @@ export default function IsolateSequences({ activeIsolate, otuId }) {
     const sequences = sortSequencesBySegment(activeIsolate.sequences, otu.schema);
 
     const Sequence = data_type === "barcode" ? BarcodeSequence : GenomeSequence;
-    let sequenceComponents = map(sequences, sequence => <Sequence key={sequence.id} {...sequence} />);
+    let sequenceComponents: JSX.Element | JSX.Element[] = map(sequences, sequence => (
```
ReeceHoffmann: Shouldn't this always be an array of JSX elements, just in some cases an empty or array of one?

ReeceHoffmann (commented):
Looks good, just a few small comments

src/js/sequences/components/IsolateSequences.tsx
raastinn: That was my initial thought too but we redefine it later as a single element

``` if (!sequenceComponents.length) {
        if (data_type === "barcode" && !hasTargets) {
            sequenceComponents = (
                <NoneFoundSection noun="targets">
                    <Link to={`/refs/${id}/manage`}>Create one</Link>
                </NoneFoundSection>
            );
        } else {
            sequenceComponents = <NoneFoundSection noun="sequences" />;
        }
    }

igboyes: @raastinn Make changes as necessary to simplify the typing.

igboyes:
@raastinn Resolve conflicts and I'll merge.

src/js/sequences/components/Barcode/BarcodeSequenceForm.tsx
raastinn: @igboyes this isn't resolved yet have to figure out a way to fix this then push the fix. However other comments and conflicts are resolved now

PR #478, opened 2024-07-02
chore: make login request via react query

jasont7:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:


src/js/wall/LoginForm.tsx
```diff
@@ -30,23 +29,32 @@ const LoginContainer = styled.div`
 type LoginFormProps = {
     /** Error message for the login process. */
     error: string;
-    /** Callback to handle login submission */
-    onLogin: (username: string, password: string, remember: boolean) => void;
 };
 
 /** Handles the user login process. */
-export function LoginForm({ error, onLogin }: LoginFormProps) {
+export function LoginForm({ error }: LoginFormProps) {
```
ReeceHoffmann:
```suggestion
export default function LoginForm({ error }: LoginFormProps) {
```
Without needing to connect to redux, we can directyl export this component as default

Also does it still make sense to get the error from props? The redux implementation derives the text of the error from the message returned from the API. I think it would make sense to do the same thing here in the mutation. Good chance react-query has some nice support for doing this. If not you can use `onError` to handle it yourself

src/js/wall/LoginForm.tsx
```diff
@@ -30,23 +29,32 @@ const LoginContainer = styled.div`
...
+    const dispatch = useDispatch();
+    const loginMutation = useLoginMutation();
+
+    function onSubmit({ username, password, remember }) {
+        loginMutation.mutate(
+            { username, password, remember },
+            {
+                onSuccess: data => {
+                    dispatch({ type: LOGIN.SUCCEEDED, payload: data });
+                },
```
ReeceHoffmann, line 41: Not suitable action to do the same thing? I am fine with this if not. 

Our general redux pattern is to define actions, but given that this will be removed shortly I will merge as is

src/js/wall/LoginWall.tsx
```diff
@@ -14,7 +14,9 @@ export function LoginWall({ reset }: LoginWallProps) {
     return (
         <WallContainer>
             <WallDialog>
-                <WallLoginContainer>{reset ? <ResetForm /> : <LoginForm />}</WallLoginContainer>
+                <WallLoginContainer>
+                    {reset ? <ResetForm /> : <LoginForm error="Invalid username or password" />}
```
ReeceHoffmann:
```suggestion
                    {reset ? <ResetForm /> : <LoginForm  />}
```

Generally speaking errors will either be defined by the component itself (ex. form validation) or sent from the API. 

Since this is an API error the message should be derived from the API whenever possible. In cases where it is not possible the error should be specified at the lowest level of the component tree that makes sense. 

In this case I think you can simply get the error message from the API response.

src/js/wall/Queries.ts
```diff
@@ -0,0 +1,30 @@
...
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import { Response } from "superagent";
+
+/**
+ * Factory object for generating account query keys
+ */
+export const accountKeys = {
+    all: () => ["account"],
+    details: () => ["account", "details"] as const,
+};
```
ReeceHoffmann: Better to import the pre-existing `accountKeys` factory function from `account/queries`. Prevents potential drift that could occur over time

src/js/wall/__tests__/LoginForm.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    beforeEach(() => {
+        mockDispatch = vi.fn();
+        (useDispatch as jest.Mock).mockReturnValue(mockDispatch);
+
+        mockLoginMutation = {
+            mutate: vi.fn(),
+            isError: false,
+        };
+        (useLoginMutation as jest.Mock).mockReturnValue(mockLoginMutation);
```
ReeceHoffmann: While this works okay, the preference is to mock via `nock`. nock intercepts API requests, allowing for more of the request lifecycle to be simulated and therefore tests the queries themselves as well.

ReeceHoffmann (changes requested):
Generally looks good, just a few small changes

src/js/wall/LoginForm.tsx
```diff
@@ -41,13 +36,15 @@ export function LoginForm({ error }: LoginFormProps) {
         loginMutation.mutate(
             { username, password, remember },
             {
-                onSuccess: data => {
-                    dispatch({ type: LOGIN.SUCCEEDED, payload: data });
+                onSuccess: () => {
+                    dispatch(loginSucceeded());
```
ReeceHoffmann: Does this still work correctly with the password reset flow? I think this `loginSucceeded` makes the assumption that password reset will never be required. I think you would need to define a new action to get fully correct behavior. I am good with either your old version or creating a new action. Since the fix is so temporary the main objective imo is having the UI continue to work seamlessly on all versions

src/js/wall/__tests__/LoginForm.test.tsx
```diff
@@ -1,74 +1,78 @@
...
     it("should display error message on login failure", async () => {
-        mockLoginMutation.isError = true;
+        const username = "test_Username";
+        const password = "Password";
+        const errorMessage = "An error occurred during login";
 
-        renderWithProviders(<LoginForm error="Invalid username or password" />);
+        const scope = nock("http://localhost")
+            .post("/api/login", { username: username, password: password, remember: false })
+            .reply(400);
```
ReeceHoffmann: I would emulate returning the error message via the API. Makes for a less brittle test, (won't break if the hardcoded error messages changes) and tests the functionality we care more about

src/js/wall/__tests__/LoginForm.test.tsx
```diff
@@ -1,74 +1,78 @@
...
-        await userEvent.type(screen.getByLabelText("Password"), "Password");
+        renderWithProviders(<LoginForm />);
+
+        await userEvent.type(screen.getByLabelText("Username"), username);
+        await userEvent.type(screen.getByLabelText("Password"), password);
         await userEvent.click(screen.getByRole("button", { name: "Login" }));
 
-        await waitFor(() => expect(screen.getByText("Invalid username or password")).toBeInTheDocument());
+        await waitFor(() => {
+            expect(screen.getByText(errorMessage)).toBeInTheDocument();
```
ReeceHoffmann: the `FindBy` family of async DOM queries combines the get by and wait for in a convenient package. https://testing-library.com/docs/dom-testing-library/api-async/

ReeceHoffmann (changes requested):

src/js/wall/LoginForm.tsx
jasont7: Good catch. I was unsure whether the payload was being utilized but you are right. I'll keep my old version instead of creating a new action since it is temporary and so that it doesn't get confused with the existing action.

src/js/wall/__tests__/LoginForm.test.tsx
jasont7: I was thinking the same thing, but when I return a response object here (i.e., `.reply(400, { message: "Error message" })`), I can only access the message via `response.data.message` (instead of `response.body.message` which is the way we currently do it). Since the `data` attribute doesn't exist in the `Resonse` type, this causes an error when I try to access it in the login form. Do you have any suggestions for working around this?

ReeceHoffmann: So I did a bit of testing and was able to access the error message no problem at `error.response.body.message` with the following .reply configuration `.reply(400, { message: "test Error" });` 

While doing it I noticed a couple problems with your mocks that are probably what is giving you trouble. Will add those as additional comments for clarity sake

src/js/wall/__tests__/LoginForm.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    afterEach(() => nock.cleanAll());
+
+    it("should call mutate() with correct values when submitted", async () => {
+        const username = "test_Username";
+        const password = "Password";
+
+        const scope = nock("http://localhost")
+            .post("/api/login", { username: username, password: password, remember: true })
```
ReeceHoffmann:
```suggestion
            .post("/api/account/login", { username: username, password: password, remember: true })
```

src/js/wall/__tests__/LoginForm.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+        await userEvent.type(passwordField, password);
+        expect(passwordField).toHaveValue(password);
+
+        expect(screen.getByLabelText("Remember Me")).toHaveAttribute("data-state", "unchecked");
+        await userEvent.click(screen.getByLabelText("Remember Me"));
+        expect(screen.getByLabelText("Remember Me")).toHaveAttribute("data-state", "checked");
+
+        await userEvent.click(screen.getByRole("button", { name: "Login" }));
+
+        scope.isDone();
```
ReeceHoffmann:
```suggestion
        scope.done();
```
Is done only returns a bool indicating if it is done, `scope.done()` raises an error if the mocks are not satisfied. You will (almost) always want to use `scope.done()` since it will let you know if your mocks were actually used. Less essential when your mocks are only for data fetching, but really important for when calling the API is the end goal.

src/js/wall/__tests__/LoginForm.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    it("should display error message on login failure", async () => {
+        const username = "test_Username";
+        const password = "Password";
+        const errorMessage = "An error occurred during login";
+
+        const scope = nock("http://localhost")
+            .post("/api/login", { username: username, password: password, remember: false })
```
ReeceHoffmann:
```suggestion
            .post("/api/account/login", { username: username, password: password })
```
Corrects API route and removes `remember`. Superagent removes values that are undefined from the request. Another option is to ensure that remembered is always `true` or `false`

src/js/wall/__tests__/LoginForm.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+        await userEvent.type(screen.getByLabelText("Username"), username);
+        await userEvent.type(screen.getByLabelText("Password"), password);
+        await userEvent.click(screen.getByRole("button", { name: "Login" }));
+
+        await waitFor(() => {
+            expect(screen.getByText(errorMessage)).toBeInTheDocument();
+        });
+
+        scope.isDone();
```
ReeceHoffmann:
```suggestion
        scope.done();
```
Same as above, apply anywhere else that I may have missed

src/js/wall/__tests__/LoginForm.test.tsx
jasont7: Yes, those changes fixed the problem. Thank you!

ReeceHoffmann (approved):

PR #479, opened 2024-07-02
refactor: rework create sample to use react hook form

raastinn:
## Compatiblity

changes.

Common breaking changes:

Choose one:

## Pre-Review Checklist
 - [x] All changes are tested

 
## Screenshots:

### DIFF
--- a/src/js/samples/api.ts
+++ b/src/js/samples/api.ts
@@ -1,29 +1,5 @@
-import { Request } from "../app/request";
+import { Request } from "@app/request";
 import { Sample } from "./types";
-
-export const find = ({ term, labels, workflows, page = 1 }) => {
-    const request = Request.get("/samples").query({
-        page,
-    });
-
-    if (term) {
-        request.query({ find: term });
-    }
-
-    if (workflows) {
-        request.query({ workflows });
-    }
-
-    if (labels) {
-        labels.forEach(label => request.query({ label }));
-    }
-
-    request.sortQuery();
-
-    return request;
-};
-
-export const get = ({ sampleId }) => Request.get(`/samples/${sampleId}`);
 
 export const update = ({ sampleId, update }) => Request.patch(`/samples/${sampleId}`).send(update);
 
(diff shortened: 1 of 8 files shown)


src/js/samples/api.ts
```diff
@@ -67,6 +52,46 @@ export function getSample(sampleId: string): Promise<Sample> {
     return Request.get(`/samples/${sampleId}`).then(res => res.body);
 }
 
+/**
+ * Creates a sample
+ *
+ * @param name - The sample name
+ * @param isolate - The isolate
+ * @param host -
+ * @param locale -
```
ReeceHoffmann:
```suggestion
 * @param host - the name of the host species the virus was identified in
 * @param locale - the geographical location of the virus detection
```

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -132,187 +103,180 @@ type formValues = {
...
+    const {
+        control,
+        formState: { errors },
+        hasRestored,
+        handleSubmit,
+        register,
+        reset,
+        setValue,
+        watch,
+    } = usePersistentForm<FormValues>({
```
ReeceHoffmann, line 127: Are you handling the possibility of a sample file no longer being valid between loads somewhere?

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -132,187 +103,180 @@ type formValues = {
...
+                labels: [],
+                subtractionIds: [],
+            },
         },
     });
+    const mutation = useCreateSample();
+
+    useEffect(() => {
+        setValue("group", toString(account?.primary_group?.id));
+    }, [account]);
```
ReeceHoffmann, line 147: Why this instead of setting it as a `defaultValue`?

src/tests/fake/samples.tsx
```diff
@@ -240,7 +240,7 @@ export function mockApiCreateSample(
             subtractions,
             group,
         })
-        .reply(201, sample);
+        .reply(200, sample);
```
ReeceHoffmann: Is this change based off what the API is currently returning? In theory the API should be returning 201, not sure if in practice it is though

ReeceHoffmann (changes requested):

src/js/samples/components/Create/CreateSample.tsx
raastinn: Yeah the form deselects the file if it is deleted between loads.

src/js/samples/components/Create/CreateSample.tsx
raastinn: Since we're setting it before we ensure that `Account` is finished loading, the default value would always just be null.

PR #493, opened 2024-07-22
feat: show warning when viewing reads in pathoscope

raastinn:
## Changes
- Added `Alert` to show reads warning when showing reads in pathoscope view

## Pre-Review Checklist
* [x] All changes are tested.
 
## Screenshots

Screenshot from 2024-07-22 13-46-09


igboyes (changes requested):
* Use `<p>` instead of `<ul>`. This is too condensed.
* Use  a header element like `h3` and check if the spacing is better. This is also a block element unlike `strong`, so no wrapping div will be necessary.
* I don't think any of these nested divs are necessary

raastinn:
Screenshot from 2024-07-24 11-38-37

`h2` or `h3` didn't look good since the spacing was off. Just put it into a `p` and made it bold.

The current `div` that is there is for ensuring that its displayed in a column rather than inline

PR #520, opened 2024-07-29
refactor: reimplement color base component in tailwind

jasont7:
## Pre-Review Checklist
* [x] All changes are tested.
 
## Screenshots
Screenshot from 2024-08-01 14-45-55

### DIFF
--- a/src/js/base/Color.tsx
+++ b/src/js/base/Color.tsx
@@ -2,36 +2,51 @@ import { cn } from "@utils/utils";
 import React, { useCallback } from "react";
 import { Input } from "./Input";
 
-const colors = [
-    { name: "Gray 1", class: "bg-gray-200" },
-    { name: "Gray 2", class: "bg-gray-500" },
-    { name: "Gray 3", class: "bg-gray-700" },
-    { name: "Red 1", class: "bg-red-200" },
-    { name: "Red 2", class: "bg-red-500" },
-    { name: "Red 3", class: "bg-red-700" },
-    { name: "Yellow 1", class: "bg-yellow-300" },
-    { name: "Yellow 2", class: "bg-yellow-500" },
-    { name: "Yellow 3", class: "bg-yellow-700" },
-    { name: "Green 1", class: "bg-green-300" },
-    { name: "Green 2", class: "bg-green-500" },
-    { name: "Green 3", class: "bg-green-700" },
-    { name: "Blue 1", class: "bg-blue-300" },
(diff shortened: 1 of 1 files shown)


src/js/base/Color.tsx
```diff
@@ -46,79 +43,57 @@ const colors: string[] = [
...
-    margin-top: 10px;
-`;
+    return (
+        <button
+            type="button"
+            title={color}
+            onClick={handleClick}
+            className={
+                "flex-1 h-full transition-transform hover:-translate-y-0.5 focus:outline-none focus:ring-2 focus:ring-white focus:ring-offset-2 focus:ring-offset-gray-300 focus:z-10 first:rounded-l-sm last:rounded-r-sm"
+            }
```
raastinn: Try and use the `cn` function from our utils files. Similar to how its done on the `Label` base component

src/js/base/Color.tsx
```diff
@@ -46,79 +43,57 @@ const colors: string[] = [
...
+            style={{ backgroundColor: `#${color}` }}
```
raastinn: Put colour into classNames and keep it as names rather than hex format. This is also done in `Label` so you can use that as a template

src/js/base/Color.tsx
```diff
@@ -46,79 +43,57 @@ const colors: string[] = [
...
+        <div>
             <Input
                 id={id}
                 value={value}
                 onChange={(e: React.ChangeEvent<HTMLInputElement>) => onChange(e.target.value)}
             />
-            <ColorGrid>
+            <div className="flex h-9 mt-2.5">
                 {colors.map(color => (
                     <ColorSquare key={color} color={`#${color}`} onClick={onChange} />
```
raastinn, line 106: This should just have to be `color={color}` after changes

src/js/base/Color.tsx
jasont7: Should I still keep a mapping between tailwind colors and the hex codes? When I replace this with just the names as in the `Label` component, the API calls fail because the format of the color code is invalid (i.e., the `Color` component works fine with this change, but there is a downstream failure in the `LabelForm`).

jasont7: I pushed this change so you can see the error too

igboyes: The color has to be hex in this case. This is a clearer solution than putting it in a CSS class.
```js
style={{ backgroundColor: `#${color}` }}
```

ReeceHoffmann (approved):

PR #522, opened 2024-07-29
refactor: create subtraction file selector not accessible

raastinn:
## Changes
- Also uninstalled `redux` packages that got re-installed after removal
- Updated the `BoxGroupSectionSelect` base component to handle keyboard accessibility 

## Pre-Review Checklist
* [x] All changes are tested.
 
## Screenshots


igboyes (changes requested):
I haven't checked this out locally, but I can't see how these changes would resolve the issue.

Maybe I should have been more specific. The selector cannot take focus through the keyboard. It is not accessible.

Try focusing on the **Name** field. Then, use the tab key to navigate through the accessible form elements. You will notice that the selector element and the file items within it cannot be focussed with tab.

Looks at `Select` or `ToggleGroup` in Radis for the behaviour we need (try tabbing through the examples). It is okay to completely change the component, like moving to a select box solution.

raastinn:
> I haven't checked this out locally, but I can't see how these changes would resolve the issue.
> 
> Maybe I should have been more specific. The selector cannot take focus through the keyboard. It is not accessible.
> 
> Try focusing on the **Name** field. Then, use the tab key to navigate through the accessible form elements. You will notice that the selector element and the file items within it cannot be focussed with tab.
> 
> Looks at `Select` or `ToggleGroup` in Radis for the behaviour we need (try tabbing through the examples). It is okay to completely change the component, like moving to a select box solution.

That makes more sense. My current solution is just working off the component we already have. If you think Radix would be better I can change to that.

ReeceHoffmann:
I checked the branch out and imo it does address the basic problem as described. I susepct assigning the `role=button` is what is making it work. The said would the `listbox` and `option` labels for the main list and each individual option make more sense?

Using Radix or Shadcn is also fine with me, none of the components are really drop-in though. Both the `combobox` and the `select` are really built around using a drop down menu. @raastinn If you haven't alreayd double check me on this.

Once we are happy that the aria-label is correct I am happy to merge and we can decide about conversion to a drop-down in another PR

src/js/base/BoxGroupSectionSelect.tsx
```diff
@@ -30,4 +31,29 @@ export const SelectBoxGroupSection = styled(BoxGroupSection)<SelectBoxGroupSecti
...
+};
+
+export function BoxGroupSectionSelect({ active, children, className, onClick }: BoxGroupSectionSelectProps) {
+    function handleKeyDown(e) {
+        if (e.key === "Enter") {
+            onClick();
+        }
+    }
+    return (
+        <SelectBoxGroupSection
```
ReeceHoffmann: nitpick
```suggestion
export function SelectBoxGroupSection({ active, children, className, onClick }: BoxGroupSectionSelectProps) {
    function handleKeyDown(e) {
        if (e.key === "Enter") {
            onClick();
        }
    }
    return (
        <StyledSelectBoxGroupSection
```

ReeceHoffmann (approved):

PR #528, opened 2024-08-06
feat: install shadcn and reimplment pagination

raastinn:
## Pre-Review Checklist
* [x] All changes are tested.
 
## Screenshots
Screenshot from 2024-08-14 12-17-56


src/js/app/globals.css
```diff
@@ -0,0 +1,42 @@
...
+        --destructive-foreground: 210 40% 98%;
+        --border: 214.3 31.8% 91.4%;
+        --input: 214.3 31.8% 91.4%;
+        --ring: 222.2 84% 4.9%;
+        --radius: 0.5rem;
+        --chart-1: 12 76% 61%;
+        --chart-2: 173 58% 39%;
+        --chart-3: 197 37% 24%;
+        --chart-4: 43 74% 66%;
+        --chart-5: 27 87% 67%;
```
ReeceHoffmann: Are these colors correct? Seem like they are shadcn defaults rather than virtool colors

ReeceHoffmann (changes requested):
Overall looks pretty good. Couple things:

1. Noted in another comment, But I would take a closer look at the default css styles being defined by shadcn. I don't think they really match the current style. Would be worth looking at if those defaults defns are actually helpful.
2. The file structure of the `pagination.tsx` does not match convention. Each component should be its own file in the folder located at `base/pagination`. I would also move the `pagination.tsx` into this folder.

ReeceHoffmann (approved):

PR #529, opened 2024-08-06
fix: ensure only one index can be selected for analysis creation

raastinn:
## Pre-Review Checklist
* [x] All changes are tested.
 
## Screenshots

Screenshot from 2024-08-13 14-40-44


ReeceHoffmann:
This soln could easily cause confusion for the end user since the look of the element visually implies that ability to select more than one index. Then after one is selected the user is locked out from selecting another with no indication on why. 

I think the soln here is to switch to an element that inherently implies the ability to only select one. I think a `combobox` could work well here since it retains the ability to filter, while only implying a single element can be selected.

ReeceHoffmann (approved):

PR #555, opened 2024-08-29
chore: upgrade administrator to v6 api

jasont7:
## Pre-Review Checklist
* [x] All changes are tested.
* [ ] All tests pass in my local environment
 
## Screenshots


jasont7:
Tests currently failing:

`CreateUser`: seems to be a path resolution issue as the test html indicates a blank page. This is strange because a similar render in the `UserDetail` test does not give the same issue.

`UserDetail`: Only `UserGroups` section is failing - I suspect unrelated to path issues as before since other components of the `UserDetail` test are fine.

ReeceHoffmann:
Just looking at the annotations GH is showing me, all of the errors look like they are getting stuck with a loader. Generally that is nock mocking problem. I would look for changes that could result in the mocked and used URLs having a mismatch?

PR #570, opened 2024-09-26
Use wouter

ReeceHoffmann:
**Changes**
- Replaces react router with wouter
   - replaces all uses of `location` `state` with urlSearchParams
   - Updates tests as required to support the conversion

## Pre-Review Checklist
* [x] All changes are tested.
 
## Screenshots


src/js/account/components/API/APIKeys.tsx
```diff
@@ -1,17 +1,19 @@
...
 import styled from "styled-components";
+import { Link, useSearch } from "wouter";
 import APIKey from "./APIKey";
 import CreateAPIKey from "./CreateAPIKey";
 
 const APIKeysHeader = styled(Box)`
     align-items: center;
     display: flex;
     font-size: ${getFontSize("lg")};
+    font-size: ${getFontSize("lg")}
```
igboyes: ?

src/js/account/components/API/__tests__/APIKeys.test.tsx
```diff
@@ -57,7 +57,10 @@ describe("<APIKeys />", () => {
         });
 
         it("should render correctly when newKey = empty", async () => {
-            renderWithMemoryRouter(<APIKeys />, [{ state: { createAPIKey: true } }]);
+            renderWithMemoryRouter(<APIKeys />, "/");
+
+            await userEvent.click(await screen.findByRole("link", { name: "Create" }));
+            await new Promise(r => setTimeout(r, 1000));
```
igboyes: What is this doing?

tailwind.config.js
```diff
@@ -3,6 +3,15 @@ export default {
     prefix: "",
     theme: {
         extend: {
+            colors: {
+                primary: {
+                    DEFAULT: "#3C8786",
+                    dark: "#2B6E74",
+                    darkest: "#1E5661",
+                    light: "#6AB7AF",
+                    lightest: "#E8F5F5",
```
igboyes, line 12: Are the values other than `DEFAULT` actually used?

src/js/account/components/API/__tests__/APIKeys.test.tsx
```diff
@@ -70,7 +73,7 @@ describe("<APIKeys />", () => {
 
         it("should render correctly when newKey is set", async () => {
             const scope = mockApiCreateAPIKey("test", createFakePermissions({ remove_job: true }));
-            renderWithMemoryRouter(<APIKeys />, [{ state: { createAPIKey: true } }]);
+            renderWithMemoryRouter(<APIKeys />, "?openCreateKey=true");
```
igboyes: * Memory router is the only router used in tests (right?) so I am not clear on the benefit of renaming this function and all of its usages. There is no `renderWithBrowserRouter`.
* I would put the `?` bit in `renderWithMemoryRouter` so it doesn't need to be included in every call.
* Hardcoding the path or search parameters in every test seems fragile to me. Would using `formatSearchParams` make more sense.

EDIT: I realized after that `renderWithMemoryRouter` just takes a path, including search parameters. This is a bummer in terms of having to include the `?` everywhere.

src/js/account/components/API/CreateAPIKey.tsx
```diff
@@ -60,12 +60,11 @@ type FormValues = {
  * Displays a dialog to create an API key
  */
 export default function CreateAPIKey() {
-    const navigate = useNavigate();
-    const location = useLocation();
     const [newKey, setNewKey] = useState("");
     const [copied, setCopied] = useState(false);
     const [showCreated, setShowCreated] = useState(false);
     const mutation = useCreateAPIKey();
+    const [openCreateKey, setOpenCreateKey] = useUrlSearchParams("openCreateKey");
```
igboyes: I think we might be better served by a hook dedicated to dialogs.

```ts
const [open, setOpen] = useDialogParam("CreateKey");
```

* `open` will always be a boolean. No need for converting types outside the hook.
* Calling `setOpen` with `false` drops the key from the search parameters.

It would be nice not to have the name be `useDialogSearchParam`, but if the term param is used to mean something else in other places, the distinction will have to be made.

src/js/administration/components/administrators/Administrators.tsx
```diff
@@ -10,8 +11,9 @@ const renderRow = roles => item => <AdministratorItem key={item.id} user={item}
 
 export const ManageAdministrators = () => {
     const [term, setTerm] = useState("");
+    const [pageParam] = useUrlSearchParams("page");
 
-    const page = parseInt(new URLSearchParams(window.location.search).get("page")) || 1;
+    const page = parseInt(pageParam) || 1;
```
igboyes, line 16: Since the `page` search parameter is a common pattern, I'd dedicate a hook to it that returns an integer and handles the default case as well.

src/js/analyses/components/AnalysisDetail.tsx
```diff
@@ -28,14 +28,9 @@ const UnsupportedAnalysis = styled(Box)`
...
-type AnalysisDetailProps = {
-    /** Match object containing path information */
-    match: match<{ analysisId: string; sampleId: string }>;
-};
-
 /** Base component viewing all supported analysis */
-export default function AnalysisDetail({ match }: AnalysisDetailProps) {
-    const { analysisId, sampleId } = match.params;
+export default function AnalysisDetail() {
+    const { analysisId, sampleId } = useParams<{ analysisId: string; sampleId: string }>();
```
igboyes: OK...

This wouter API is kind of making the decision that "params" means search parameters in our codebase. If we agree this is going to be the case, fine. If not, we should wrap this hook thinly with `useSearchParams`.

src/js/analyses/components/AnalysisList.tsx
```diff
@@ -3,31 +3,26 @@ import { ContainerNarrow, LoadingPlaceholder, NoneFoundBox, Pagination } from "@
...
-
 /**
  * A list of analyses with filtering options
  */
-export default function AnalysesList({ match }: AnalysisListProps) {
-    const sampleId = match.params.sampleId;
-    const [urlPage] = useUrlSearchParams<number>("page");
+export default function AnalysesList() {
+    const { sampleId } = useParams<{ sampleId: string }>();
+    const [urlPage] = useUrlSearchParams("page");
```
igboyes: Elsewhere, you called this `pageParam`.

src/js/analyses/components/Create/CreateAnalysis.tsx
```diff
@@ -63,30 +61,29 @@ export default function CreateAnalysis({ hmms, sampleId }: CreateAnalysisProps)
         createAnalysis.mutate({ refId, subtractionIds: subtractions, sampleId, workflow, mlModel });
     }
 
-    function onChangeWorkflow(workflow: Workflows) {
-        history.push({ state: { createAnalysis: workflow } });
+    function onOpenChange(open) {
+        setOpenCreateAnalysis(open);
+        if (!open) {
+            setWorkflow("");
```
igboyes: Does this remove the key from the search parameters?
I think it might be good to differentiate and empty string from non-existence. Maybe an unset callback.

src/js/analyses/components/Create/__tests__/CreateAnalyis.test.tsx
```diff
@@ -39,10 +37,9 @@ describe("getCompatibleWorkflows()", () => {
         const mlModel = createFakeMLModel();
         mockApiGetModels([mlModel]);
 
-        renderWithProviders(
-            <MemoryRouter initialEntries={[{ state: { createAnalysis: Workflows.pathoscope_bowtie } }]}>
-                <CreateAnalysis {...props} />
-            </MemoryRouter>,
+        renderWithMemoryRouter(
+            <CreateAnalysis {...props} />,
+            `/samples/${sample.id}/analyses?openCreateAnalysis=true&workflow=pathoscope_bowtie`,
```
igboyes: As mentioned above, these hardcoded pathnames are going to be fragile. If not the existing URL utilities you wrote, maybe a function in `describe`?

src/js/analyses/components/NuVs/NuVsToolbar.tsx
```diff
@@ -10,22 +10,28 @@ import { AnalysisViewerSort } from "../Viewer/Sort";
  * Displays a toolbar for managing and filtering NuVs
  */
 export default function NuVsToolbar({ analysisId, results, sampleName }: NuVsExportProps) {
-    const [filterORFs, setFilterORFs] = useUrlSearchParams<boolean>("filterOrfs", true);
-    const [filterSequences, setFilterSequences] = useUrlSearchParams<boolean>("filterSequences", true);
-    const [find, setFind] = useUrlSearchParams<string>("find", "");
-    const [sortKey, setSortKey] = useUrlSearchParams<string>("sort", "length");
+    const [filterORFs, setFilterORFs] = useUrlSearchParams("filterOrfs", "true");
```
igboyes: I would really like to handle the type conversions automatically... Converting wherever this hook is used is a leaky abstraction.

src/js/analyses/components/NuVs/NuVsViewer.tsx
```diff
@@ -1,8 +1,8 @@
 import { NuVsList } from "@/analyses/components/NuVs/NuVsList";
+import NuVsToolbar from "@/analyses/components/NuVs/NuVsToolbar";
```
igboyes, line 2: Can this be?
```suggestion
import NuVsToolbar from "@analyses/components/NuVs/NuVsToolbar";
```

src/js/analyses/components/Pathoscope/Isolate.tsx
```diff
@@ -39,7 +39,7 @@ const StyledPathoscopeIsolateWeight = styled.strong`
 `;
 
 export function PathoscopeIsolateWeight({ pi, reads }) {
-    const [showReads] = useUrlSearchParams<boolean>("reads");
+    const [showReads] = useUrlSearchParams("reads");
```
igboyes: Wait... is this a boolean?!

src/js/app/App.tsx
```diff
@@ -1,11 +1,10 @@
 import { LoadingPlaceholder } from "@/base";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
 import { useAuthentication, useRootQuery } from "@wall/queries";
-import { History } from "history";
 import React, { Suspense } from "react";
-import { Router } from "react-router-dom";
-import { CompatRouter } from "react-router-dom-v5-compat";
 import { ThemeProvider } from "styled-components";
+import { Router as Wouter } from "wouter";
```
igboyes: Why is the `as` necessary here?

src/js/app/Main.tsx
```diff
@@ -7,9 +7,8 @@ import { useQueryClient } from "@tanstack/react-query";
 import { includes } from "lodash-es";
 import React, { lazy, Suspense, useEffect } from "react";
 import { Helmet, HelmetProvider } from "react-helmet-async";
-import { Redirect, Route, Switch } from "react-router-dom";
-import { CompatRoute } from "react-router-dom-v5-compat";
 import styled from "styled-components";
+import { Redirect, Route as WouterRoute, Switch as WouterSwitch } from "wouter";
```
igboyes: ?

src/js/base/pagination/__tests__/Pagination.test.tsx
```diff
@@ -18,11 +17,7 @@ describe("<Pagination />", () => {
     });
 
     it("Should render correctly when pageCount=6 and currentPage=1", () => {
-        renderWithProviders(
-            <MemoryRouter initialEntries={[{ pathname: "/samples/files", search: "?page=1" }]}>
-                <Pagination {...props} />
-            </MemoryRouter>,
-        );
+        renderWithMemoryRouter(<Pagination {...props} />, "/samples/files?page=1");
```
igboyes: The pathname should be a `const` in the `describe` scope.

src/js/utils/hooks.tsx
```diff
@@ -33,85 +30,99 @@ export function useElementSize<T extends HTMLElement>(): [React.MutableRefObject
...
     }
 
-    firstRender.current = false;
+    return search;
+}
+
+function createUseUrlSearchParams() {
+    const cache = { search: "" };
+
+    return function useURLSearchParams(key: string, defaultValue?: string): [string, (newValue: string) => void] {
```
igboyes:
```suggestion
    return function useUrlSearchParams(key: string, defaultValue?: string): [string, (newValue: string) => void] {
```

src/js/sequences/hooks.ts
```diff
@@ -36,15 +35,14 @@ export function useExpanded() {
  * @returns The active sequence
  */
 export function useGetActiveSequence() {
-    const location = useLocation<LocationType>();
+    const [openEditSequence] = useUrlSearchParams("openEditSequence");
```
igboyes: The key `openEditSequence` breaks with the pretty consistent convention elsewhere that the value will be a boolean.
I think a name like `editSequenceId` might be better even though it has the downside that we now have a dialog opening in response to a search param key other then `open...`.

src/js/sequences/components/Sequence/SequenceButtons.tsx
```diff
@@ -33,7 +32,8 @@ const SequenceHeaderButtons = styled.span`
  * Displays icons for the sequence item to close, edit, or remove
  */
 export default function SequenceButtons({ id, onCollapse }) {
-    const [locationState, setLocationState] = useLocationState();
+    const [, setOpenEditSequence] = useUrlSearchParams("openEditSequence");
+    const [, setOpenRemoveSequence] = useUrlSearchParams("removeSequence");
```
igboyes: Why the naming difference here?

src/js/sequences/components/EditSequence.tsx
```diff
@@ -1,31 +1,34 @@
...
-import { useLocation } from "react-router-dom";
 import EditBarcodeSequence from "./Barcode/EditBarcodeSequence";
 import EditGenomeSequence from "./Genome/EditGenomeSequence";
 
 /**
  * A component to manage the editing of sequences
  */
 export default function EditSequence() {
-    const location = useLocation<LocationType>();
+    const [activeIsolate] = useUrlSearchParams("activeIsolate");
```
igboyes: I think we can just call this param `isolate` or `isolateId`. Whatever it is just needs to be consistent with similar situations like `editSequence` (or `editSequenceId`).

---
I think this hook should be called `useSearchParam`.

1. There are no search params other than those in the URL, so `Url` is no needed.
2. The hooks only handles one search param key, so it should be singular.

I also think we should have a thin wrapper around `useParams` from router, called `useSearchParams` so our naming is consistent and we avoid sprinkling `wouter` imports and function calls all over our codebase.

src/js/sequences/components/AddSequenceLink.tsx
```diff
@@ -48,11 +47,7 @@ export default function AddSequenceLink({ dataType, refId }: AddSequenceLinkProp
             }
         }
 
-        return (
-            <StyledAddSequenceLink onClick={() => setLocationState(merge(locationState, { addSequence: true }))}>
-                Add Sequence
-            </StyledAddSequenceLink>
-        );
+        return <StyledAddSequenceLink onClick={() => setOpenAddSequence("true")}>Add Sequence</StyledAddSequenceLink>;
```
igboyes, line 50: Can this just be a plain link without an `onclick` handler?

src/js/samples/components/Create/CreateSample.tsx
```diff
@@ -107,6 +108,8 @@ type FormValues = {
  * A form for creating a sample
  */
 export default function CreateSample() {
+    const [, navigate] = useLocation();
```
igboyes, line 111: What do you think of using our own hook `useNavigate` to handle this? This will avoid importing `wouter` and provide a more focussed API.

src/js/references/components/ReferenceToolbar.tsx
```diff
@@ -8,14 +8,14 @@ import React from "react";
...
-    const [term, setTerm] = useUrlSearchParams<string>("find", "");
+    const [term, setTerm] = useUrlSearchParams("find", "");
     const { hasPermission: canCreate } = useCheckAdminRoleOrPermission(Permission.create_ref);
 
     return (
         <Toolbar>
             <InputSearch placeholder="Reference name" value={term} onChange={e => setTerm(e.target.value)} />
             {canCreate && (
-                <LinkButton to={{ state: { createReference: true, emptyReference: true } }} color="blue">
+                <LinkButton to="?createReference=true&createReferenceType=empty" color="blue">
```
igboyes, line 18: I think `formatSearchParams` would be useful here.

src/js/references/components/ReferenceToolbar.tsx
```diff
@@ -8,14 +8,14 @@ import React from "react";
  * A toolbar which allows the references to be filtered by name
  */
 export default function ReferenceToolbar() {
-    const [term, setTerm] = useUrlSearchParams<string>("find", "");
+    const [term, setTerm] = useUrlSearchParams("find", "");
```
igboyes: Damn, I'd really like to fix the discord between `term` and `find`. This PR or another?

src/js/references/components/CreateReference.tsx
```diff
@@ -9,30 +9,37 @@ import { ImportReference } from "./ImportReference";
  * The create reference view with options to create an empty reference or import a reference
  */
 export function CreateReference() {
-    const [locationState, setLocationState] = useLocationState();
+    const [createReference, setCreateReference] = useUrlSearchParams("createReference");
```
igboyes: It might be more robust to get down to just one key triggering the visibility and type of creation.

1. Only have `createReferenceType`, which opens the dialog when set and also contains the type.
2. Have one dialog key per type: `openImportReference`, `openRemoteReference`, etc.

src/js/otus/components/OTUToolbar.tsx
```diff
@@ -18,15 +19,16 @@ type OTUToolbarProps = {
...
     const { hasPermission: canCreate } = useCheckReferenceRight(refId, ReferenceRight.modify_otu);
+    const [, setOpenCreateOtu] = useUrlSearchParams("openCreateOTU");
 
     return (
         <Toolbar>
             <InputSearch placeholder="Name or abbreviation" value={term} onChange={onChange} />
 
             {canCreate && !remotesFrom && (
-                <LinkButton to={{ state: { createOTU: true } }} color="blue" replace>
+                <Button onClick={() => setOpenCreateOtu("true")} color="blue">
```
igboyes, line 29: Shouldn't this remain a link?

src/js/indexes/components/__tests__/Indexes.test.tsx
```diff
@@ -1,17 +1,17 @@
...
-import { createFakeAccount, mockApiGetAccount } from "@tests/fake/account";
-import { renderWithMemoryRouter } from "@tests/setup";
+import { createFakeSettings, mockApiGetSettings } from "@tests/fake/admin";
 import nock from "nock";
 import React from "react";
 import { beforeEach, describe, expect, it } from "vitest";
 import { createFakeIndexMinimal, mockApiFindIndexes, mockApiGetUnbuiltChanges } from "../../../../tests/fake/indexes";
 import { createFakeReference, mockApiGetReferenceDetail } from "../../../../tests/fake/references";
-import Indexes from "../Indexes";
+import { renderWithMemoryRouter } from "../../../../tests/setup";
```
igboyes: `@tests`

igboyes (changes requested):
* Suggest writing a `useDialogSearchParam` or `useDialogParam` hook for the specialized case.
* Suggest writing a `usePageSearchParam` hook for the specialized case.
* Can we look at automatic type conversions in the hook instead of having them in many render functions. Most cases would be solved with the two above specialized hooks.
*  The `useUrlSearchParams` hook only actually handles one key, and it might be better named `useUrlSearchParam`.
* I think it might be a good idea to use a `useUrlSearchParams` wrapper wherever the wouter `useParams` hook is currently used.
* We should wrap the Wouter `Link` component and apply Tailwind classes there. This will avoid importing wouter in our components. I am not clear on how `Link` is currently being styled.
* Avoid repetitively hardcoding pathnames in tests. If this changes in the future, it is easier to change one variable.

Overall, it would be beneficial to avoid importing and using Wouter directly wherever possible to limit the blast radius if we have to replace it in the future.

src/js/account/components/API/__tests__/APIKeys.test.tsx
ReeceHoffmann: leftover debugging cruft, was looking for a race condition. Gone next push

tailwind.config.js
ReeceHoffmann: Only `DEFAULT` and `darkest` atm. Probably worth it as a near short term project to create and covert to a tailwind style pallet for primary before this gets used too widely. 

I used the old color choices now to prevent incongruity between the brightness levels where the `primary` versions are used

igboyes: Is the darker variant only used in the side nav bar? We should just change those to blue IMO.

src/js/account/components/API/__tests__/APIKeys.test.tsx
ReeceHoffmann: Fair point on the naming. Prior to this PR browser router was being used for some tests but that was fixed.

re the other points: Another option here could be a RR style abstraction that expects an object for search params and path rather than a string literal. I worry about it being a bit clunky though: couple options:

1. 
```javascript
 renderWithMemoryRouter(ui: ReactNode, path: string, searchParams: object)

```
called like
```javascript
renderWithMemoryRouter(<APIKeys />, "/", {CreateAPIKey: false});
```
or

2.

```javascript
 renderWithMemoryRouter(ui: ReactNode, location: {path: string, searchParams: Object})

```
called like
```javascript
renderWithMemoryRouter(<APIKeys />, {path: "/", searchParams: {CreateAPIKey: false}});

```
In either case the memoryRouter component will convert the passed parameters into at URL. imo version 2 is more verbose than it needs to be given that we can only show path and search params in the URL

src/js/analyses/components/NuVs/NuVsViewer.tsx
ReeceHoffmann: Huh I would have though so, but on testing it seems like the answer is no. Might be because NuvsViewer is part of `analyses`, though I don't know why that would matter

src/js/analyses/components/Pathoscope/Isolate.tsx
ReeceHoffmann: Yeah, it is just used to toggle if reads are shown as weight or "count".

src/js/app/App.tsx
ReeceHoffmann: It isn't, will be fixed

PR #655, opened 2025-08-15
Reduce load times when sorting and filtering iimi results

ReeceHoffmann:
- Prevent unnecessary re-rendering of Iimi overview graphs
  - Reduces initial load time by preventing complete rerenders of the graphs during load
  - Greatly reduces time for sorting graphs by eliminating the need to re-render the graphs
  - Improve performance when changing filtering criteria by only needing to render newly displayed graphs
- fix small type

## Checklist

Think before checking the following items:

- [x] I have considered backwards and forwards compatibility.
- [x] All changes are tested.
- [x] All touched code documentation is updated.

## Screenshots


igboyes:
I don't think this commit type will result in a release.

PR #936, opened 2026-05-07
feat(references): add archive and unarchive UI

ReeceHoffmann:
[## Summary

- adds Archive/Unarchive controls on the reference detail header and a confirmation dialog with 409 handling for the official reference
- references list gains an Active/Archived toggle (driven by the `?archived=` search param) and an `Archived` outline badge on archived rows
- archived references show an `Archive` icon tile, muted title, outline badge, and a visible Unarchive button on the detail header; Edit pencil is hidden, not just disabled
- archived state is propagated into nested views: OTU header gets the same treatment with an archive glyph in the breadcrumb, and pencil/trash are removed from the OTU header
- removes (not disables) every mutating affordance the API would 409 on while archived: Add Isolate, isolate edit/star/trash, Create Sequence, sequence pencil/trash, Add Segment, schema edit/remove/reorder, OTU history revert, Source Types editing, Remote install/refresh, index rebuild
- the corresponding mutation dialogs refuse to open via search params when archived, so deep links and back-button navigation can't reach a 409
- Source Types and Settings explicitly label the asymmetry (`read-only — archived` vs `rights still editable`) so the half-disabled Settings tab reads as intentional
- includes two unrelated drive-bys: login redirect now returns to the originally requested page, and the users list response maps `documents` → `items` so the table renders again](https://virtool.local/refs?archived=false&find=&page=1)

References list view:

image

Archived list: 

image

Detailed View:

image

Archived Detailed view:

image

Indexes:

image

Settings:

image

Detailed OTU view:

image

detailed otu schema:

image

Detailed otu history

image


igboyes (changes requested):
Try to keep message minimal and consistent. I think the "Archived" label is sufficient in most places.

If something like the rebuilt button is hidden when archived, it is not necessary to show that it is disabled when archiving. This pattern happens a lot. I think this is an obvious Claude-ism.

Move label in reference list items to far right. That was it is consistently, horizontally positioned and doesn't compete with the reference name for horizontal space.

Image

Self-evident.

Image

Worthwhile, but should be capitalized. I don't think the icon adds anything. There are other places where more wordy lines saying the same thing are used. Use this short message there to be consistent. For example:

Image

PR #1065, opened 2026-06-09
fix: surface real errors on primary route data instead of spinning

igboyes:
## Summary

- Adds `RouteError` as the router's `defaultErrorComponent`, replacing indefinite loading spinners with actionable error states — 401/403/404 get specific messages, everything else gets a \"Try again\" button that clears the error and reruns the loader
- Converts primary-data queries across samples, references, OTUs, settings, and user detail to `useSuspenseQuery`, letting the framework handle loading and errors declaratively and removing all `if (isPending || !data)` guards
- Documents the two-tier loading/error policy (Tier 1: suspense + route error boundary for primary data; Tier 2: inline `isError` check for secondary data) in `docs/queries.md` and `AGENTS.md`

### DIFF
--- a/apps/web/src/administration/queries.ts
+++ b/apps/web/src/administration/queries.ts
@@ -13,6 +13,7 @@ import {
 	useMutation,
 	useQuery,
 	useQueryClient,
+	useSuspenseQuery,
 } from "@tanstack/react-query";
 import type { ErrorResponse } from "@/types/api";
 import type { AdministratorRoleName, Settings } from "./types";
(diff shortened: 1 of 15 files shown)


igboyes:
Thanks for the reviews! Addressed in 066824b9. Rundown:

**`server/users/functions.ts` — `throw` consistency (CodeRabbit) ✅ Done**
Added `throw` to the other four `rethrowAsHttp(err)` sites (146/170/186/204) so the whole file matches the `getUser` handler.

**`getUser` — could this `throw undefined`? (Sourcery) — No, it's safe**
`rethrowAsHttp` is typed `(err: unknown): never` and every branch ends in a `throw` (the fallthrough is `throw err`). It never returns, so the inner throw always fires before the outer `throw` expression is evaluated — `throw undefined` is unreachable. The `throw` exists purely so the handler's *inferred* return type stays `User` instead of `User | undefined` (a bare call statement doesn't terminate control-flow inference under `createServerFn().handler()`), which is what lets the suspense consumer treat `data` as non-nullable.

**`administration/queries.ts` & `refs|samples/settings.tsx` — arrow → method syntax (CodeRabbit) — Skipping**
Every `queryFn` and route `loader` in the codebase uses the arrow form (e.g. `sampleQueryOptions`, `userQueryOptions`, `referenceQueryOptions`; `samples/$sampleId` and `users/$userId` loaders). Converting only these would diverge from the established pattern, and `pnpm check` (Biome, the enforced gate) passes as-is. Keeping it consistent.

**OTU history loader — narrow `error` before `.response.status` (Sourcery / CodeRabbit) — Keeping as-is**
This matches the sibling loaders in the same route group (`refs/$refId/route.tsx`, `refs/$refId/otus/$otuId/route.tsx`), and CodeRabbit notes nothing enforces unknown-catch narrowing. Introducing an `isHttpError` guard here would be inconsistent with its neighbours; standardising error narrowing across all loaders is a worthwhile but separate cross-cutting cleanup.

PR #1086, opened 2026-06-29
feat: remove iimi analysis viewer and workflow option

igboyes:
## Summary
- Removes the remaining iimi frontend code now that the backend no longer returns iimi analyses: the analysis results viewer (`Iimi/` components and its iimi-only charts), iimi types, formatting utilities, fixtures, and the iimi sort options.
- Drops `iimi` from the workflow enums and display-name maps (`jobs/types.ts`, `app/utils.ts`, `AnalysisWorkflow`), so it no longer appears as a supported workflow.
- Completes VIR-2551, following the earlier removal of iimi creation and ML model support (VIR-2548). Grep for `iimi` is clean; typecheck, lint, and the analyses/jobs test suites pass.


igboyes:
Addressed in cf4512b9:

- \`supportedWorkflows\` is now typed \`AnalysisWorkflow[]\` in \`analyses/utils.ts\`.
- \`sortKeys\` in \`Viewer/Sort.tsx\` is now \`Record<AnalysisWorkflow, string[]>\`, and the \`workflow\` prop is typed \`AnalysisWorkflow\` so the lookup stays in sync.

Not applying \`AnalysisWorkflow\` to \`sortTitles\` — its keys are sort-field names (coverage, depth, e, …), not workflows, so the union does not apply there.

PR #1088, opened 2026-07-02
feat: rework analysis creation with accessible selectors

igboyes:
## Summary
- Replace the subtraction dual-pane picker with an accessible multi-select combobox (new reusable `@base/MultiSelectComboBox` built on Downshift), showing selections as removable chips over a searchable menu.
- Swap the workflow tabs for a radio group above a single shared form — the subtraction and reference inputs are identical across workflows, so the tabs only ever swapped in a duplicate form. Also relabels the submit button to "Create" and tightens the dialog to its content.
- Give the reference selector an actionable empty state ("build a reference index") instead of a silently unopenable dropdown.
- Simplify the reference index rebuild flow: `RebuildIndex` is now a self-contained "Create" dialog gated behind build permission and unbuilt changes, dropping the `openRebuild` deep-link history-state plumbing.
- Add tests for the subtraction, reference, and workflow selectors.


igboyes:
Addressing Sourcery review:

**CreateAnalysisForm — empty compatibleWorkflows:** Skipping. `getCompatibleWorkflows` always returns `pathoscope` (the filter only gates `nuvs` on HMMs), so the list is never empty and `workflow` always defaults to at least `pathoscope`. An undefined workflow can't be submitted.

**RebuildIndex — success feedback / duplicate submissions:** The dialog already closes on success via the mutation's `onSuccess`. Added `disabled={mutation.isPending}` on the submit button in 10a5b560 to prevent duplicate submissions while the rebuild is in flight.

PR #1092, opened 2026-07-02
feat: show a shared error state when a page fails to load

igboyes:
## Summary
- Add a reusable `@base/ErrorState` component (icon, message, optional action) that accepts the standard color palette, and render it from `RouteError`'s retryable branch.
- Wire an `errorComponent` on the root route so render failures in the root shell are caught, not just child-route failures covered by the router's `defaultErrorComponent`.
- Document the shared primitive and root boundary in `docs/queries.md`; add tests for `ErrorState`.

Closes VIR-2480.

### DIFF
--- a/apps/web/src/base/ErrorState.tsx
+++ b/apps/web/src/base/ErrorState.tsx
@@ -8,6 +8,7 @@ type ErrorStateProps = {
 	children?: ReactNode;
 	className?: string;
 	color?: ErrorStateColor;
+	icon?: ReactNode;
 	message?: string;
 };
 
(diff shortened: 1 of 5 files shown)


igboyes:
Addressed in 9463b780:
- Icon accessibility: default `CircleAlert` is now `aria-hidden="true"` — the `<strong>` message carries the meaning.
- Icon configurability: added an optional `icon` prop to override the default.

Skipped:
- Semantic heading: keeping `<strong>` for consistency with the sibling `NotFound` component; a route-level error isn't necessarily the page heading.
- `h-96`: already overridable — `className` is merged last via `cn`, so callers can replace the height.

PR #1099, opened 2026-07-03
feat: link label chips to their filtered samples

igboyes:
- Wrap each label chip in the labels list with a `Link` to `/samples` pre-filtered by that label's ID
- Add hover styling to the chip so it reads as clickable


igboyes:
Re the two suggestions:
- Hover/cursor styling: it's already applied to the `SampleLabel` span, which is the sole child of the `Link` and is `inline-flex` sized to its own content — so the clickable area and the hover-highlighted area are already identical. Moving the classes onto the `Link` wrapper wouldn't change behavior, so leaving it as-is.
- Search param shape: confirmed the `/samples` route's `labels` search schema is `z.array(z.number()).default([]).catch([])`, which matches `search={{ labels: [id] }}` (array of numbers). No change needed.

PR #1112, opened 2026-07-06
ci: clean up CI workflow

igboyes:
## Summary
- Remove the check-commit job, since PR titles gate semantic-release rather than individual commits
- Extract shared checkout/pnpm/node/install steps into a reusable `.github/actions/setup` action and add timeout-minutes to every job
- Add a concurrency group that cancels stale `pull_request` runs, and move semantic-release deps to devDependencies run via `pnpm exec`


igboyes:
Re: the `release` job install — declining this one. Moving semantic-release into pinned, lockfile-managed devDependencies (run via `pnpm exec`) is the point of this change; `pnpm dlx` would pull unpinned versions and bypass the `minimumReleaseAge` supply-chain guard in `pnpm-workspace.yaml`. A filtered install wouldn't cover root devDependencies. `--frozen-lockfile` is the intended behavior here.

PR #1118, opened 2026-07-08
fix: restore visible loading spinner animation

igboyes:
## Summary
- fix the Loader ring rendering as a solid, non-spinning circle because tailwind-merge collapsed the transparent bottom border into the same color-conflict group as the solid sides
- emit per-side border color classes so the transparent bottom survives the merge and the spin animation reads correctly
- narrow `Loader`'s `color` prop to the shared six-color palette and fix two call sites passing unsupported values that were silently falling back to gray


igboyes:
Addressed in 2264bb42: `colorToClass` is now exported and the test derives its expected classes from it instead of duplicating the Tailwind strings, so palette changes stay in sync automatically.

On the exhaustiveness point: `colorToClass` is already typed as `Record<LoaderColor, string>` with a literal object, so TypeScript already errors on any missing or extra key — no further safeguard needed.

PR #1141, opened 2026-07-13
fix: stop authenticated users from enumerating other users' jobs

igboyes:
## Summary
- Scope `findJobs` and `getJob` to the acting user unless they hold an administrator role. Any authenticated user could previously page through every job on the instance and fetch any job by id, exposing owner handles, workflow state and progress, and the reconstructed sample/index/subtraction/analysis ids a job ran against.
- The scope constrains the counts and totals as well as the page of items — instance-wide aggregates would otherwise leak activity behind jobs the caller cannot see. `getJob` reports another user's job as missing rather than forbidden, so a 403 cannot be used to confirm a job exists.
- Adds the first real-Postgres integration suite (`server/jobs/`), built on the existing testcontainer and a per-suite Postgres schema.

## Note
This does not fully close the vulnerability on its own. Python's `/jobs` and `/jobs/{job_id}` have no owner scoping either — they fall back to `DefaultRoutePolicy`, which admits any authenticated client — and the session cookie is shared between the two services, so the enumeration stays reachable at `/api/jobs`. A matching change on the Python side is needed; VIR-2668's premise that this was a check dropped during the TypeScript migration turned out to be wrong, so there is no backend fix already in place.


igboyes:
This is opposed to the design of the app.

PR #1151, opened 2026-07-13
fix: soften the hover and focus halo on icon buttons

igboyes:
## Summary
- Replace icon buttons' solid, high-contrast hover/focus backgrounds with a soft tinted (10–20% opacity) halo
- Swap the hover text color from white to a darker shade of the button's own color for better legibility and less visual harshness

### DIFF
--- a/apps/web/src/base/IconButton.tsx
+++ b/apps/web/src/base/IconButton.tsx
@@ -39,21 +39,22 @@ export default function IconButton({
 				"p-2.5",
 				"rounded-full",
 				"text-inherit",
-				"hover:text-white",
-				"focus:text-white",
 
 				{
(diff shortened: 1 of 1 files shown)


igboyes:
Re Sourcery's review:

- **Black not changing hover text color**: intentional, not an inconsistency — `black` is already the darkest step in the palette, so there's no darker shade to move to on hover the way `blue-500 → blue-600` etc. do. The other colors get a text-color shift because they have room to darken; black doesn't.
- **Focus contrast at 10–20% opacity**: this halo sits on top of existing `text-*` icon color, which is unchanged by the halo itself — only the background tint is faint, not the icon or its focus ring. No accessibility regression expected, but happy to revisit if an audit flags something concrete.

No code changes from this review.

PR #1153, opened 2026-07-13
test: stop booting a postgres container for component tests

igboyes:
## Summary
- The web project's jsdom component tests never import `src/server` — the client transform strips server function bodies along with the server-only imports behind them, so `config.ts` and `db/pg.ts` are never reached. Booting Postgres for them was pure overhead, so `globalSetup` is dropped from that project. Component tests now run with no Docker at all.
- Filter the `nitro:*` plugins out of the Vitest plugin list. Nitro is the server bundler, which tests never ask to serve anything, and it holds file handles open that stalled shutdown by ~10s on every run. This was previously blamed on the empty `teardown()`, but a test importing nothing with no `globalSetup` at all still hung, while keeping `globalSetup` and dropping nitro did not. It also silences the `ReferenceError` spam from `@opentelemetry/instrumentation`.
- Remove the no-op `teardown()` exports rather than having them `stop()` the containers, which would defeat `withReuse()` and make every local run pay a fresh boot for no benefit — CI discards the runner anyway.

### DIFF
--- a/apps/web/src/server/storage/test/globalSetup.ts
+++ b/apps/web/src/server/storage/test/globalSetup.ts
@@ -200,6 +200,5 @@ export async function setup() {
 	await Promise.all([startGarage(), startAzurite()]);
 }
 
-// Left empty on purpose, matching the postgres setup: stop() would remove the
-// container and defeat withReuse() on the next run.
-export async function teardown() {}
+// There is deliberately no teardown, matching the postgres setup: stop() would
+// remove the containers and defeat withReuse() on the next run.
--- a/apps/web/src/tests/globalSetup.ts
+++ b/apps/web/src/tests/globalSetup.ts
@@ -18,7 +18,7 @@ export async function setup() {
 	process.env.VT_STORAGE_S3_BUCKET = "virtool-test";
 }
 
-// Left empty on purpose. stop() removes the container, so withReuse() finds
-// nothing to match on the next run and starts a fresh one. Clean up with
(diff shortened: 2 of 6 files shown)


igboyes:
Both points addressed in b4fd8195.

**Plugin name prefixes.** Agreed, and the fix was to stop filtering by name at all rather than harden the match. Nitro is now gated on `mode !== "test"` in `vite.config.js`, where Sentry is already gated on `command === "build"`. Neither plugin is constructed for tests, so `vitest.config.js` needs no plugin filtering. Confirmed there are now zero sentry- and nitro-named plugins under `{command: "serve", mode: "test"}` — the old `startsWith("sentry")` filter was in fact already dead, since `command === "build"` never held there.

**Guarding the client-transform assumption.** Good catch — that assumption carries the whole change, and silently regressing it would put Docker back in the component-test path. The `web` project now aliases `@server/config` and `@server/db/pg` to `src/tests/postgresGuard.ts`, which throws on import. These are the two modules that reach Postgres at import time (`config.ts` parses env; `db/pg.ts` opens a client), so a client file importing either now fails with an explicit message rather than quietly requiring a database.

Verified the guard actually fires rather than assuming it does: regressing `src/labels/queries.ts` to import `@server/config` directly fails the suite with the guard error, and reverting it goes green again. `pnpm test` (131 files / 736 tests), `pnpm check`, `pnpm typecheck`, and `pnpm build` all pass.

PR #1164, opened 2026-07-14
fix: meet WCAG AA contrast for placeholder and error text

igboyes:
## Summary

- `text-gray-400` placeholder and muted text (the shared input placeholder, `SelectButton`, `ReadSelectorSlots`, and subtraction `Attribution`) measured **2.60:1** on white. Bumped to `text-gray-500` (**4.84:1**) to clear the WCAG AA 4.5:1 floor. Only 5 of the 40 `text-gray-400` hits are actually text — ~22 are `EmptyMedia`, which is `aria-hidden` and presentational, so genuinely exempt; the rest are icons, tracked in VIR-2748.
- `InputError` was `text-red-500` at **3.82:1**; now `text-red-600` (**4.76:1**). Also bumped `text-xs` to `text-sm`, since at the 14px root that was a 10.5px error message — unreasonably small for the text most likely to be read under stress — and swapped the banned `min-h-[18px]` arbitrary class for `min-h-5`.
- The nav's About button rendered `text-gray-400` on the teal bar at **1.81:1**, effectively invisible. It now inherits white (**4.72:1**) and picks up the same hover pill and focus ring as `NavLink`. It could not simply drop its `color` prop — `IconButton` defaults `color` to `black`.

Ratios are computed from the oklch values Tailwind 4.3 ships; the computed "before" figures match the 2.60 and 3.81 reported in the issue. The `html { font-size: 14px }` root was evaluated and deliberately left alone — the large-text exemption would not rescue any of these failures even at a 16px root, and WCAG sets no minimum font size.

Closes VIR-2693.


igboyes:
Declining the centralization suggestion — the data points the other way.

`text-gray-500` is used **59 times across 35 files**, and **54 of those predate this PR**. It is already the app's established muted step; the 5 sites here are joining that convention, not introducing a new hardcoded one. Extracting a shared utility for *these 5* would leave the other 54 on the raw class, which fragments rather than centralizes.

The placeholder half is also already centralized: `inputBaseClasses` in `base/styles.ts` covers every `Input`/`InputSimple`, which is exactly why the fix there was a one-line change. `SelectButton` is a separate Radix trigger with its own class list.

The genuinely correct centralization is a semantic color token in `@theme` (which currently holds only the brand teal and the font) applied across all 59 sites. That's a design-system change, app-wide, and out of scope for a contrast fix.

PR #1169, opened 2026-07-14
fix: pass aria attributes through to form inputs

igboyes:
## Summary
- `Input` and `TextArea` dropped every prop outside a fixed allowlist, silently discarding `role`, `aria-invalid`, `aria-describedby`, `aria-expanded`, `aria-activedescendant`, `required`, and `autoComplete` — breaking `ComboBox`'s combobox role/active-descendant for screen readers and password fields' autocomplete hints.
- Type `Input`/`TextArea` as their native element props and spread the rest through, drop the unused `as` prop and dead `register` prop, and type `ComboBoxSearch`'s `getInputProps` so a dropped prop fails the build instead of hiding behind an implicit `any`.
- Drive the invalid border/ring from `aria-invalid` via a shared `inputInvalidClasses` constant, folding a surrounding `InputGroup` error into it through `useIsInvalid`, so the visual and assistive-tech states can't disagree.

### DIFF
--- a/apps/web/src/base/ComboBox.tsx
+++ b/apps/web/src/base/ComboBox.tsx
@@ -1,5 +1,5 @@
 import { cn } from "@app/cn";
-import { useCombobox } from "downshift";
+import { type UseComboboxReturnValue, useCombobox } from "downshift";
 import { ChevronDown } from "lucide-react";
 import type { ReactElement } from "react";
 import WrapRow from "./ComboBoxItem";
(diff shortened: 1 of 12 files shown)


igboyes:
On the two review points:

1. **forwardRef** — not needed under React 19. `ref` is a plain prop, and `InputProps = ComponentProps<"input"> & {...}` already includes a typed `ref?: Ref<HTMLInputElement>` that flows through the `{...props}` spread onto the native element. This is stronger typing than the old hand-rolled `Ref<HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement>` union. Consumers keep passing refs (and `react-hook-form` `register()`, which carries its own ref) unchanged, and the green Typescript check confirms no call site relied on a removed prop.

2. **Unconditional invalid classes** — intended. `inputInvalidClasses` are `aria-invalid:`-scoped variants, so the red border/ring only paints when `aria-invalid` is set; that's the whole point of making the attribute the single source of truth. I checked the call sites — none toggle error visuals via custom classes or props, so nothing regresses under the new model.

PR #1171, opened 2026-07-14
fix: revoke sessions on deactivation, password change, or forced reset

igboyes:
## Summary
- Deactivating a user, changing their password, or forcing a reset now deletes that user's sessions in the same transaction as the change, closing the window where the old credentials still authenticated.
- Replaces the write-only `invalidate_sessions` flag (which nothing read) with a direct `invalidateUserSessions` call, matching the self-service reset path. The column stays in the Drizzle mirror only to satisfy the NOT NULL insert default; dropping it and fixing the parallel Python bug are tracked in VIR-2750 and VIR-2751.
- Adds `updateUser` tests (session revocation, non-revoking edits, per-user isolation, transaction rollback) and updates `docs/auth.md` / `docs/server-push.md`.

Closes VIR-2671.

### DIFF
--- a/apps/web/src/server/auth/core.test.ts
+++ b/apps/web/src/server/auth/core.test.ts
@@ -176,7 +176,6 @@ describe("resetPassword", () => {
 		);
 		expect(after?.forceReset).toBe(true);
 		expect(after?.lastPasswordChange).toEqual(before?.lastPasswordChange);
-		expect(after?.invalidateSessions).toBe(before?.invalidateSessions);
 
 		// The session deletion rolled back with it.
 		const surviving = await db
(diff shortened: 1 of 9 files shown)


igboyes:
Re: Sourcery's two suggestions on `updateUser` — declining both, with reasons:

1. **Revoke only on actual change (no-op detection):** `password` is stored hashed, so cheaply detecting a no-op password update isn't possible without hashing to compare — no-op detection would only work for `active`/`force_reset` and be inconsistent across the credential fields. Over-revoking a session when an admin re-saves an unchanged value is the security-safe direction (worst case is a re-login), so the presence-based check is intentional.

2. **Derive `revokeSessions` from `patch` keys:** `patch` also carries `handle`, which must *not* revoke sessions, so deriving the flag from `Object.keys(patch)` would revoke on a rename. The three credential fields are enumerated explicitly for exactly this reason; a shared helper wouldn't be clearer than the current three-line check.

PR #1177, opened 2026-07-15
fix: standardize chart, progress, and label colours on the palette

igboyes:
## Summary
- Replace the hardcoded hex colours (and the hidden blue in the SelectBox focus glow) across `ProgressCircle` and the quality charts (`Bases`, `Nucleotides`, `Sequences`) with Tailwind palette tokens, matching how `Button`, `Badge`, and the quality-chart CSS already work. This shifts the charts and progress circles slightly by design (most visibly the azure blue).
- Extract the label default colour into a shared `DEFAULT_LABEL_COLOR` constant, imported by both the client form and the server fallback so the two can't drift apart.
- Centralize `max-h-[90vh]` onto `DialogContent` and normalize stray small-px arbitrary Tailwind values onto the standard spacing scale.

### DIFF
--- a/apps/web/src/analyses/components/Create/CreateAnalysisDialogContent.tsx
+++ b/apps/web/src/analyses/components/Create/CreateAnalysisDialogContent.tsx
@@ -8,7 +8,7 @@ export default function CreateAnalysisDialogContent({
 	children: ReactNode;
 }) {
 	return (
-		<DialogContent className={cn("max-h-[90vh]", "overflow-auto", "w-[700px]")}>
+		<DialogContent className="w-[700px]">
 			<div className={cn("flex", "flex-col")}>{children}</div>
 		</DialogContent>
 	);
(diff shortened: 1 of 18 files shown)


igboyes:
Re: Sourcery's two suggestions — declining both:

1. **Tightening `colorToVar`'s key type:** the map mixes a base color set with derived `${color}Lightest` template-literal keys, and `grey` uses a `greyLight` variant (special-cased in `getTrackColor`) rather than `greyLightest`. A strict union wouldn't cleanly model that asymmetry, and the map is a fully test-covered internal detail — little safety gained for the churn.

2. **Deriving `DEFAULT_LABEL_COLOR` from a CSS variable:** this constant is a persisted data value, not a style. `normalizeColor` returns it to be stored in Postgres as a literal hex, and the label form seeds a color picker with it — neither can consume `var(--color-gray-300)`. The JSDoc already documents its gray-300 provenance so it won't silently drift.

PR #1182, opened 2026-07-15
fix: unify base component colours on a shared palette type

igboyes:
## Summary
- Introduce a shared `PaletteColor` type (`@base/types`) and have every base component consume it instead of redeclaring the colour union locally, so the palette stays uniform across the surface.
- Add `IconColor` (`PaletteColor | "black"`) for icon-based components (`Icon`, `IconButton`, `Circle`) that need black in addition to the shared palette.
- Add the missing `orange` variant to `Button`/`Label`, widen `ProgressBarAffixed` to the full palette, standardise `ProgressCircle` on `gray`, and document the convention in AGENTS.md.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -341,8 +341,11 @@ patterns.
 ### Base component color props
 
 Base components in `src/base/` that expose a `color` prop should accept the
-shared palette: `"blue" | "green" | "gray" | "orange" | "purple" | "red"`. Don't
-add one-off colors or trim the set per component — keep the surface uniform.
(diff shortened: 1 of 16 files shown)


igboyes:
On the `ProgressBarAffixed` note: behaviour is unchanged. `color` still defaults to `"blue"` via the destructuring default, so an omitted `color` resolves to `colorToClass["blue"]` = `bg-blue-600` exactly as before. The removed `|| "bg-blue-600"` fallback was dead code — it could only fire for an off-palette string, which the new `PaletteColor` type makes impossible. Declining.

PR #1191, opened 2026-07-16
chore: stop biome from scanning build output

igboyes:
## Summary

- The Stop hook scanned the repo root, walking `apps/web/.output` (28M) and `apps/web/.nitro` (14M) of minified bundles and source maps. Both are gitignored, but `biome.json` sets no `vcs.useIgnoreFile`, so Biome never consults `.gitignore` and descends into them. Biome 2.5.3 deadlocks on that output — every worker thread parks in `futex` with no IO in flight and the process never returns.
- These artifacts have been present since May; 2.4.16 handled them in seconds. The regression arrived with `aab070a5`, bumping `@biomejs/biome` 2.4.16 → 2.5.3. Measured on `apps/web/.nitro`: **2.4.16 exits in 3s, 2.5.3 never returns.**
- Scope the hook to the paths the `check` script already uses, exclude the build-output directories so a bare `biome check .` cannot wedge either, and switch `npx` → `pnpm exec` to match the repo's package manager. `pnpm check` stays green; the hook now runs in ~2s.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -22,7 +22,7 @@
 				"hooks": [
 					{
 						"type": "command",
-						"command": "npx biome check --fix --unsafe . 2>/dev/null || true",
+						"command": "pnpm exec biome check --fix --unsafe apps/web/src packages 2>/dev/null || true",
 						"statusMessage": "Formatting with biome..."
 					}
 				]
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ dist-types
 *.tsbuildinfo
 .output
 .nitro
+.tanstack
 doc
 coverage
 /data
--- a/biome.json
+++ b/biome.json
@@ -1,7 +1,16 @@
 {
 	"$schema": "https://biomejs.dev/schemas/2.4.12/schema.json",
 	"files": {
-		"includes": ["**", "!**/routeTree.gen.ts", "!**/node_modules", "!**/dist"]
+		"includes": [
+			"**",
+			"!**/routeTree.gen.ts",
+			"!**/node_modules",
+			"!**/dist",
+			"!**/dist-types",
+			"!**/.output",
+			"!**/.nitro",
+			"!**/.tanstack"
+		]
 	},
 	"css": {
 		"parser": {
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,10 @@
 {
 	"$schema": "https://biomejs.dev/schemas/2.4.12/schema.json",
+	"vcs": {
+		"enabled": true,
+		"clientKind": "git",
+		"useIgnoreFile": true
+	},
 	"files": {
 		"includes": [
 			"**",


igboyes:
Declining both Sourcery suggestions, with reasons:

**`vcs.useIgnoreFile` instead of manual excludes** — it is not a substitute here, only a partial overlap. Per Biome's docs, `useIgnoreFile` reads `.gitignore` and `.git/info/exclude`, so it covers `.output`, `.nitro`, `dist`, `dist-types`, and `node_modules`, but it would miss the two entries that matter most:

- `apps/web/.tanstack` is **not** gitignored. It exists on disk today and only escapes `git status` because `.tanstack/tmp` is currently empty and git does not track empty directories. The moment TanStack writes into it, `useIgnoreFile` lets Biome descend into exactly the kind of build output this PR is fixing.
- `routeTree.gen.ts` is checked in deliberately, so no ignore file will ever exclude it.

Both would still need explicit `!` entries, leaving a split list — the excludes come from two sources instead of one — and adding a `vcs` block makes file discovery depend on git state for no reduction in the list. The manual excludes are the whole story in one place.

**Stop-hook path coverage** — confirmed intentional. `apps/web/src packages` is exactly the scope of the `check` and `format` scripts in `package.json`, which is what CI's `check-biome` job gates on. The hook matching the gate is the point: it cannot format something CI does not check, or skip something CI does. `apps/web` is the only app in the workspace. If an app is added, the `check` script has to be updated regardless, and the hook is updated in the same edit.

One narrowing this does introduce, deliberately: `apps/web/{tailwind,vite,vitest}.config.js` sit outside the scoped paths, so the hook no longer formats them (the old `biome check .` did). `pnpm check` does not cover them either, so there is no CI divergence.

igboyes:
Correcting my previous comment: I have enabled `vcs.useIgnoreFile` in df35dbe6.

I declined it on the grounds that it could not replace the manual excludes. That much is true, but "cannot replace" is not "not worth adding", and I conflated the two. Tested directly:

```
biome check apps/web/.nitro   →  "provided but ignored", exits immediately
```

So it does cover the build output on its own wherever `.git` is present, and it covers whatever gets gitignored next without anyone having to remember `biome.json`. That is worth having.

It is a second layer rather than a replacement, so the explicit excludes stay — they hold in the two cases `useIgnoreFile` does not cover: `routeTree.gen.ts` is tracked, so no ignore file can exclude it, and a no-git checkout (the Dockerfile copies source without `.git`) would walk back into the `.nitro` deadlock this branch fixes.

Also gitignored `apps/web/.tanstack`, which was the actual gap — build output that no ignore file listed, unnoticed only because `.tanstack/tmp` is empty and git does not track empty directories. It belongs in `.gitignore` on its own merits.

Verified nothing silently drops out of the check: no tracked file in the repo matches an ignore pattern (`git ls-files | git check-ignore --no-index --stdin`), and the counts are unchanged — `pnpm check` still 868 files, while a bare `biome check .` now finishes in 2s at 882 files where it previously never returned.

My second point stands as written: the hook scope intentionally mirrors the `check` script that CI gates on.

PR #1193, opened 2026-07-16
refactor: bring references and otus onto the query conventions

igboyes:
## Summary
- Split `useUpdateSourceTypes` into `useUpdateReferenceSourceTypes` and `useUpdateDefaultSourceTypes` in `references/queries.ts`, moving the form/undo machinery it bundled into a new `useSourceTypeEditor` hook beside the components that share it, and dropping a dead `onSuccess` branch that could never fire.
- Move the OTU context provider into `otus/components/CurrentOtuContext.tsx` so `otus/queries.tsx` can become a plain `.ts` file, and fold `getGenbank` into a `useGetGenbank` mutation now that `Accession` no longer needs its own state to drive it.
- Route cross-feature imports through the `@references/*` alias, leaving `references/hooks.ts` with only the reference-rights helpers.

Closes VIR-2711


igboyes:
@sourcery-ai on both points:

**404 handling** — good flag, and it was the one real risk here, so I verified it rather than reasoned about it. Superagent sets both `err.status` and `err.response.status` on an HTTP error, so the two are equivalent at runtime; `error.response?.status` is the shape `ErrorResponse` declares and what the rest of the repo already uses (e.g. `useFetchOtu`). Pinned it in 01df085f with a test that drives the real Auto Fill button through `CreateSequence` for both a hit and an unknown accession. Negative control: flipping the check to `=== 499` fails the test, so it isn't passing vacuously.

**Generic `UpdateSourceTypes`** — declining. `useSourceTypeEditor` never reads the mutation's data or error; it only calls `mutate(list, { onSuccess })` with zero-arg callbacks, and its return value exposes neither. Threading `TData`/`TError` through would infer types that nothing consumes. The one type that matters, the `string[]` variables, is already fixed, so passing a mutation with the wrong variables is still a compile error — and callers keep full typing on their own `useUpdateReferenceSourceTypes(refId)` result if they want `data`/`error`.

PR #1194, opened 2026-07-16
fix: keep file search results in sync with the search term

igboyes:
## Summary

- `useInfiniteFindFiles` passed `term` to the request as the `find` query parameter but left it out of the React Query key, so searches for different terms shared a cache entry and served the previous term's pages without refetching.
- Adds `term` to the `infiniteList` key. `QueryKeyFilter` already permits `undefined`, so no signature change was needed.
- Latent today — none of the three call sites (`uploads/hooks.ts`, `CreateSample.tsx`, `SubtractionCreate.tsx`) passes `term`. The key gains a trailing `undefined` segment for them, retiring the current cache entries once and changing no behaviour.

Closes VIR-2707.

### DIFF
--- a/apps/web/src/uploads/queries.ts
+++ b/apps/web/src/uploads/queries.ts
@@ -43,7 +43,7 @@ export function useInfiniteFindFiles(
 	term?: string,
 ) {
 	return useInfiniteQuery<FileResponse>({
-		queryKey: fileQueryKeys.infiniteList([type, per_page]),
+		queryKey: fileQueryKeys.infiniteList([type, per_page, term]),
 		queryFn: ({ pageParam }) =>
 			findFiles(type, pageParam as number, per_page, term),
 		initialPageParam: 1,


igboyes:
Declining both Sourcery points.

**Normalizing `term` in the key.** The key must describe the request that was actually sent. Normalizing only the key would collapse `""` and `undefined` onto one cache entry while `findFiles` still sends `find=` versus omitting it — reintroducing the exact bug this PR fixes, just narrower. Normalizing the request too is a behaviour change to the request layer, out of scope here, and would make uploads the only feature that does it: `indexes`, `references`, `groups`, `users`, `analyses`, `hmm`, and `subtraction` all put `term` in the key raw. `useInfiniteFindGroups` is the direct analogue and does `infiniteList([per_page, term])`.

**Debouncing.** There is no calling code to review — none of the three call sites passes `term` today. When a file-search box is wired up it will use `useDebouncedDraft` from `@app/hooks`, the established pattern in `OtuToolbar` and `ReferenceToolbar`, which debounces upstream of the hook and so is unaffected by what the key contains.

PR #1197, opened 2026-07-17
test: cover the query client's retry and auth-error handling

igboyes:
## Summary

- Extracts the retry predicate and `handleAuthenticationError` out of `router.tsx` into `app/queryErrors.ts` so they can be reached from a test. No behavior change — the VIR-2700 fix itself landed in 43a0c630.
- Covers both: the retry guard giving up on 401/403/404 from the Python API and on the auth error names carried across the server-function boundary, bounded retries otherwise, and `endSession` firing on a 401 but deliberately not on a 403 (a forbidden user's session is valid — they just lack the role).
- The real-boundary serialization test VIR-2700 also asked for can't be written at unit level: mocking the server function means nothing serializes, and the alternatives assert TanStack internals rather than behavior. Deferred to VIR-2783 (Playwright).

### DIFF
--- a/apps/web/src/app/__tests__/queryErrors.test.ts
+++ b/apps/web/src/app/__tests__/queryErrors.test.ts
@@ -1,0 +1,94 @@
+import {
+	handleAuthenticationError,
+	type QueryError,
+	shouldRetryQuery,
+} from "@app/queryErrors";
+import {
+	FORBIDDEN_ERROR_NAME,
+	UNAUTHORIZED_ERROR_NAME,
+} from "@virtool/contracts";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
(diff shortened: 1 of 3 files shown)


igboyes:
Addressed the second point in bbe368c8 — the `[401, 403, 404]` list is now a named `NON_RETRYABLE_STATUSES` constant.

Declining the first (centralizing the auth-error names): `handleAuthenticationError` and `shouldRetryQuery` match different sets on purpose. `handleAuthenticationError` ends the session only on 401; a 403 means the session is valid but the role is wrong, so bouncing to the login wall would be wrong. `shouldRetryQuery` refuses to retry on both. A shared constant would invite aligning the two and reintroduce that bug, so the divergence is intentional.

PR #1206, opened 2026-07-17
refactor: share the paginated route schema and per-page constant

igboyes:
## Summary
- Adds `@app/pagination` with a shared `Paginated` type, `paginated()` search-schema helper, and `DEFAULT_PER_PAGE` constant so list routes stop re-declaring `page: number` and hardcoding `25`.
- Migrates all eleven paginated routes (users, hmms, jobs, indexes, otus, refs, sample analyses/files, samples, subtractions/files) onto the shared helper and constant.
- Documents the convention in `AGENTS.md` so new paginated routes follow it by default.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -243,6 +243,11 @@ So, in a route's critical exports:
   dependency-free coercion helpers in `@app/searchParams` and type the function
   as `(input: Partial<T> & SearchSchemaInput) => T` — the `SearchSchemaInput`
   tag is what keeps `<Link search={{ page: 2 }}>` partial.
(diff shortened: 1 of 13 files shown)


igboyes:
Re: the `paginated` input type — keeping it as `{ page?: unknown }` on purpose. It's a coercion helper layered on `@app/searchParams`, whose contract is to accept an unknown raw value (the router hands over JSON-parsed search params that a malformed URL can make non-numeric) and normalize it via `num()` so the page still renders. Typing the input as `Partial<Paginated> & SearchSchemaInput` would narrow `page` to `number`, misrepresenting exactly the input this helper exists to sanitize, with no type-safety gain since callers already pass without repeating any shape.

PR #1210, opened 2026-07-17
fix: give d3 charts a text alternative for screen readers

igboyes:
## Summary
- Sets `role="img"`, `aria-label`, and a `<title>` on the SVGs drawn by the quality, Nuvs, and Pathoscope d3 charts so screen readers announce a description instead of skipping the graphic entirely.
- Derives each label from the chart's underlying data (e.g. read position count, ORF strand and position range) so it stays accurate as the data changes.
- Adds tests covering the new labels for `Bases`, `Nucleotides`, `Sequences`, `NuvsOrf`, `NuvsSequence`, and `PathoscopeSequence`.

### DIFF
--- a/apps/web/src/analyses/components/Nuvs/NuvsOrf.tsx
+++ b/apps/web/src/analyses/components/Nuvs/NuvsOrf.tsx
@@ -11,11 +11,19 @@ function draw(element, maxLength, pos, strand) {
 
 	const width = element.offsetWidth - 30;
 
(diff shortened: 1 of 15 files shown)


igboyes:
Both applied in f4039016: extracted a shared `labelSvg` helper in `@samples/charting` (now used by `createSvg` and the Nuvs/Pathoscope draw functions), and added a `pluralize` helper in `@app/format` for the interpolated counts.

PR #1216, opened 2026-07-17
chore: stop vite dev server full reloads on cold start

igboyes:
## Summary
- The Vite dev server reoptimized dependencies and forced a full page reload on every cold start instead of applying HMR. `server.warmup.clientFiles` globbed `./src/{app,base,nav}/**/*.{ts,tsx}`, which also matched the `__tests__` files in those trees — warming them pulled `@tests/setup` (nock, vitest, `@faker-js/faker`) and testing-library into the client optimize graph *after* the scan, triggering a late-discovery reoptimize and reload. Scoped the warmup globs off test files.
- Removed the dead `numbro` entry from `optimizeDeps.include`; it isn't a dependency and logged `Failed to resolve dependency: numbro` on every start.
- Verified against the issue's repro (`DEBUG=vite:deps vite --force`): a single clean optimize pass ("the scanner found every used dependency"), no `new dependencies found` waves, no `reloading`, no test-only packages in the graph, and no numbro warning. `holdUntilCrawlEnd: false` is kept and now documented as safe because the plugin-provided entries make the cold-start scan complete.

Closes VIR-2759.

### DIFF
--- a/apps/web/vite.config.js
+++ b/apps/web/vite.config.js
@@ -122,6 +122,13 @@ export default defineConfig(({ command, mode }) => ({
 			}),
 	],
 	optimizeDeps: {
+		// The `tanstackStart` plugin points the client scanner at `client.tsx` and
+		// `router.tsx`, so a cold-start scan reaches every used dependency in one
+		// pass ("the scanner found every used dependency"). That is the condition
+		// under which releasing the first optimize early is safe — with it,
+		// nothing is discovered late to trigger a reoptimize and a full page
+		// reload. It stops being safe if the scan is starved again, e.g. by
+		// warming test files (see the `warmup` note below).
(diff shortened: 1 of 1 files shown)


igboyes:
Both suggestions declined, with reasons:

1. `.spec` patterns: the repo has no `.spec` files — the convention is `.test.{ts,tsx}` in `__tests__/` dirs. `!./src/**/__tests__/**` already catches every app/base/nav test regardless of filename, and `!./src/**/*.test.{ts,tsx}` covers the rest. Adding negations for an unused naming convention is speculative clutter.

2. Shared constant: the negated patterns already apply globally to every positive glob in the same `clientFiles` array (tinyglobby merges them), so any future positive entry added there is automatically covered by the existing exclusions. A single-use constant would add indirection without reducing real risk.

PR #1233, opened 2026-07-20
fix: make label color picker keyboard and screen-reader accessible

igboyes:
## Summary
- Turn the preset color squares into a native radio group (`fieldset` with a group `aria-label` and `<input type="radio">` swatches) so the picker is operable by keyboard, with arrow-key navigation between swatches.
- Give every preset color a human-readable accessible name (e.g. "Dark blue") announced to screen readers instead of only exposing the raw hex value.
- Reflect the current value as the checked radio and add a selected-state ring, keeping the existing hover/focus styling.

### DIFF
--- a/apps/web/src/base/Color.tsx
+++ b/apps/web/src/base/Color.tsx
@@ -1,80 +1,108 @@
 import { cn } from "@app/cn";
 import Input from "./Input";
 
-const colors: string[] = [
+type PresetColor = {
+	// The color in hex format, without a leading "#"
+	hex: string;
+	// The accessible name announced for the swatch
+	name: string;
+};
+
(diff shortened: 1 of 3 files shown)


igboyes:
Addressing Sourcery's review:

- **Ring styles**: consolidated in e364b55 — hoisted the shared `ring-offset-2 ring-offset-gray-300` geometry (inert without an active `ring-2`) so focus-within and checked now only toggle `ring-2` and swap the ring color.
- **`group` prop**: keeping it. The group name must be unique per picker instance so two `<Color>` radio groups on one page don't collide, so it's derived from the required `id`. Having `ColorSwatch` compute it internally would still mean threading `id` down — a lateral change, not a simplification.

PR #1235, opened 2026-07-20
chore: stop reporting expected auth rejections to Sentry

igboyes:
## Summary

- Auth middleware throws `UnauthorizedError`/`ForbiddenError` by design on missing/insufficient sessions, but Sentry's server-function instrumentation was capturing these routine 401/403s as incidents (VIRTOOL-5).
- Add a `beforeSend` hook that drops events whose exception matches one of the expected auth error names, keyed on the shared name constants from `@virtool/contracts`.

### DIFF
--- a/apps/web/src/instrument.server.ts
+++ b/apps/web/src/instrument.server.ts
@@ -1,4 +1,5 @@
 import * as Sentry from "@sentry/tanstackstart-react";
+import { dropExpectedAuthErrors } from "@server/sentryFilters";
 import { getCommonOptions } from "@virtool/sentry";
 
 // Server-side Sentry initialisation. Imported for its side effect at the top of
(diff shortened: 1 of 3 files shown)


igboyes:
Re Sourcery's two suggestions — declining both, with reasoning:

1. **Relaxing `isExpectedAuthError` to duck-type `{ name: string }` instead of `instanceof Error`.** The middleware throws only real `UnauthorizedError`/`ForbiddenError` instances (`throw new UnauthorizedError()`), so `instanceof Error` always holds on the `originalException` path. The non-Error/serialized case is already covered by the second check, `eventReportsAuthError`, which reads `exception.values[].type` off the built event. Loosening the instance check trades precision for a shape our own code can't produce.

2. **Factoring out the `exception.values` shape / narrowing to reduce optional chaining.** There's a single filter and a single access site today; extracting a shared shape helper now is abstraction for a second caller that doesn't exist. Worth revisiting if/when a second Sentry filter lands.

PR #1256, opened 2026-07-23
fix: hide reference details from users who cannot see them

igboyes:
## Summary
- Enforce the same visibility rule on the single-reference detail read that already applies to the reference list, so a non-admin, non-member caller gets a 404 instead of the reference's data.
- Add `checkReferenceVisibility` to `references/data.ts`, treating a hidden reference and a missing one identically to avoid leaking existence.
- Cover the new check with tests for a blocked non-member, a full administrator, and a member reached only through a group.

### DIFF
--- a/apps/web/src/server/references/data.ts
+++ b/apps/web/src/server/references/data.ts
@@ -420,6 +420,39 @@ export async function checkReferenceRight(
 	return false;
 }
 
+/**
+ * Whether `actor` may see a reference at all — the single-row form of
+ * {@link referenceVisibilityFilter}, so detail read enforces the same rule the
+ * list does. A full administrator sees every reference; otherwise ownership, any
+ * user membership row, or any group membership row grants visibility, regardless
+ * of the rights flags on that row.
+ *
+ * Returns `false` for both a missing reference and one the actor cannot see, so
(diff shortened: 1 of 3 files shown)


igboyes:
Re: moving the visibility check into the data-layer `getReference` — declining, for two reasons:

1. **Authorization belongs in the handler, not `data.ts`.** Per this repo's server layering, `data.ts` owns pure domain/persistence and `functions.ts` enforces the auth floor and any row-dependent rule after the read. A visibility gate inside `getReferenceImpl` would push authorization down into the data layer, which the architecture deliberately keeps out of it.

2. **The other `getReference` call sites must *not* apply visibility.** `getReference` is called internally by `createReference`, `updateReference`, and `setReferenceArchived` as a post-mutation re-read to return the fresh row. Those have no actor and should never 404 — folding a visibility filter in would break them. So the consistency the suggestion is aiming for would actually be incorrect for the majority of callers; only the `getReference` *server function* should enforce it, which is exactly where the check lives now.

The `checkReferenceVisibility` helper keeps the single-row rule in one place (mirroring `referenceVisibilityFilter` for the list) while leaving enforcement to the handler.

PR #1257, opened 2026-07-23
feat: serve samples from the typescript server

igboyes:
## Summary
- Move the six sample endpoints (list, detail, create, update, delete, rights) from the Python REST API onto TanStack Start server functions, with the shared `has_sample_right` predicate exported for the Analyses work to reuse.
- Expand the `legacy_samples` Drizzle mirror and move sample wire shapes into `@virtool/contracts`; add `createJob` / `reserveUploads` so sample creation can schedule its job and hold its uploads.
- Migrate every sample and dependent analysis test from the nock-based `tests/api/samples.ts` fakes to `tests/server-fn/samples.ts` stubs, and mark Samples as Built in `docs/database.md`.


apps/web/src/server/samples/data.ts
```diff
@@ -0,0 +1,1229 @@
...
+			color: row.color ?? "",
+			description: row.description ?? "",
+		});
+		bySample.set(row.sampleId, list);
+	}
+
+	return bySample;
+}
+
+async function getUsersByIds(
```
igboyes: I haven't looked at the usage for this, but would guess it should be a join or should be a broadly shared function.

apps/web/src/server/samples/data.ts
```diff
@@ -0,0 +1,1229 @@
...
+	] = await Promise.all([
+		getLabelsBySample(db, [sampleId]),
+		getWorkflowTagsBySample(db, [sampleId]),
+		getSampleJobs(db, jobIds),
+		getUsersByIds(db, userIds),
+		getSubtractionsBySample(db, sampleId),
+		getArtifacts(db, sampleId),
+		getReads(db, sampleId),
+		row.group_id != null ? getGroup(db, row.group_id) : Promise.resolve(null),
+	]);
```
igboyes, line 698: A lot of this should be joins.

apps/web/src/server/samples/data.ts
```diff
@@ -0,0 +1,1229 @@
...
+	sampleId: number,
+): Promise<number | null> {
+	const [row] = await db
+		.select({ userId: legacySamples.user_id })
+		.from(legacySamples)
+		.where(eq(legacySamples.id, sampleId))
+		.limit(1);
+
+	return row ? row.userId : null;
+}
```
igboyes, line 753: An ownerless sample is an invalid state that should surface as an error.

apps/web/src/server/samples/data.ts
```diff
@@ -0,0 +1,1229 @@
...
+		return;
+	}
+
+	const rows = await db
+		.select({ id: labels.id })
+		.from(labels)
+		.where(inArray(labels.id, labelIds));
+
+	const found = new Set(rows.map((row) => row.id));
+	const missing = labelIds.filter((id) => !found.has(id));
```
igboyes, line 821: Can this not be a single query?

packages/contracts/src/samples.ts
```diff
@@ -1,17 +1,153 @@
...
+	| "running"
+	| "succeeded";
+
+/** A workflow a sample's creation job can run. */
+export type SampleJobWorkflow =
+	| "build_index"
+	| "create_sample"
+	| "create_subtraction"
+	| "nuvs"
+	| "pathoscope";
```
igboyes: This doesn't make sense. These are all possible workflows. Only create_sample builds out a sample.

packages/contracts/src/samples.ts
```diff
@@ -1,17 +1,153 @@
...
+	pathoscope: WorkflowState;
+};
+
+/** A lifecycle state a sample's creation job can be in. */
+export type SampleJobState =
+	| "cancelled"
+	| "failed"
+	| "pending"
+	| "running"
+	| "succeeded";
```
igboyes: True for all jobs.

igboyes:
Addressed the two Codex review threads (resolved):

- **Reads/artifacts on legacy samples** (b0898884): `getReads` and `getArtifacts` now resolve the storage id from `legacy_id` and OR against the text `sample` column, matching the delete path, so legacy rows with a null `sample_id` are found.
- **Reserving non-reads files** (d6bd64ab): `reserveUploads` now requires each id to resolve to a visible reads upload (`type = reads`, ready, not removed) before reserving; a reference/subtraction, unfinished, or removed upload is rejected as `UploadNotFoundError` so none is reserved.

(Replied here rather than inline because there's a pending review on the PR that would otherwise swallow the replies.)

packages/contracts/src/samples.ts
```diff
@@ -1,17 +1,153 @@
...
+
+/** The creation job embedded in a sample. */
+export type SampleJobNested = {
+	created_at: string;
+	id: number;
+	progress: number;
+	state: SampleJobState;
+	user: UserNested;
+	workflow: SampleJobWorkflow;
+};
```
igboyes: This is just job nested. The only differentiator for a job nested in a sample is that workflow must be "create_sample".

packages/contracts/src/samples.ts
```diff
@@ -1,17 +1,153 @@
...
+	notes: string;
+	nuvs: boolean | string;
+	pathoscope: boolean | string;
+	ready: boolean;
+	user: UserNested;
+	workflows: SampleWorkflows;
+};
+
+/** A complete sample, as returned by the detail endpoint. */
+export type Sample = SampleMinimal & {
```
igboyes, line 102: Should be camelCase.

packages/contracts/src/samples.ts
```diff
@@ -1,17 +1,153 @@
...
+
+	/** The read-length range */
+	length: number[];
+
+	/** Data for the sequences chart */
+	sequences: number[];
+};
+
+/** A sample reduced to the fields shown in resource listings. */
+export type SampleMinimal = {
```
igboyes, line 84: Should be camelCase.

apps/web/src/server/samples/data.ts
igboyes: Done in 50126b33 — joined `users` onto the sample row in both getSample and findSamples and dropped getUsersByIds.

apps/web/src/server/samples/data.ts
igboyes: 50126b33 — the one-to-one owner and group now join onto the sample row. The one-to-many relationships (labels, subtractions, reads, artifacts, workflow tags) stay as separate result sets batched by the page's ids.

apps/web/src/server/samples/data.ts
igboyes: 7791a6df — getSample and getSampleOwnerId now throw SampleOwnerlessError instead of degrading to a blank id-0 owner. The list path stays tolerant so a single corrupt row can't fail a whole page.

apps/web/src/server/samples/data.ts
igboyes: It's already a single query — one `SELECT id FROM labels WHERE id IN (...)`. The JS diff afterward only builds the "Labels do not exist: …" message listing the missing ids; a bare COUNT would lose that. Left as-is — happy to switch to a count if you'd rather drop the per-id detail.

packages/contracts/src/samples.ts
igboyes: 57da855c — a sample's embedded job is now typed `workflow: "create_sample"`.

packages/contracts/src/samples.ts
igboyes: 57da855c — replaced by a shared `JobState` in @virtool/contracts.

packages/contracts/src/samples.ts
igboyes: 57da855c — SampleJobNested is now the shared `JobNested` narrowed to `workflow: "create_sample"`.

packages/contracts/src/samples.ts
igboyes: 57da855c — the whole sample wire shape (minimal, detail, rights, and the paginated envelope) is camelCase now, matching the references migration.

packages/contracts/src/samples.ts
igboyes: 57da855c — camelCased.

PR #1279, opened 2026-07-28
refactor: import data functions under their own names

igboyes:
## Summary
- Drops all 83 `... as ...Impl` import aliases from the fifteen server `functions.ts` modules that still had them, finishing what 96c5fe85 started on `indexes`. The `Fn` suffix already distinguishes a server function from the data function it wraps, so the alias only gave one thing a second name.
- Keeps the alias in `auth/policy.test.ts` and `groups/functions.test.ts`, where it earns its place — each defines a local `seedGroup`/`addToGroup` that closes over `db` and genuinely shadows the imported fixture.
- Records the convention in `AGENTS.md` under **Naming**; `docs/code-style.md` already said aliasing wasn't needed for this pair.

Every rename was checked for shadows first. The only two plain names appearing elsewhere in their own file — `findJobs` and `createOtu` — turned out to be prose in comments, not bindings.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -702,7 +702,9 @@ The basics:
   (`loginFn`, `getSampleFn`) — it's an RPC call, not a plain function,
   and the suffix marks that at every call site. The domain function it
   wraps keeps the plain name (`login`, `getSample`) and never crosses
-  the network.
+  the network. `functions.ts` imports that data function under its own
+  name — the `Fn` suffix already keeps the two apart, so don't alias it
+  to `...Impl` on the way in.
 - **Comments:** Default to none. Document *why* when non-obvious, not
   *what*.
 - **Concurrency:** Independent awaits go in `Promise.all` — don't pay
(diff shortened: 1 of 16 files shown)


igboyes:
Not taking this one here.

`rethrowAsHttp` is typed `(err: unknown): never` and unconditionally throws, so no handler ever returns `undefined` at runtime. All three `messages/functions.ts` call sites are byte-identical on `origin/main` — this PR only renamed imports and touched no control flow.

There is a real defect underneath, but it is a typing one, not the one described. Because `rethrowAsHttp` is a `const` with no explicit type annotation, TypeScript declines to apply never-returning control-flow analysis to the call, so the handlers' inferred return types silently pick up `| undefined` and leak it to client call sites. Verified: the same pattern in a probe file errors with `TS2366`.

That affects 8 bare call sites across 3 files (`messages` 3, `groups` 4, `tasks` 1), all pre-existing. Fixing 3 of them inside a mechanical-rename PR would be both out of scope and inconsistent, so it gets its own change.

PR #1305, opened 2026-08-04
test: deduplicate the pathoscope hit helper into a shared fake

igboyes:
## Summary
- Four test files each declared their own `createHit`, so every change to `PathoscopeHit` cost four identical edits. `createFakePathoscopeHit` now lives in `src/tests/fake/analyses.ts` alongside the other fakes.
- Its defaults are fixed rather than faked: the export tests assert the exact TSV a hit renders to, and every column of that string comes from a default left unoverridden.
- `hooks.test.tsx` was the one variant, defaulting its fields to zero. It moves onto the shared defaults — every hit it builds overrides the five fields its assertions read.

Closes VIR-2912.

### DIFF
--- a/apps/web/src/analyses/__tests__/hooks.test.tsx
+++ b/apps/web/src/analyses/__tests__/hooks.test.tsx
@@ -10,47 +10,30 @@ import type {
 	FormattedPathoscopeAnalysis,
 } from "@analyses/types";
 import { renderHook } from "@testing-library/react";
-import type { PathoscopeHit } from "@virtool/contracts";
+import { createFakePathoscopeHit } from "@tests/fake/analyses";
 import type { ReactNode } from "react";
 import { describe, expect, it, vi } from "vitest";
-
-function createHit(overrides: Partial<PathoscopeHit>): PathoscopeHit {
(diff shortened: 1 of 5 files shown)


igboyes:
Declining the second helper.

`hooks.test.tsx` is not coupled to the defaults. Every hit it builds overrides `id`, `name`, `coverage`, `depth` and `pi`, and those five fields are the only ones its assertions read — sorting by each of them, plus the coverage cutoff. Nothing it inherits from the fake is observable, so a zeroed variant would produce identical results.

The file that *is* coupled is `table.test.ts`, deliberately: it asserts the exact TSV a hit renders to, and every column of that string comes from an unoverridden default. That is why the defaults are fixed rather than faked, and there is a comment on the function saying so.

A second "neutral" helper would also reinstate the problem this PR removes. VIR-2912 exists because four divergent copies of the helper meant four identical edits per contract change; shipping two variants makes it two. `main` moved under this branch mid-review with exactly that kind of change (`2d773341`, adding fields to `PathoscopeIsolate` and `PathoscopeSegment`), which is the case in point.

PR #1329, opened 2026-08-07
chore: serve workflow metadata reads from the jobs api

igboyes:
## Summary

- Adds six read endpoints — `GET /samples/{id}`, `/subtractions/{id}`, `/indexes/{id}`, `/analyses/{id}`, `/refs/{id}`, `/settings` — so a running workflow can fetch the records behind its job, at Python's own resource paths with no prefix.
- Each handler reuses the data function `apps/web` already calls and maps it to a narrow camelCase `Workflow*` shape at the handler boundary, so no field is renamed inside a data function the web client also reads. Reads take `{ db }` and nothing else, so a read handler cannot reach storage or build a derived artifact.
- `Read`, `SubtractionFile` and `IndexFile` gain a recorded `storageKey`, selected from the row and never composed — a migrated object keeps whatever prefix it was written under, so a response without the key leaves a workflow unable to reach the bytes at all.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,14 +21,19 @@ This is a **pnpm monorepo**:
   mirroring Python's `virtool/jobs/main.py` (`api-jobs-service`, ClusterIP,
   **no ingress** — that absence is the security boundary). Serves
   `/health/live`, `/health/ready`, a token-gated `/metrics`, the two
-  cache endpoints — `GET /caches/{key}` and `POST /caches` — and the three
+  cache endpoints — `GET /caches/{key}` and `POST /caches` — the three
   finalize routes — `PATCH /subtractions/{id}`, `PATCH /samples/{id}`,
-  `PATCH /analyses/{id}` — today. Image:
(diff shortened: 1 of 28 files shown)


igboyes:
Declining the `handleReadResource` extraction, with reasons:

1. **The shape is service-wide, not read-specific.** The cache handlers, the three finalize routes and the five job lifecycle handlers that landed in #1327 all repeat the same `requireJobRequest` → `parseRowId` → map → `catch NotFoundError` sequence. Abstracting only the six reads leaves the service half-abstracted, which is worse than uniformly explicit.

2. **`handleGetSettings` doesn't fit the shape.** It takes no id and has no 404 — the settings row is a singleton and `getSettings` seeds it. So the helper would cover 5 of 6.

3. **The catch has to name the specific error class, and that matters here.** `getSample` also throws `SampleOwnerlessError` — a sample row with no owner is an invalid state, and it must keep surfacing as a 500. A generic helper is exactly the thing that invites someone to widen the catch to `AppError` later and silently convert that bug into a 404.

The per-handler `catch` naming one error class is the honest version of this, and it is about 6 lines each.

PR #1342, opened 2026-08-08
chore: harden the jobs api and share its plumbing across services

igboyes:
## Summary

- Corrects defects found reviewing `apps/jobs-api`: the finalize routes never checked that the calling job owns the row it flips ready, a job login authenticated a zero-padded id, a crashed request was counted as a success by the metrics middleware, a finalize naming no files could leave a parent ready with no file rows, and a thrown handler reached neither pino nor Sentry.
- Extracts the plumbing three services had been duplicating or going without — `@virtool/service`'s shutdown controller and `@virtool/sentry`'s pino log stream — and collapses the jobs API's own repetition into `parseJsonBody`, `requireRowId` and `finalizeResource`. The extracted shutdown controller also divides its budget per step rather than pooling it, so a listener that hangs no longer starves the pool drain and the Sentry flush that would have recorded the failure.
- Reconciles the Drizzle mirror with Python's CHECK constraints, drops the casts that stood in for it, and publishes the SPA's job wire shapes from `@virtool/contracts` so the row-to-wire mapping has one implementation instead of two.


igboyes:
Addressed in bc1d6c04.

The shutdown budget is now divided rather than pooled: every step, each registered hook included, gets an equal share of the time still left when it starts and is abandoned once that share is spent, with a step that finishes early handing its remainder on. A `closeListener` that never settles no longer starves `closeDatabase` and `flushSentry` — which are what record the failure, so losing them lost the evidence exactly when there was some.

Two notes on scope. An abandoned step is not cancelled, so a genuinely stuck socket can still leave the process to be reaped rather than draining on its own; what the division buys is that everything after it runs. And on the socket itself, both services already call `closeIdleConnections()` alongside `close()`, so idle keep-alive sockets — the realistic case, given undici's dispatcher on the runner side and the kubelet's probes on the tasks side — are dropped rather than waited out. Force-closing connections that are actually in flight would abort requests mid-response, which is worse than a bounded wait.

The docblock already claimed "a listener that will not close still leaves the pool to drain", which was true only for one that throws; that gap is now closed in the code and the wording in docs/tasks.md and docs/jobs-api.md.

PR #1364, opened 2026-08-12
fix: show new references without a page reload

igboyes:
## Summary

- Create, clone and import were the only reference mutations with no cache invalidation, so a new reference stayed invisible until a hard reload. The `/refs` loader uses `ensureQueryData`, which returns the cached page rather than refetching, and `ReferenceList` reads the same list key through `useSuspenseQuery` without remounting — nothing in the flow triggered a refetch. All three now invalidate `referenceQueryKeys.lists()` on success, matching `useArchiveReference`/`useUnarchiveReference`.
- The `references` domain was fully wired on the SSE path but `packages/data/src/references/data.ts` published no `client_events` frames at all, so a reference created, renamed, archived or reassigned by one user stayed invisible to everyone else until they reloaded. `createReference` now emits a `create` frame after its transaction commits; `updateReference`, `setReferenceArchived` and the six membership mutations emit an `update` frame. A no-op update and a mutation that threw both stay silent.
- Regression tests cover both halves — the empty-create and clone paths in `ReferenceList.test.tsx`, and the published frames in a new `packages/data/src/references/data.test.ts` — each verified to fail against the unfixed code.

Closes VIR-2943.

### DIFF
--- a/apps/web/src/references/components/__tests__/ReferenceList.test.tsx
+++ b/apps/web/src/references/components/__tests__/ReferenceList.test.tsx
@@ -160,16 +160,8 @@ describe("<ReferenceList />", () => {
 			mockGetAccount(createFakeAccount({ permissions }));
 
 			const created = createFakeReferenceMinimal({ name: "Fresh Reference" });
-			let items = [references];
 
-			referenceServerFnMocks.findReferencesFn.mockImplementation(async () => ({
-				foundCount: items.length,
-				totalCount: items.length,
-				page: 1,
-				pageCount: 1,
-				perPage: 25,
-				items,
-			}));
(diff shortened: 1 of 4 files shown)


igboyes:
Applied the second point in 62a80382 — the two regression tests now re-stub through the shared `mockFindReferences` helper instead of reassigning a captured `items` array.

Declining the first. The three create hooks are three of ten mutations in `references/queries.ts`, and the other seven each spell their `invalidateQueries` inline too — `useArchiveReference` and `useUnarchiveReference` invalidate the same `lists()` key that way. Extracting a helper for only these three would make them read differently from their neighbours to save four lines, and the next hook added would have to pick a side.

PR #1374, opened 2026-08-13
fix: stop reporting a client that leaves mid-request as an incident

igboyes:
## Summary
- Sentry's `beforeSend` filter already dropped the `Error: aborted` node raises on a closed socket, but not the `AbortError: This operation was aborted` srvx raises when a response outlives its client — so every browser tab that unloaded mid-response reported a false incident, with the request's own status already recorded as a 200.
- The steady source is `POST /monitoring`: the browser SDK flushes its envelopes on unload, so the tab is usually gone before the tunnel has finished streaming Sentry's reply back to it.
- Matched on the name paired with node's exact bare-`abort()` message, so an `AbortSignal.timeout` or an `abort(reason)` stays reportable.

Fixes VIRTOOL-N

### DIFF
--- a/apps/web/src/server/__tests__/sentryFilters.test.ts
+++ b/apps/web/src/server/__tests__/sentryFilters.test.ts
@@ -24,6 +24,13 @@ function connResetError(message: string): Error {
 	return Object.assign(new Error(message), { code: "ECONNRESET" });
 }
 
+/** The reason node mints for a bare `abort()`, as srvx fires on a closed socket. */
+function bareAbortReason(): unknown {
+	const controller = new AbortController();
+	controller.abort();
+	return controller.signal.reason;
+}
+
 describe("dropExpectedClientErrors", () => {
 	it("drops an event whose original exception is an UnauthorizedError", () => {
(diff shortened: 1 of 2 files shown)


igboyes:
Declining both, with reasons.

**Message-string brittleness / guard on `name` alone.** The pairing is load-bearing, not incidental. `name === "AbortError"` on its own is exactly the widening the sibling predicate's comment already warns against for `ECONNRESET` — it would swallow a genuine `abort(reason)` from our own code, and dropping a real incident is silent. If node reworded its default, this filter stops matching and the noise reappears in Sentry, which announces itself immediately. Between a failure mode that shows up as noise and one that shows up as nothing, this is the right way round.

There's also no "narrower context signal" available: `beforeSend` gets the exception and the event, not the request or its signal.

Centralizing the literal means a single-use constant one line from the paragraph explaining why it's that exact string. That makes it harder to track, not easier.

**Helper ordering.** The file is top-down throughout — `dropExpectedClientErrors` sits above all four of its helpers, and each helper above whatever it calls. Moving `isClientDisconnect` below its two branches would make it the one exception to that.

PR #1390, opened 2026-08-14
fix: label the subtraction step Compute GC and Count

igboyes:
## Summary
- `create_subtraction`'s first step displayed as `Compute Gc And Count`. It declared no `name`, so `defineWorkflow` fell back to title-casing the id, which capitalizes `Gc` and `And`. It now declares `Compute GC and Count` explicitly, the label Python shows. The id stays `compute_gc_and_count`, so the cutover contract is untouched.
- Adds `workflow.test.ts`, which this app was missing — it pins both step ids and their order, both display names, non-empty descriptions, and the absent `result`, matching what `create-sample` and `nuvs` already have.

Closes VIR-2970

### DIFF
--- a/apps/create-subtraction/src/steps/computeGcAndCount.ts
+++ b/apps/create-subtraction/src/steps/computeGcAndCount.ts
@@ -17,6 +17,9 @@ import type { CreateSubtractionStep } from "./types";
  */
 export const computeGcAndCountStep: CreateSubtractionStep = {
 	id: "compute_gc_and_count",
+	// Title-casing the id gives `Compute Gc And Count`, which is not the label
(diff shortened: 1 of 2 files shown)


igboyes:
Declining the `PYTHON_STEPS` suggestion.

Deriving the expected ids and names from the workflow would make the test tautological — it would assert that the implementation equals itself, and pass no matter what a step is renamed to. The hardcoded table is the point: it is an independent record of what Python declares, and the ids are a cutover contract, so a rename has to fail here rather than be absorbed silently. That is also why `apps/nuvs/src/workflow.test.ts` already carries the same table in the same shape.

PR #1391, opened 2026-08-14
fix: remove the unassignable Spaces administrator role

igboyes:
## Summary
- The role picker offered `spaces`, but assigning it 500s: Python's migration `997cf9a66f10` remapped those rows to `base` and added an `administrator_role_valid` CHECK permitting only `full`, `settings`, `users` and `base`. Dropped it from `AdministratorRoleName`, the level map, `ADMINISTRATOR_ROLES` and the request validator.
- That same migration dropped the `administratorrole` type, so the Drizzle mirror was describing a type Postgres no longer has. `users.administrator_role` is now `text().$type<AdministratorRoleName>()`, matching `jobs.state`.
- Fixed two comments citing deleted Python code, and dropped the `Free Spaces` line from the `base` role description to match Python's current docstring.

### DIFF
--- a/apps/web/src/server/auth/middleware.test.ts
+++ b/apps/web/src/server/auth/middleware.test.ts
@@ -378,7 +378,7 @@ describe("requireAdminRole", () => {
 		).rejects.toBeInstanceOf(ForbiddenError);
 	});
 
-	it.each(["full", "settings", "spaces", "users", "base"] as const)(
+	it.each(["full", "settings", "users", "base"] as const)(
 		"allows a full administrator to satisfy a %s requirement",
 		async (requiredRole) => {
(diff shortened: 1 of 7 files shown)


igboyes:
Applied in 8b7e9578: `AdministratorRoleName` is now derived from an `ADMINISTRATOR_ROLE_NAMES` tuple, which the request validator's `z.enum` consumes directly — same shape as `PERMISSION_NAMES`.

`ADMINISTRATOR_ROLES` stays a literal: it carries each role's display name and description, mirrored from the docstrings on Python's `AdministratorRole` enum, so there is nothing to derive them from. Its `id` is already typed `AdministratorRoleName` and `data.test.ts` pins the list. The test fake is a fixture and is meant to be an independent copy.

